PR #31, opened 2017-03-19
Update package.json

MrYann:
Added predefined possibilities to some options : either the default compiler pdflatex, or a working setting for latexmk with the lualatex setting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,12 +89,20 @@
                 "latex-workshop.compiler": {
                     "type": "string",
                     "default": "pdflatex",
-                    "description": "LaTeX compiler to use. Reload to take effect. Default: pdflatex."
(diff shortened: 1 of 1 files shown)


James-Yu:
The enum values are quite strange imo. Will incorporate the idea shortly tho. Thanks for contributing!

MrYann:
The enum values are there to present some pre-defined possibilities for the given option. 

So I simply wrote the default value that you provide in the read.me, plus some alternative that I am using (i.e. latexmk with lualatex ; I think latexmk standard is not relevant, since it uses pdflatex, which is already covered by your default option)

Please dont hesitate to tell if I miswrote something.

MrYann:
Also, since you always chain %compiler% and %arguments%, it seems to only complicate things to have separate options. One option for %compile command% might be easier : one line for each different setup.

Issue #6, opened 2017-01-16
Error handling


denis-ko:
Is it possible to implement a better (latex related) error handling?

James-Yu:
Yes it is on the plan. I need to find a good latex log parser yet.

James-Yu:
Latest v0.2.5 implements a LaTeX log parser to better present the errors.

denis-ko:
Are you planning to put these errors/warnings into "diagnosticCollection"? It would be nice to be able to navigate using the "Problems" panel.

James-Yu:
I am still thinking about that. On the one hand it's really cool. But on the other hand the status can only be refreshed upon compilation, and may have some problem with latexmk. Will consider that a second time in the following days.

PR #35, opened 2017-03-24
Report Latex log warnings and errors as vscode diagnostics

jabooth:
This PR changes the behaviour of how errors and warnings detected in the LaTeX log are displayed to the user. Presently, LaTeX-Workshop creates a custom summary of errors and warnings and presents them to the user in the Output Window. As suggested by https://github.com/James-Yu/LaTeX-Workshop/issues/6#issuecomment-273995000, this PR instead builds a `vscode.DiagnosticCollection` from the parsed warnings and sets them on a per-file basis. 

This means after compilation, problems appear in the dedicated 'Problems' window in VS Code, and the user can click on each problem to be navigated to where the error occurred:
image

This unifies errors from LaTeX-Workshop into the same interface that all other warnings and errors are shown in vscode.

As the log parsing information presently appended into the Output window is now redundant (the exact same information is displayed in Problems, with the benefit of rich linking to source files), I removed this as part of this PR. I also don't force focus of the Output window if there is an error (as I would suggest a user would now more commonly like to visit Problems instead).

Finally, I replaced the modal 'compilation failed' banner (which has to be manually dismissed) with a more subtle error message in the bottom status bar. This is perhaps a little more controversial, but I was personally finding it a little 'heavyweight' to have to dismiss the failed error every time. Suspect I might have overstepped the mark a little there, but thought it was worth raising the debate.

Really nice work on this extension, which is proving a lifesaver for my thesis.

From one James to another, hope this is of some use @James-Yu!

### DIFF
--- a/src/compile.ts
+++ b/src/compile.ts
@@ -66,9 +66,7 @@ export async function compile(here = false) {
         });
         // Wait command finish
         await promise.catch((err) => {
-            latex_workshop.workshop_output.append(String(err));
-            latex_workshop.workshop_output.show();
-            vscode.window.showErrorMessage(`Step ${cmd_idx + 1} exited with code ${err.code}. See LaTeX Workshop and Compiler log for details.`);
+            vscode.window.setStatusBarMessage(`Step ${cmd_idx + 1} failed (exit code: ${err.code}).`, 6000);
             error_occurred = true;
         });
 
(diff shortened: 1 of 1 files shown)


James-Yu:
Great contribution from James!

PR #36, opened 2017-03-25
improved citation autocompletion

jabooth:
Few small improvements to citation autocompletion:

1. Add guard to check `citationKey` exists before creating a completion item for entry
2. Display the title in the `detail` section (so it's visible for the highlighted completion)
3. Display all information in the detail view
4. Allow completion based on title as well as citation key 

cite_preview


jabooth:
@James-Yu sounds great! Really looking forward to seeing how the rewrite looks!

James-Yu:
Here it comes with the new `1.0.0` version. Please take a look!

PR #37, opened 2017-03-26
Fix activation language (latex not LaTeX)

jabooth:
I had to tweak the activation language to `latex` (lower case) to get the new v1.0.0 of the Extension to activate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
     "hint"
   ],
   "activationEvents": [
-    "onLanguage:latex"
+    "onLanguage:latex",
+    "onLanguage:LaTeX"
   ],
   "main": "./out/src/main",
   "contributes": {
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "hint"
   ],
   "activationEvents": [
-    "onLanguage:LaTeX"
+    "onLanguage:latex"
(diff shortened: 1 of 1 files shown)


James-Yu:
It's quite strange, as all occurrences of `latex` has been changed to `LaTeX`.
Can you please also append `LaTeX` as an activation language?

package.json
```diff
@@ -26,7 +26,7 @@
     "hint"
   ],
   "activationEvents": [
-    "onLanguage:LaTeX"
```
James-Yu: Can have multiple activation languages.

James-Yu:
Thanks for the quick commit!

jabooth:
@James-Yu so that's really odd, if I have both specified it reliably works all the time (I'm just putting a breakpoint here in [`main.ts`](https://github.com/James-Yu/LaTeX-Workshop/blob/87e5b998a24e7ad6a655829e118f041ba32e3a06/src/main.ts#L16)) - omitting either one seems to cause it to sometimes not be activated..? Anyhoo this definitely fixes it for me!

James-Yu:
My guess is on vscode 'may' cache some configs on language/extension. Glad it fixed the problem :p

PR #40, opened 2017-03-26
move all language ids to lowercase

jabooth:
1. Fixes up the `registerCompletionItemProvider` call for `latex`
2. Moves all language provider IDs to `lowercase-with-dashes` style (e.g. LaTeX Beamer)

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,10 +47,9 @@
         "configuration": "./syntax/syntax.json"
       },
       {
-        "id": "LaTeX",
+        "id": "latex",
         "aliases": [
-          "LaTeX",
-          "latex"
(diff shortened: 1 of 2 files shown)


James-Yu:
This should have been fixed in the latest 1.0.2 version. Will check back later as I'm commuting to work :p

James-Yu:
Double checked and confirmed they were implemented in https://github.com/James-Yu/LaTeX-Workshop/commit/56c84cc4fc15778353e6a17535e06d133f1a96fa . Thanks all the same!

jabooth:
Ah apologies, managed to not pull the latest to check!

PR #41, opened 2017-03-26
Small improvements to map/forEach in citation provider

jabooth:
in general, `array.map()` should be used when a modified array is being made:
```ts
new_array = old_array.map(x => x * 2)
```
(notice the function passed must return something, here we use the fat arrow implicit return property)

whilst `array.forEach()` should be used when you want to just do a for loop over elements without building a new array.

This just tidies up the citation handler to follow these best practices.

BTW, love that you've moved to semicolonless style, personal fave of mine!

### DIFF
--- a/src/providers/citation.ts
+++ b/src/providers/citation.ts
@@ -15,16 +15,16 @@ export class Citation {
     }
 
     provide() : vscode.CompletionItem[] {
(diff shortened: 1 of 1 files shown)


James-Yu:
Nicely done!

PR #44, opened 2017-03-27
better solution to make dependent file lookup robust

jabooth:
Supersedes #43 as a better solution.

Here we only try appending .tex again if we try and fail to read the file. As in general it is probably rare for most tex files to have periods in the name (and as it's quite common to have includes with `.tex` omitted) this avoids doing `n_files` exists checks unlike #43.

As I now have to remove bad entries from the `.texFiles` property, it made sense to turn this into a `Set`, which has O(1) insert/check exists/remove.

### DIFF
--- a/src/manager.ts
+++ b/src/manager.ts
@@ -118,7 +118,7 @@ export class Manager {
             let result = inputReg.exec(content)
             if (!result)
                 break
-            const inputFile = result[1];
+            const inputFile = result[1]
             let inputFilePath = path.resolve(path.join(rootDir, inputFile))
             if (path.extname(inputFilePath) === '') {
                 inputFilePath += '.tex'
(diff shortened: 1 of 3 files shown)


James-Yu:
Thanks for the PR! I did not consider that when coding this segment.

It would be better to use [fs.existsSync](https://nodejs.org/api/fs.html#fs_fs_existssync_path) to avoid the try catch nightmare. nvm I will implement it after merging this PR.

jabooth:
@James-Yu ah that's what I did in my previous solution, but I would suggest this version is better. If you add the `existSync` check that's quite a lot of additional checks every time you come across any file without a `.tex` extension included, which could be quite a lot!

This way we only pay the performance penalty when we find files with periods in them, which is rare.

James-Yu:
Just see this comment after I pushed. Let me find a workaround...

James-Yu:
This one should work as intended and the extra argument is avoided https://github.com/James-Yu/LaTeX-Workshop/commit/741ef6cbc0495133044311ab7c21c843d197b3da . Finger crossed...

Issue #39, opened 2017-03-26
Feature Proposal: add chktex support?


jabooth:
Not sure if this is best handled separately but since I've done a little work on this figured I would at least document it here in case it's of use.

In case you aren't familiar, [Chktex](http://www.nongnu.org/chktex/) is a LaTeX linter - it performs static analysis on `.tex` files and reports numerous warnings for many times of common issues people run into when writing LaTeX.

It was simple enough to setup a task in vscode to run chktex:
```json
{
    "version": "0.1.0",
    "tasks": [
        {
            "taskName": "build",
            "command": "chktex",
            "isShellCommand": true,
            "args": ["-q", "-v0", "thesis.tex"],
            "showOutput": "never",
            "problemMatcher": {
                "owner": "chktex",
                "fileLocation": ["relative", "${workspaceRoot}"],
                "pattern": {
                    "regexp": "^(.*):(\\d+):(\\d+):(\\d+):(.*)$",
                    "file": 1,
                    "line": 2,
                    "column": 3,
                    "code": 4,
                    "message": 5
                },
                "severity": "warning"
            }
        }
    ]
}
```
This parses results that look like this:
```
fitting/supplementary.tex:592:5:44:Wrong length of dash may have been used.
```
and turns them into Problems:
![image](https://cloud.githubusercontent.com/assets/1312873/24331426/1757e6f6-122c-11e7-90ec-b224226a1242.png)

I then used [triggertaskonsave](https://marketplace.visualstudio.com/items?itemName=Gruntfuggly.triggertaskonsave) to launch this every time I save a .tex file:
```json
    "triggerTaskOnSave.tasks": {
        "build": ["**/*.tex"]
    },
```
This setup actually works great and I'm totally happy with it. Saying that it's a little complex to put together for a lot of users, so wrapping this up in an extension would make it a lot more accessable, and make it portable across projects.

I considered making a Chktex extension to do this, but before I did that I wondered if you might be interested in a PR that adds Chktex support to LaTeX Workshop. I'd have to write a lot of the same boilerplate code to find the root tex file for instance, so I could see it making more sense as feature within LaTeX Workshop rather than it's own extension.

Ideally I think chktex linting would be an optional feature, that defaults to being on if they user has chktex installed (can always be disabled/cusomised in settings).

If enabled, chktex would be invoked for the root file on every save in parallel with the build, with results being parsed and shown in the problems window.

[...]

PR #45, opened 2017-03-27
chktex fixes, now asynchronous to compiler errors

jabooth:
There we a few issues in the chktex parser, including:

1. Handling messages containing colons
2. Handling the file in the report (sometimes absolute, sometimes relative)

I fixed these by using the regexp mentioned in #39 which I've been testing the last week or so, fairly sure that's solid.

I also changed this so there are two separate diagnostic collections - one for the linter, one for the compiler. As the linter is fired at different times to the compiler, it's important to keep them separate so that we can clear and update them independently without impacting each other. If not, the linter will fire after a short pause in typing, and when it reports back new changes it would also clear any current compiler errors, which doesn't make sense.

I also fixed a typo I introduced yesterday (`diagnostic_severity`) apologies!

With these changes chktex is working great, thanks for the swift feature addition @James-Yu!

### DIFF
--- a/src/manager.ts
+++ b/src/manager.ts
@@ -18,6 +18,10 @@ export class Manager {
         this.extension = extension
     }
 
+    get rootDir() {
+        return path.dirname(this.rootFile)
+    }
+
     tex2pdf(texPath) {
         return `${texPath.substr(0, texPath.lastIndexOf('.'))}.pdf`
     }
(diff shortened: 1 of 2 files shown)


src/manager.ts
```diff
@@ -18,6 +18,10 @@ export class Manager {
         this.extension = extension
     }
 
+    get rootDir() {
+        return path.dirname(this.rootFile)
+    }
+
```
jabooth: added this property to save having to do this common operation elsewhere in the parser

James-Yu:
Looks really good!

PR #48, opened 2017-03-27
check for citekey for references

### LINKED ISSUE #47: Bibtex parsing fails on @string aliases
Seems the bibtex parsing is struggling in the new version, and I don't know if it's simple to gain the robustness back.

I'm looking into it now. It may make more sense to go back to relying on a third party provider to cover weird edge cases in bibtex.

jabooth:
Fixes #47. 

It's valid bibtex to create aliases using `@string`. 

For instance we can do:
```bibtex
@string{CVIU="Computer Vision and Image Understanding"}
```
And then later do
```
@article{zafeiriou2015survey,
  title={A survey on face detection in the wild: past, present and future},
  author={Zafeiriou, Stefanos and Zhang, Cha and Zhang, Zhengyou},
  journal=CVIU,
  volume={138},
  pages={1--24},
  year={2015},
  publisher={Elsevier}
}
```
and bibtex will fill in the full name.

This `@string` pattern gets caught by the current regexp for detecting the start of an item:
```re
/@(\w+){/g
```
If we additionally check for a citation key followed by comma (allowing for whitespace) we will no longer match on these @string aliases, which avoids an error when we inevitably fail to parse the string in `splitBibItem`.

### DIFF
--- a/src/providers/citation.ts
+++ b/src/providers/citation.ts
@@ -43,7 +43,7 @@ export class Citation {
         if (!fs.existsSync(bib))
             return items
         let content = fs.readFileSync(bib, 'utf-8').replace(/[\r\n]/g, ' ')
-        let itemReg = /@(\w+){/g
+        let itemReg = /@(\w+){\s*\w+\s*,/g
         let result = itemReg.exec(content)
         let prev_result = undefined
         while (result || prev_result) {


James-Yu:
A problem with the new regex is that bibitems with `bib-key` keys will be skipped. Maybe we need more logic for the detection.

What I'm thinking is to whitelist all valid document types, and screen out others when scanning the file.

James-Yu:
I suppose this one should work just fine. Mind have a test?

jabooth:
@James-Yu, yup, your solution works fine, and seems more principled, so long as the list of entities is exhaustive. Nice!

James-Yu:
I've checked multiple sources [here](https://nwalsh.com/tex/texhelp/bibtx-7.html) [here](http://bib-it.sourceforge.net/help/fieldsAndEntryTypes.php) and [here](https://en.wikipedia.org/wiki/BibTeX#Entry_types), all gives the same enumeration so I suppose it's complete :p

PR #51, opened 2017-03-28
Disambiguate ChkTeX and LaTeX diagnositics

jabooth:
Minor refinements to diagnostics.

Now we can disambiguate ChkTeX linting from LaTeX compiler errors:
screenshot 2017-03-28 08 39 14

I also now report the ChkTeX warning number to enable easy comparison with the documentation:
http://www.nongnu.org/chktex/ChkTeX.pdf


James-Yu:
Nicely done!

PR #56, opened 2017-03-28
Initial implementation of real-time as-you-type linting

jabooth:
This PR separates linting into two categories:

1. **`activeFile`** - a real-time check of just the current `.tex` file *in it's current (unsaved) state*.
2. **`rootFile`** - a full check of all `.tex` files in the project starting from the root, *only considering saved files*.

This separation is needed if we want to be able to offer real-time linting on unsaved files, as we need to be able to give `chktex` the current unsaved active file, without giving up on the nice security of full-project linting.

`activeFile` linting is now run at the times where all linting was previously performed (i.e. after a timeout on an `onDidChangeTextDocument` event). A temporary copy of the current file is saved and the linting run on that, with the temp file cleaned up afterwards (just as before). A flag is passed to `chktex` to prevent `\input` directives from being followed. This reduces the runtime significantly, allowing for a quick response of issues on the active file.

`rootFile` linting is simply what was always done in the past - a full lint from the root file, with chktex itself following `\input` directives. This linting is now only run on save in parallel to a build if the `latex-workshop.build_after_save` flag is set.

the same `latex-workshop.linter` flag is respected but now for both types of linting.

This seems to work well and fixes a current issue with the linting - that although it fires after a delay, for all files except the root file it does not respond to changes in the editor, which makes the real-time aspect pointless (you have to save to have chktex see the new version).

In doing this I moved the callback-style code in the `linter` module to `async`/`await` with Promises to avoid too many nested callbacks - hope that's a style you don't mind!

### DIFF
--- a/src/linter.ts
+++ b/src/linter.ts
@@ -4,63 +4,99 @@ import * as vscode from 'vscode'
 import * as path from 'path'
 import * as fs from 'fs'
 import * as cp from 'child_process'
+import * as tmp from 'tmp'
 
 import {Extension} from './main'
+
+function writeFile(filename: string, data: any) {
+    return new Promise((resolve, reject) => {
+        fs.writeFile(filename, data, (err, result) => {
+            if (err) {
+                reject(err);
+            } else {
+                resolve(result)
+            }
+        })
+    })
+}
 
 export class Linter {
     extension: Extension
     linterFile: string
     linterTimeout: NodeJS.Timer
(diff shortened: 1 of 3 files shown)


James-Yu:
I've no problem with `async`/`await`, though I am not quite familiar with them. A lot of work done bro!

jabooth:
@James-Yu thanks mate! Just landed #57 which refines this a little more, hope you like it! :)

PR #57, opened 2017-03-29
Linting refinements (avoid tmp file, more robust error reporting)

jabooth:
This PR does the following:

1. Split the linter command into `linter_command_active_file` and
`linter_command_root_file` to be able to customise independently
2. Moves to have these commands specified as arrays (`[cmd, arg1, arg2]`) to avoid whitespace issues when invoking
3. Directly pipe the content of the active file to `chktex` rather than making a temporary file which needs cleaning up
2. Parse the length of the reported issue and include this in diagnostics
3. Reduce default linting interval to 300ms
4. More robustly handle issues around the linter subprocess call

Here's a little video of the linting experience on a small demo project now:
demo_chktex_2

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,17 +157,22 @@
         "latex-workshop.linter": {
           "type": "boolean",
           "default": false,
-          "description": "Lint LaTeX when no document changes for a defined period of time."
+          "description": "Enable linting LaTeX with ChkTeX.\nThe active document will be linted when no document changes for a defined period of time.\nThe full project will be linted from the root on file save."
         },
-        "latex-workshop.linter_command": {
-          "type": "string",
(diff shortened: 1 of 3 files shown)


James-Yu:
Looks awesome! Great work!

One more thing, I would like to ask if you may join the repo as a collaborator for faster development and management. I believe that will make the extension perfect!

jabooth:
Thanks @James-Yu!

I would love to join you on the repo as a collaborator, many thanks for asking! :)

I think you're doing a fantastic job with this extension, and there is every chance to make LaTeX Workshop + VSCode the clear tool of choice for LaTeX writing *period*.

Super happy to try and make that happen by helping out however I can!

PR #58, opened 2017-03-29
minor tidying up of configuration

jabooth:
- The search exclusion of `out` wasn't working without the `'**'` glob
- Updating a renamed key in `tasks.json` (`isWatching` now `isBackground`)
- Removed two unused deps (`hasbin`, `mocha`)
- Add types for `glob`, `tmp`, `ws`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,8 +4,8 @@
         "out": false // set this to true to hide the "out" folder with the compiled JS files
     },
     "search.exclude": {
-        "out": true // set this to false to include "out" folder in search results
+        "out/**": true // set this to false to include "out" folder in search results
     },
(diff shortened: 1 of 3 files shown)


jabooth:
Btw @James-Yu I'll continue to do PRs if that's OK so you can see all changes I make in case there is something you don't like. If I'm pretty confident it's fine shall I just merge immediately or wait for your thoughts?

With this one as I'm removing those two unused deps I would wait for your OK regardless, just in case I'm missing something!

James-Yu:
Please feel free to merge when you feel good about the changes! Well done!

jabooth:
@James-Yu awesome, will do going forwards! 👍

PR #67, opened 2017-03-30
Refine #66 to use existing lintActiveFileIfEnabled func

jabooth:
@James-Yu apologies for you needing to do #66, this just tweaks that to use the existing function for this purpose.


James-Yu:
Did not noticed that function. Awesome!

PR #71, opened 2017-04-02
Use `chokidar` to watch file changes

James-Yu:
This PR re-writes the file dependency logic.

Now all files are registered in a file watcher provided by `chokidar`. Their contents will be read and parsed only when the files are changed (by vscode or external editors). This change should reduce a lot of disk IO when intellisense is frequently invoked, as previously all files are looked into upon intellisense calls (one second interval though).

The new logic is as follows. Whenever the root file changes, all dependencies are traversed. If any of the files is not watched before, it is added to the watcher and parsed immediately. When a file is watched, their contents will be parsed on changes. Results are cached for intellisense. Delete a file will remove it from the watcher.

Notable changes start from https://github.com/James-Yu/LaTeX-Workshop/compare/watch-files?expand=1#diff-fd1d96e4c46ceca0dcabbce45e9df5f5R103 and https://github.com/James-Yu/LaTeX-Workshop/compare/watch-files?expand=1#diff-fd1d96e4c46ceca0dcabbce45e9df5f5R114 . Intellisense code is updated.


James-Yu:
@jabooth Do you mind check if this PR works correctly on complex structures of LaTeX projects? I have checked on a fairly simple one and both intellisense (citation, reference, command) and real-time edits (from vscode and other editors) work. Thanks!

jabooth:
@James-Yu See #72 for my review of this PR. On the whole though really like it, great improvement too!

PR #73, opened 2017-04-02
Add tslint to project, style fixes

jabooth:
This PR adds [tslint](https://palantir.github.io/tslint/) to keep the code quality of LaTeX Workshop high.

This PR is on top of #72 (and hence #71) and so should be merged after those are first merged. If we need to make changes to #72 we will just have to merge those changes in with this PR too.

I've added settings for tslint that I feel match the current code style for LaTeX Workshop, and encourage best practices. In this PR I have implemented the recommended fixes found by tslint.

The largest source of changes are the following two rules:

1. Variables should be `const` if they are never mutated.
2. `if`/`else`/`for` statements should always have curly braces.

I believe that these two rules are well worth following, but if you disagree @James-Yu let me know! (Some thoughts on the [second one](http://stackoverflow.com/a/8020255)).

The other things fixed in here are minor things like tabs instead of spaces, missing newlines on the end of files. 

It looks like a lot of changes but it's all stylistic - there no new functionality added in this PR.

Finally, this PR also tightens the settings in `tsconfig.json` to be a little stricter in terms of type safety. This currently leads to 52 typescript compiler errors, which we should address. We can either:

1.  Leave those for now and address them in future PRs. Just slightly annoying that Code will warn us that there were errors compiling when trying to debug. (it's not an issue, it's the same code we've been using for ages, we are just been stricter now in spotting potential issues!)
2. Dial back the `tsconfig.json` settings until we can do a PR which ups the settings and includes all fixes

Whichever you prefer!


James-Yu:
I was just too lazy to set up these linting things. Thanks bro!

Let me check if there are ways to fix these debug errors 👍

PR #78, opened 2017-04-03
Spellchecker

James-Yu:
This PR add a spell checking feature to the extension. A new command `spell` can check the spellings of current document, regardless `.tex` or not. A new configuration `spellcheck` can control auto checking. If enabled, current document (changes) are checked upon opening new, changing active editor and changing contents.

Code actions for mis-spell correction and ignoring (local and global) can be added. That's the next plan :p


jabooth:
@James-Yu Wowsa, huge! 😮 😄 

Have to say this is currently the biggest weakness of the experience of VS Code for LaTeX so love that you are working on this feature!

At the same time we should be a little cautious here just because it sounds like Microsoft is finally getting serious about adding an [official spell checker](https://github.com/Microsoft/vscode/wiki/Roadmap#core-editor) (https://github.com/Microsoft/vscode/issues/20266). That's currently got a milestone of April 2017 so not too long...

We should perhaps start being active in that discussion so that we can be sure that when an official spell checker ships it's designed to work great with LaTeX documents.

Still in the meantime this is a great addition. I'll give it a test and let you know my thoughts! 👍

James-Yu:
aww never knew that. Will check the issue when I come back home :p

jabooth:
Also my current solution is to use [vscode-spell-checker](https://github.com/Jason-Rev/vscode-spell-checker) which I've found pretty good with some config for LaTeX (but I still get a lot of false postitives). It looks like the author @Jason3S (hi!) has done [some nice work](https://github.com/Microsoft/vscode/issues/20266#issuecomment-288855724) to keep the checker really efficient by preprocessing the `.dic` files, might be something we can learn from here!

James-Yu:
I checked the 'normal' (non-coding) spell checkers and no one works really well for LaTeX. Will check vscode-spell-check soon.

I also followed the discussion on vscode official spell checker. It seems really promising, but M$ did a bad job in following its plans (at least for some features of vscode. Noticed a few postponed for more than five times). If other extensions don't work as good, maybe we can first release our version and wait for the official one. What do you think @jabooth ? :bowtie:

jabooth:
>  maybe we can first release our version and wait for the official one. What do you think @jabooth ? :bowtie:

@James-Yu yeah I think that's a fine strategy! I'm sure Microsoft might let this slip (it already went from March -> April...) so I can see the value in solving the problem for users now.

> I checked the 'normal' (non-coding) spell checkers and no one works really well for LaTeX

Yeah I know what you mean. And LaTeX is quite an unusual case. 

Broadly speaking I think you have quite different needs for different programming languages. You are unlikely to write prose in C++, so the spell checker is more about dealing with isolated words and handling the fact that those words might be joined in interesting ways like `camelCase` or `underscore_separated`.

On the opposite end of the spectrum is something like Markdown where you basically want a full-blown spelling and grammar checker that is akin to what you have in something like Word/Google Docs.

LaTeX is in-between those words - you want sophisticated checking when writing prose, but you need the checker to not be showing false-positives for `$$` `\commands{}` etc. Perhaps that niche will mean there is value in having a custom spell checker, at least for a while!

Jason3S:
Interesting discussion.

Spell checking can be challenging.  One of the biggest as @jabooth mentions is false positives.

Since I was trying to create a generic spell checker that worked for many programming languages and text formats, I took the approach of having a generic parser that separated out words which were then compared to a set of dictionaries.

To make it work for specific file types, I created "keyword" dictionaries. Here is the one for [LaTeX](https://github.com/Jason3S/cspell/blob/master/dictionaries/latex.txt).  It is possible to have file format specific dictionaries and regex to include / exclude text from the spell checker.

Adding the spell checker to the editor will have some advantages if it is able to take advantage of the grammar parser.  The it will be able to spell check *only strings* or *only comments* or even have a different language associated with *strings* vs the rest of the code.

To be clear, checking the spelling of a word was the easy part.  Making it configurable and extendable was the hard part (and still needs some work).

@jabooth if you have any suggestions on how to improve LaTeX parsing and checking, please let me know, I can work with you on making a Pull Request.

James-Yu:
Just checked Code Spellchecker and have to say it's awesome! @Jason3S great job!

@jabooth I added two regexs to the ignore list and the result is good:
- `\\\\\\w*(\\[.*?\\])?(\\{.*?\\})?` to exclude all commands,
- `\\$.+?\\$` to exclude everything inside math-mode
I know it might be somehow radical (is this word correct? 🤣 ). But it seems not giving me false-positives -- at least in main contents.

Kudos again to @Jason3S ! I will check your code and see if I can make a PR for LaTeX support.

PR #81, opened 2017-04-04
Add a quickpick menu for all commands

James-Yu:
This PR add a quickpick menu for all commands. The menu can be invoked by either clicking the status bar item or right click latex files. To make the context menu more compact, many items are removed.

Another small change is that when clicking the status bar item, previous message is displayed again.


James-Yu:
Tested for a while, seems fine.

PR #82, opened 2017-04-04
organise autocomplete, refine presentation

jabooth:
This PR, doesn't add any new completions after a7b0769, but it does:

1. Begin to organise the completions by category (e.g. math mode commands, text sizing commands etc). For now this is unused, but if we wanted to make the completion more sophisticated (e.g. only offer math mode completions between `$$`) it would be easier to start doing so.
2. Adds `detail` keys for a number of items which provide the unicode symbol for the math item:

screenshot 2017-04-04 09 17 11

I found this was a great reference:

http://milde.users.sourceforge.net/LUCR/Math/unimathsymbols.pdf

And before I started going too far doing this manually I almost considered if we could parse this to produce a really exhaustive list of completions with 'preview' details as above...We could even put the comments section into the `description`, so the user could hit `cmd` + `space` and see an explanation of what that LaTeX command does..! Imagine if LaTeX Workshop basically offered full LaTeX documentation when you hit  `cmd` + `space` 😮 ...

Anyway, an idea for the future.

3. Displays a better string for what will be inserted for the command:
screenshot 2017-04-04 09 21 28

this is closer to how sharelatex do it and I find it's nice and reassuring. :)

### DIFF
--- a/src/providers/command.ts
+++ b/src/providers/command.ts
@@ -54,18 +54,25 @@ export class Command {
         this.suggestions = []
         Object.keys(suggestions).forEach(key => {
             const item = suggestions[key]
-            const command = new vscode.CompletionItem(item.command, vscode.CompletionItemKind.Keyword)
+            const command = new vscode.CompletionItem(`\\${item.command}`, vscode.CompletionItemKind.Snippet)
             if (item.snippet) {
                 command.insertText = new vscode.SnippetString(item.snippet)
+            } else {
+                command.insertText = item.command
             }
             if (item.documentation) {
                 command.documentation = item.documentation
             }
+            if (item.detail) {
+                command.detail = item.detail
+            }
             this.suggestions.push(command)
         })
+
         Object.keys(this.envSnippet).forEach(key => {
             const item = this.envSnippet[key]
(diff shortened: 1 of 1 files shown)


James-Yu:
Awesome!

Let me check if we can get the data from that pdf file. 😀

jabooth:
@James-Yu:
http://milde.users.sourceforge.net/LUCR/Math/data/unimathsymbols.txt

A plain text varient...!

jabooth:
Here's more info on this whole project:
http://milde.users.sourceforge.net/LUCR/Math/

James-Yu:
@jabooth Nice catch! Brought out my python!

jabooth:
@James-Yu ahhh I do love myself a little Python...here's my 'day job':
http://www.menpo.org/

Btw it seems all this stuff is provided under the LateX Project Project Licence...but on first glance that seems totally fine and compatible with MIT (although of course I am not a lawyer!) It's VERY comprehensive all this stuff, wonder if it will generate too much info! Can't wait to see how you get on!

James-Yu:
Here it comes!

PR #84, opened 2017-04-04
(#82) Add uni-math symbols

James-Yu:
This PR added [Unicode symbols](http://milde.users.sourceforge.net/LUCR/Math/) to default command hints. All symbols are provided with their unicode presentation, some of them have documentation (parsed one, not manually corrected).

e.g.
image


James-Yu:
This ugly python code is used to parse the [source file].(http://milde.users.sourceforge.net/LUCR/Math/data/unimathsymbols.txt)

```python
import json

data = {}

with open('unimathsymbols.txt', encoding='utf-8') as f:
    for line in f:
        if line[0] is '#':
            continue
        segments = line.split('^')
        if segments[3] is '':
            continue
        if segments[3][0] is '\\':
            segments[3] = segments[3][1:]
        data[segments[3]] = {
            'command': segments[3],
            'detail': segments[1],
            'documentation': segments[7]
        }
        if segments[6] is not '':
            data[segments[3]]['requirement'] = segments[6]

json.dump(data, open('unimathsymbols.json', 'w', encoding='utf-8'), ensure_ascii=False)
```

jabooth:
Amazing!

Few suggestions:

1. We should keep the Python file used to parse the symbols somewhere for later use. Maybe we make a new `support` top level folder and put it in there? We can add a brief README to explain what the code does and obviously exclude it from being downloaded as part of the extension.
2. add a `.strip()` on the end of 
```
            'documentation': segments[7].strip()
```
to get rid of the `\n` on each `documentation` entry.
3. change the export to:
```py
json.dump(data, open('unimathsymbols.json', 'w', encoding='utf-8'), 
          indent=4, separators=(',', ': '), sort_keys=True, ensure_ascii=False) 
```
to keep the data a little easier to parse for a human :)

src/providers/command.ts
```diff
@@ -60,7 +60,7 @@ export class Command {
         Object.keys(suggestions).forEach(key => {
             const item = suggestions[key]
             const backslash = key === ' ' ? '' : '\\'
-            const command = new vscode.CompletionItem(`${backslash}${item.command}`, vscode.CompletionItemKind.Snippet)
+            const command = new vscode.CompletionItem(`${backslash}${item.command}`, vscode.CompletionItemKind.Function)
```
jabooth: nice, yeah was thinking it would be cool to distinguish these more!

James-Yu:
Done! 😆

jabooth:
@James-Yu fantastic! Only other question is performance, have we tried it to see how VS code holds up?!

jabooth:
at least on my iMac it seems pretty fine!

James-Yu:
Haven't tried profiling, it seems just fine. Let me change the loading to asynchronous/callback manner later. :P

jabooth:
Yeah sounds good!

Btw one other change - the automatically parsed docs are waaay better than the few I manually added, so I think we should actually remove them (it seems they currently take precedence over the parsed, which I think is the right pattern in general just here they are worse!)

jabooth:
So that's anything in data/commands at the bottom below `sqrt` at least...

James-Yu:
These commits optimize startup and real-time performance. Now files are loaded and parsed in callbacks - asynchronous. These files are parsed and `CompletionItem`s are created on initialization. So in real-time intellisense only user-input commands are parsed and added to the initialized list.

😃 😆 😃 😆 😃 😆 😃 😆

jabooth:
@James-Yu very nice!

in practice it seems to work really well for me!

only one thing is that I worry that for common cases having so many autocompletes might make things a little harder to find.

screenshot 2017-04-04 13 01 52

For instance, I imagine most people here really want the `\begin{equation}...` completion rather than the more exotic `/errbarblacksquare` 😆 

I tried to play with the `.sortKey` property to make all the math envs completions sort to the bottom - I think that would be a good solution, but I can't get it to work! I tried this:
```ts
if (supressOrder) {
    command.sortText = `ZZZ${command.sortText ? command.sortText : command.label}`
}
```
but it doesn't seem to work. Any ideas?

jabooth:
@James-Yu ah it seems to be related to the `vscode.CompletionItemKind` if I make the `\begin...` snippets have type `vscode.CompletionItemKind.Class`:
screenshot 2017-04-04 13 12 08

Issue #90, opened 2017-04-06
More Detail to Citation Autocomplete


jsinglet:
Often times the citation key isn't enough to identify a citation. What would be really helpful, rather than the citation key, would be an abbreviated version of the full BibTeX entry. At a minimum:

citation key: 
Authors, Year in: Publication
Title

Being able to narrow this with typing would be ideal. If this is not underway I may be able to help with this feature.

James-Yu:
Now you can use author names, title, and citation key to filter the autocomplete items. The detailed information is in the blue information icon. I did not quite get what you mean by 'abbreviated version'. Can you please elaborate more?

![image](https://cloud.githubusercontent.com/assets/4210342/24734261/f7fad4cc-1aae-11e7-93f8-d9c4f6fb7995.png)

jsinglet:
Ah, I just updated and now I see that. However, I still can't see the titles. Like in your screenshot they are cut off (as well as the author information). Is there a way to be able to see more?

James-Yu:
The autocomplete window width is fixed in vscode. Alternative suggestions are always welcomed.

jsinglet:
https://github.com/Microsoft/vscode/issues/369  

Oh my....

James-Yu:
Yes it seems the final solution they worked out was the blue icon. Unfortunately it seems that we cannot modify it on the extension side. 😢

jabooth:
Yeah when I implemented this richer citation intellisense I was a little frustrated at these restrictions. 

#### Current mitigations

**1.** we have implemented it so that although you might not be able to *see* all the information, you can still filter based on all the info. By that I mean if you know you are looking for the paper:

 _Why is it that all research papers have pathologically long titles: a study_

and you remember 'pathologically' being in the title, you should be able to do:
```
\cite{patho
```` 
And even if the key is `booth2006unrelated` the entry should still show up.

**2.** `cmd` / `ctrl` + `space` triggers the exanded info from the keyboard rather than hitting the blue i.


#### Other ideas - Citation Browser

In addition to the current intellisense feature, we could have a model window for citations that drops down from the top of VS Code. I was thinking we could call it something like `Citation Browser` - you could trigger it whenever you want to hunt down a specific citation in a window where we will be able to show more info.

I was thinking we could do this using whatever API we currently use for the `Latex Workshop Actions` panel.

#### Other ideas - Rearrange info

[...]

PR #91, opened 2017-04-06
Customiz-able citation intellisense, citation browser

James-Yu:
This PR adds a `latex-workshop.citation_intellisense_label` config item to control what to display in citation intellisense window. Default is `bibtex key` which is identical with what it is now. Below is a demo for `title`:
image

This PR also add a citation browser, accessible from the action dropdown menu:
image

Select any items in the menu will insert its key.

#90

### DIFF
--- a/src/providers/citation.ts
+++ b/src/providers/citation.ts
@@ -81,14 +81,20 @@ export class Citation {
         const pickItems: vscode.QuickPickItem[] = items.map(item => {
             return {
                 label: item.title as string,
-                description: '',
-                detail: `Key: ${item.key}, authors: ${item.author?item.author:'Unknown'}`
+                description: `${item.key}`,
+                detail: `Authors: ${item.author ? item.author : 'Unknown'}, publication: ${item.journal ? item.journal : (item.publisher ? item.publisher : 'Unknown')}`
             }
         })
         vscode.window.showQuickPick(pickItems, {
             placeHolder: 'Press ENTER to insert citation key at cursor'
         }).then(selected => {
-            console.log(selected)
+            if (!selected) {
+                return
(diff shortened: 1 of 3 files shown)


src/providers/citation.ts
```diff
@@ -35,10 +35,32 @@ export class Citation {
...
             const citation = new vscode.CompletionItem(item.key, vscode.CompletionItemKind.Reference)
             citation.detail = item.title
-            citation.filterText = `${item.author} ${item.title} ${item.journal}`
+            switch (configuration.get('citation_intellisense_label') as string) {
+                case 'bibtex key':
+                default:
+                    citation.label = item.key
+                    break
+                case 'title':
+                    if (item.title) {
```
jabooth: I think here we should set
```ts
citation.detail = undefined
```
so that the title is not clipped by vscode trying to show the title again in the detail field..

src/providers/citation.ts
```diff
@@ -35,10 +35,32 @@ export class Citation {
...
+                        citation.label = item.title as string
+                    } else {
+                        citation.label = item.key
+                    }
+                    break
+                case 'authors':
+                    if (item.author) {
+                        citation.label = item.author as string
```
jabooth: Perhaps here too. Basic idea being: 

1. by default we show you the bibtex key for sure and try and show as much as the title as possible in `detail`.
2. You can override this if you rather the view be maximised for either `title` or `author`, but in these cases we should hide the `detail` to give you as much room as possible for the thing you have selected.

James-Yu:
Forgot that 😝

jabooth:
citation browser is awesome!

Issue #92, opened 2017-04-06
Proposal - move latex workshop actions to top level actions?


jabooth:
Right now the only commands latex workshop adds to vscode's global set of commands seems to be 

<img width="630" alt="screenshot 2017-04-06 10 08 26" src="https://cloud.githubusercontent.com/assets/1312873/24746439/0867d772-1ab1-11e7-8247-2e38a100bab6.png">

I think here we should provide all commands we currently expose under `LaTeX Workshop Actions`

<img width="641" alt="screenshot 2017-04-06 10 09 19" src="https://cloud.githubusercontent.com/assets/1312873/24746474/252cf662-1ab1-11e7-8a85-80f9fa118812.png">

Perhaps with a a common `LaTeX Workshop: ` prefix - that way users typing `latex` into the top level can execute any command we expose in one click rather than two.

BTW, I still like the idea of the `LaTeX Workshop Actions` menu being accessable through right click as this is great for discoverability! But I think the 'vscode' way of doing this is to encourage users to try `cmd`/`ctrl` + `shift` + `p` and then typing any command directly.

jabooth:
Sorry @James-Yu I've updated the body of the issue now haha!

James-Yu:
Easy job!

PR #94, opened 2017-04-06
(#92) Add LaTeX Workshop prefix in quick menu

James-Yu:
This PR adds a `LaTeX Workshop: ` prefix to commands in quick menu.
image

Should the citation browser be added as a vscode command or as is?

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,6 +149,11 @@
         "category": "LaTeX Workshop"
       },
       {
+        "command": "latex-workshop.citation",
+        "title": "Open citation browser",
+        "category": "LaTeX Workshop"
+      },
+      {
         "command": "latex-workshop.actions",
         "title": "LaTeX Workshop Actions"
       }
--- a/package.json
+++ b/package.json
@@ -125,23 +125,28 @@
     "commands": [
       {
         "command": "latex-workshop.build",
-        "title": "Build LaTeX project"
+        "title": "Build LaTeX project",
+        "category": "LaTeX Workshop"
       },
       {
         "command": "latex-workshop.view",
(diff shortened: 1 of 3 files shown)


James-Yu:
Better!
image

jabooth:
@James-Yu 

> Should the citation browser be added as a vscode command or as is?
 
Yes I think it should be there! Any command from LaTeX Workshop should really be exposed this way I feel.

jabooth:
@James-Yu haha you're too fast for me to keep up! 😉

jabooth:
@James-Yu I think we should maybe remove `LaTeX Workshop Actions` from this list, it seems pretty redundant now? It should still exist to be accessable from right click, but for the top level command it doesn't really have a role?

James-Yu:
That would be a bit difficult. afaik the quick menu shows all vscode commands, and only these commands can be added to right click context menu and status bar click command. Will check the API again to see if there are ways to overcome it.

jabooth:
Is there an API to trigger the quick menu with a string in it?

So we could trigger with `>  LaTeX Workshop` pre-inserted which will give you exactly the above?

jabooth:
Ohh I mis-read, we need a command for the context menu click so this wouldn't work would it... :|

James-Yu:
I'm afraid this is the best we can achieve

image

jabooth:
Hmm maybe we could we do:
```
LaTeX Workshop: Show all actions
```
So it's consistent with the others? Idk though, it's minor :)

James-Yu:
Sounds good. A small tweak

PR #96, opened 2017-04-07
Change some configs, alerts on finding obsolete ones

James-Yu:
This PR changes `linter_command_active_file` and `linter_command_root_file` configs. ChkTeX commands are separated into `linter_command` `string` + `linter_arguments_` `string` + required arguments (hard-coded, documented in config comments). Seems work well.

In addition, if the extension finds obsolete configs, i.e., `linter_command_active_file` and `linter_command_root_file`, it will show a warning message indicating what is deprecated and what is the replacement. Hope it will not be too painful.

This is a start of the re-work on cleaning up configs. Suggestions welcomed!

### DIFF
--- a/src/viewer.ts
+++ b/src/viewer.ts
@@ -11,6 +11,7 @@ interface Position {}
 
 interface Client {
     type: 'viewer' | 'tab'
+    prevType?: 'viewer' | 'tab'
     ws?: WebSocket
     position?: Position
 }
@@ -101,12 +102,16 @@ export class Viewer {
                 client = this.clients[decodeURIComponent(data.path)]
                 if (client !== undefined) {
                     client.ws = ws
+                    if (client.type === undefined && client.prevType !== undefined) {
+                        client.type = client.prevType
+                    }
                 }
                 break
             case 'close':
                 for (const key in this.clients) {
                     client = this.clients[key]
                     if (client !== undefined && client.ws === ws) {
+                        client.prevType = client.type
                         delete client.ws
                         delete client.type
                     }
--- a/README.md
+++ b/README.md
@@ -61,22 +61,59 @@ Choose the item in the menu to have LaTeX Workshop attempt to fix the issue:
 
 ## FAQ
 ### <a name="toolchain"></a>LaTeX toolchain?
-LaTeX toolchain refers to a sequence/array of commands which LaTeX Workshop will execute sequentially when building LaTeX projects. It is set in `File`>`Preferences`>`Settings`>`latex-workshop.toolchain`. By default [`latexmk`](http://personal.psu.edu/jcc8/software/latexmk/) is used. For non-perl users, the following `texify` toolchain may worth a try:
-```
-[ "texify --synctex --tex-option=\"-interaction=nonstopmode -file-line-error\" --pdf %DOC%.tex" ]
-```
+LaTeX toolchain refers to a sequence/array of commands which LaTeX Workshop will execute sequentially when building LaTeX projects. It is set in `File`>`Preferences`>`Settings`>`latex-workshop.latex.toolchain`. This configuration item is an array of objects, which should have a `command` field and an optional `args` array field. The former defines the command that will be invoked in each step of the toolchain, while the latter defines the arguments that will be passed alongside with the command.
 
-LaTeX toolchain must always be defined as a JSON array, even if there is only one command to execute. For multiple commands, each one is represented by a string in the array.
-
-The placeholder `%DOC%` in all strings will be replaced by the quoted LaTeX root file name without `.tex` extension on-the-fly. Alternatively, you can also set your commands without the placeholder, just like what you may input in a terminal.
-
-As most LaTeX compiler accepts root file name without extension, and `bibtex` requires to do so, `%DOC%` does not include `.tex` extension. Meanwhile, `texify` requires the extension. So in the above toolchain `%DOC%` and `.tex` are concatenated for completeness. For manual toolchains as the following one, this is not a must:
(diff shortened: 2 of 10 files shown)


src/viewer.ts
```diff
@@ -11,6 +11,7 @@ interface Position {}
 
 interface Client {
     type: 'viewer' | 'tab'
+    prevType?: 'viewer' | 'tab'
```
James-Yu: This file change is by accident. The main purpose is to retain preview type when user accidentally refresh pdf viewer. Not quite related with the PR.

jabooth:
OK @James-Yu the obsolete commands stuff is great! All my comments are really about the structure of the config. I feel as it's a breaking change it's worth taking the time to get this right, so, let's debate this one out!

I took a look at some other extensions to see how they handle similar issues, see if we can learn from them. 

A few things to note from the Python extension:

1. **They use periods to create a namespace for related commands** (`python.linting`). I think this actually really helps to keep things ordered. As LaTeX Workshop gets more powerful (and configurable) having this kind of design would be useful I imagine!

```
  // Whether to lint Python files.
  "python.linting.enabled": true,

  // Whether to lint Python files when no workspace is opened.
  "python.linting.enabledWithoutWorkspace": true,

  // Whether to lint Python files using prospector.
  "python.linting.prospectorEnabled": false,

  // Whether to lint Python files using pylint.
  "python.linting.pylintEnabled": true,
  ...
```

2. **They make you provide the binary location independently from the arguments, and arguments are an array**. There's things to like about this design. Firstly, it makes programatically easy to validate and update the binary location (e.g. using #97). Secondly, it's feasible that a user will want to set one of these two things in their global user settings (e.g. the location of `chktex`) but then on a per-workspace setting edit just the arguments passed (to customize the linting warnings for instance).

```
 // Path to Pylint, you can use a custom version of pylint by modifying this setting to include the full path.
  "python.linting.pylintPath": "pylint",

...

  // Arguments passed in. Each argument is a separate item in the array.
  "python.linting.pylintArgs": [],

```
I also quite strongly believe that the better design is to **require arguments as an array and not a string**. I believe this format is widely considered best practice (at least in Python I remember this) as it avoids all the issues of escaping which can be shell/platform dependent. By handing node the actual arguments you want to pass, node can provide everything to the binary in the correct way for the current platform, which is less error prone. I think as an example here, if a user has a space in a filename and we sub it into the string form, we now have the responsibility to correctly escape that string. If we provide it as one argument in a list, node does that for us. See this [Stack Overflow](http://stackoverflow.com/a/15109975) (which is for Python but I believe the principle applies here too).

It's also nicer from the point of view of us having optional/non-optional arguments. we can simply concatenate the lists together rather than doing any error prone string join.

But saying that I know the string form is perhaps more beginner friendly...what's your view on this? Is there a strong argument for string over array that I perhaps don't know about?

James-Yu:
Great comments! I originally did not think much on the list-method, but it seems to be more "scientific". Will change changes shortly!

One problem is with `toolchain`, which is already a list. My two cents is on keeping it as is. list-of-list-of-strings seems to be too complex for users to adapt to.

Also, the idea of using dots to separate is awesome!

jabooth:
> Great comments!

Thanks @James-Yu! 👍 

> I originally did not think much on the list-method, but it seems to be more "scientific". Will change changes shortly!

Glad you agree!

> One problem is with toolchain, which is already a list. My two cents is on keeping it as is. list-of-list-of-strings seems to be too complex for users to adapt to.

Yeah this one is tricky! I agree that it gets more complex to have this done 'correctly'. Just to play it out, I guess a complete 'correct' solution would actually be a list of objects with keys `"executablePath"`, `"args"`, so something like:

```json
"latex-workshop.toolchain" : [
    {
      "executablePath": "/path/to/pdflatex",
      "args": [
          "--interaction-mode",
          "whatever",
          "--just-a",
          "example",
          "%DOC%"
      ]
    },
    {
      "executablePath": "/path/to/bibtex",
      "args": ["%DOC%"]
    }
]
```
If we did this, there would be some benefits:

1. We would be consistent in how we name/handle calling child_processes throughout the extension (not sometimes string, sometimes array).
2. The consistency could be good for users, as we would always ask for them to provide stuff in this same format.
3. We would robustly have a list of executables that the user is looking to use in their toolchain. We could consider checking the stems of these executable paths (above we would find `bibtex` and `pdflatex`) and if the file is not found we could help find it using #97.


As a mitigation against it being confusing for users, I believe we can add a schema to enforce this structure, so users would get a clear error if they make a mistake! Not too sure how we do this though? (I've noticed on some settings you will get a typescript like error if you provide the wrong type for the setting?)

This one I agree is a harder sell, I can see this both ways!

jabooth:
@James-Yu on a side note, I think we should maybe use the term `executable` everywhere we have been using `binary`...it's the more accurate term I think. 

For instance, `latexmk` is a Perl script so isn't really a binary (which implies compiled code) but is something you can execute.

It's a minor detail, but I could imagine it perhaps confusing users a little so perhaps worth addressing! Think it only impacts this and #97 so no breaking change I don't believe...?

James-Yu:
This commit does most of the renaming things. Please have a check if anyone needs a rename. 😁 

I will work on
- change `synctex` to child_process.spawn pattern
- refractor `toolchain` things
a bit later. Cheers! 😅

jabooth:
@James-Yu looks great!

Btw sorry I haven't been helping as much with the actual coding last few days, had some stuff on my plate, and this has been taking some time up:

https://github.com/latexjs/latexjs

I'm hoping though that soon that will bear fruit and perhaps be interesting for LaTeX Workshop! :D

James-Yu:
Done! @jabooth take a look?

jabooth:
Sorry @James-Yu I was AFK over the weekend but this all looks great to me. Let's see how users get on with the migration!

PR #97, opened 2017-04-07
Binary check

James-Yu:
This PR add a new command `latex-workshop.binaries` to check the validity of `synctex` and `chktex`. If anyone is invalid, the extension searches some pre-defined common paths and warns for change. If no alternative binary is found, an error message is popped.

This PR is based on #96 , and should be merged after that.

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,40 +207,27 @@
         "latex-workshop.synctex_command": {
           "type": "string",
           "default": "synctex",
-          "description": "Define the location of SyncTeX executive file."
+          "description": "Define the location of SyncTeX executive file.\nAdditional arguments, e.g., synctex modes and position of click, will be appended to this command."
         },
         "latex-workshop.linter": {
           "type": "boolean",
           "default": false,
           "description": "Enable linting LaTeX with ChkTeX.\nThe active document will be linted when no document changes for a defined period of time.\nThe full project will be linted from the root on file save."
         },
-        "latex-workshop.linter_command_active_file": {
-          "type": "array",
-          "default": [
-            "chktex",
-            "-wall",
-            "-n22",
-            "-n30",
-            "-e16",
-            "-q",
-            "-I0",
-            "-f%f:%l:%c:%d:%k:%n:%m\n"
-          ],
-          "description": "Linter command to check LaTeX syntax of the current file state in real time with ChkTeX.\nCurrent file contents will be piped to the command through stdin.\nThe format string should be kept as shown in the default as this is the format the parser expects for successful parsing."
+        "latex-workshop.linter_command": {
+          "type": "string",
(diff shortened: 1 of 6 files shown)


src/commander.ts
```diff
@@ -72,6 +72,24 @@ export class Commander {
         this.extension.completer.citation.browser()
     }
 
+    binaries() {
+        function windowsPaths(command: string, disks: string[]=['C', 'D'], tlVers: string[]=['2014', '2015', '2016', '2017']) {
```
jabooth: I would be tempted to maybe use globs to search for paths to avoid having these things hardcoded. If so, something like [`glob`](https://www.npmjs.com/package/glob) (`npm install --save glob && npm install --save-dev @types/glob`) could be used:

```typescript
import * as glob from 'glob'
 
glob("${disk}:\\TexLive\\*\bin\${command}.exe", (er, files) => {
  // files is an array of filenames. 
})
```

This will find anything matching the pattern.

src/manager.ts
```diff
@@ -218,4 +219,35 @@ export class Manager {
...
+        }
+        hasbin(configCommand, result => {
+            if (result) {
+                return
+            }
+            let found = false
+            for (const candidate of candidates) {
+                if (fs.existsSync(candidate)) {
+                    found = true
+                    vscode.window.showWarningMessage(`'latex-workshop.${config}' value '${configCommand}' is invalid. \
```
jabooth: Might be nicer here to make one of the dropdown menus (as we do for citation browser) so the user can scroll the paths and pick one.

So flow could be:

1. Warning message - the current config value is invalid, but other binaries were found. Do you want to select one of them? 

2. If yes, show the citation browser style window with text 'Select a path for the BINARY_NAME binary to replace your workspace* settings.

src/manager.ts
```diff
@@ -218,4 +219,35 @@ export class Manager {
...
+                                                      LaTeX Workshop found a possible executable file at ${candidate}. \
+                                                      Do you want to set it as 'latex-workshop.${config}'?`,
+                                                     'Yes').then(msg => {
+                        if (msg === 'Yes') {
+                            configuration.update(config, candidate, true)
```
jabooth: * from above, does this update `workspace` or `user` settings? If so should be made clear to user in messages.

package.json
```diff
@@ -154,6 +154,11 @@
         "category": "LaTeX Workshop"
       },
       {
+        "command": "latex-workshop.binaries",
```
jabooth: @James-Yu perhaps we should change this to `latex-workshop.validate_binaries` or something to make it clear we are checking something?

James-Yu:
Kinda obsolete. Close the PR for now.

PR #150, opened 2017-05-26
Trigger build with .tex file watcher

harmishhk:
Adds a file watcher to trigger builds when external program changes a `.tex` file. Proposal for this feature is in Issue #151 .

### DIFF
--- a/src/commander.ts
+++ b/src/commander.ts
@@ -14,9 +14,9 @@ export class Commander {
 
     build() {
         this.extension.logger.addLogMessage(`BUILD command invoked.`)
-        if (!vscode.window.activeTextEditor || !this.extension.manager.isTex(vscode.window.activeTextEditor.document.fileName)) {
-            return
-        }
+        // if (!vscode.window.activeTextEditor || !this.extension.manager.isTex(vscode.window.activeTextEditor.document.fileName)) {
+        //     return
+        // }
         const rootFile = this.extension.manager.findRoot()
         if (rootFile !== undefined) {
             this.extension.logger.addLogMessage(`Building root file: ${rootFile}`)
--- a/src/commander.ts
+++ b/src/commander.ts
@@ -14,9 +14,9 @@ export class Commander {
 
     build() {
         this.extension.logger.addLogMessage(`BUILD command invoked.`)
-        // if (!vscode.window.activeTextEditor || !this.extension.manager.isTex(vscode.window.activeTextEditor.document.fileName)) {
-        //     return
-        // }
+        if (!vscode.window.activeTextEditor || !this.extension.manager.isTex(vscode.window.activeTextEditor.document.fileName)) {
+            return
+        }
         const rootFile = this.extension.manager.findRoot()
         if (rootFile !== undefined) {
             this.extension.logger.addLogMessage(`Building root file: ${rootFile}`)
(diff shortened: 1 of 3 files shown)


James-Yu:
Thank you for your PR. This feature can be useful and will be merged after making the following changes
- The file extension check in `build()` should be respected. Suggested edits will be placed inline.
- A new configuration item, defaults to `false`, should be introduced to control whether auto-build on `tex` changes.

src/commander.ts
```diff
@@ -14,9 +14,9 @@ export class Commander {
 
     build() {
         this.extension.logger.addLogMessage(`BUILD command invoked.`)
-        if (!vscode.window.activeTextEditor || !this.extension.manager.isTex(vscode.window.activeTextEditor.document.fileName)) {
-            return
-        }
+        // if (!vscode.window.activeTextEditor || !this.extension.manager.isTex(vscode.window.activeTextEditor.document.fileName)) {
```
James-Yu: This file extension check should be maintained. In the future I will examine its necessity. However it should not be in this PR.

src/main.ts
```diff
@@ -152,6 +152,19 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    context.subscriptions.push(vscode.workspace.createFileSystemWatcher('**/*.tex', true, false, true).onDidChange((e: vscode.Uri) => {
+        if (vscode.window.activeTextEditor && vscode.window.activeTextEditor.document.fileName === e.fsPath) {
+            return
+        }
+        const configuration = vscode.workspace.getConfiguration('latex-workshop')
+        if (!configuration.get('latex.autoBuild.enabled') || extension.builder.disableBuildAfterSave) {
+            return
+        }
+        if (extension.manager.isTex(e.fsPath)) {
+            extension.commander.build()
```
James-Yu: Instead of calling `build()` in `commander` (which is designed to be invoked upon vscode commands), you may use the same code from https://github.com/James-Yu/LaTeX-Workshop/pull/150/files#diff-c7f5bd618dbb862973c3204edf8120d6R20 to build the file.

harmishhk:
Thanks for reviewing the PR. I was also thinking of adding the configuration item for this functionality, which I have done in the [last commit](https://github.com/James-Yu/LaTeX-Workshop/pull/150/commits/9126ac8a228de11e32ee633c23d8f2d9e2ac07c8). Please make sure you like the wording `autoBuildOnTexChange` for this configuration option which defaults to `false`. I also added a note that this configuration option only works if `autoBuild` is set to `true`.

Also, it makes sense not to touch the `build()` function of `commander.ts`, so I have added the code inline.

Regarding efficiency/performance: Just adding the `FileWatcher` does not seem to affect the overall editor performance. I observed CPU usage of `code` processes with type `watcherService` using `htop` command on `Ubuntu 16.04`. These processes does not take any CPU when idle. Only when a `TeX` file changes that trigger the build the process uses the CPU until the build is finished. I was using this FileWatcher since last three updates of LaTeX-Workshop by rebasing my `file-watcher` branch on upstream `master`, and did not notice any performance issues with the editor. I am not aware of any other profiling method that could tell me exact impact of enabling a `FileWatcher`, do you have any suggestions?

If you still think the performance can be an issue, we can check the if the configuration option `autoBuildOnTexChange` is enabled even before creating the file watcher in `main.ts`. This, however, would require the user to reload the workspace for change in this particular configuration to take effect. I personally have no problem with that option either.

James-Yu:
It seems the `filewatcher` is not causing much trouble so I will leave it as is and let the users tell.

Thanks a lot for your contribution!

PR #202, opened 2017-07-18
Add regular types from biblatex

MarkusAmshove:
Hi,

I've added some regular types for citation completion from biblatex which I was missing :-)

Taken from 2.1.1 Regular Types:
http://tug.ctan.org/macros/latex/exptl/biblatex/doc/biblatex.pdf


James-Yu:
Thanks for your contribution!

PR #203, opened 2017-07-18
Allow multiple optional parameters

MarkusAmshove:
Hello,

I've changed the autocompletion for citation to allow multiple optional parameter when using a citation function.

For example in German we include something before and after the citation in the footnote, which would look like this in the src:
```tex
\autocite[Vgl.][S.11]{somebook}
```

and would compile to:

```Vgl. Some Book S.11```

The auto completion wouldn't match it as citation because there was only one pair of square brackets allowed :-)


James-Yu:
Thanks!

PR #204, opened 2017-07-18
Basic use of the new auto-indent feature

MrYann:
Indents inside begin--end blocks (and all the preamble)


LPTK:
@MrYann @James-Yu  is there a way to disable this now? It can be quite a pain in the neck, especially when pasting code that you don't want to be automatically indented. I have to press `CMD+V` followed by `CMD+Z` (to cancel the re-indent) all the time now.

PR #205, opened 2017-07-19
Minor cleanup and overhaul to log-parser

Vogel612:
While working on a solution to the issue that error messages were not fully included in the log, I also removed some minor repetition from the preprocessing of the logging and added a reference to the repo in the readme of the addin.

The new parser now works by remembering how many empty lines follow a certain kind of entry in the sanitized log and appending lines until so many empty lines were included.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1,7 +1,7 @@
 import * as vscode from 'vscode'
 import * as path from 'path'
 
-import {Extension} from './main'
+import { Extension } from './main'
 
 const latexPattern = /^Output\swritten\son\s(.*)\s\(.*\)\.$/gm
 const latexFatalPattern = /Fatal error occurred, no output PDF file produced!/gm
@@ -15,7 +15,7 @@ const latexmkLogLatex = /^Latexmk:\sapplying\srule\s'(pdf|lua|xe)?latex'/
 const latexmkUpToDate = /^Latexmk: All targets \(.*\) are up-to-date/
 
 
-const DIAGNOSTIC_SEVERITY: {[key: string]: vscode.DiagnosticSeverity} = {
+const DIAGNOSTIC_SEVERITY: { [key: string]: vscode.DiagnosticSeverity } = {
     'typesetting': vscode.DiagnosticSeverity.Hint,
     'warning': vscode.DiagnosticSeverity.Warning,
     'error': vscode.DiagnosticSeverity.Error,
@@ -55,7 +55,7 @@ export class Parser {
         }
     }
 
-    trimLaTeXmk(log: string) : string {
+    trimLaTeXmk(log: string): string {
         log = log.replace(/(.{78}(\w|\s|\d|\\|\/))(\r\n|\n)/g, '$1')
         const lines = log.replace(/(\r\n)|\r/g, '\n').split('\n')
         let startLine = -1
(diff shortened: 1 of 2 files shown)


Vogel612:
Already submitted but (very rightfully rejected) under #205

James-Yu:
The new PR looks promising. However, can you please briefly describe why this state-based parser is better? I.e., what is the problem with the existing one? In which cases the existing parser returns wrong logs?

James-Yu:
OK I just understand the logic now. Instead of pre-concatenate the multi-line messages, this new parser actively parse the messages until empty line. Am I correct?

There are some programming style issues in the code. I will make a review and you can edit the code, or leave it as is and I will fix them after merging. In either case please leave a message 😆

src/parser.ts
```diff
@@ -78,8 +81,8 @@ export class Parser {
         }
     }
 
-    latexmkSkipped(log: string) : boolean {
-        const lines = log.replace(/(\r\n)|\r/g, '\n').split('\n')
+    latexmkSkipped(log: string): boolean {
+        const lines = log.split('\n')
```
James-Yu: I'm not an expert in LaTeX logging system. Does the compiler return `\n` instead of `\r\n` in Windows environment? An universal `\n` is highly likely, though.

src/parser.ts
```diff
@@ -88,42 +91,70 @@ export class Parser {
...
     parseLaTeX(log: string) {
-        log = log.replace(/(.{78}(\w|\s|\d|\\|\/))(\r\n|\n)/g, '$1')
         this.buildLogRaw = log
-        const lines = log.replace(/(\r\n)|\r/g, '\n').split('\n')
+        const lines = log.split('\n')
         this.buildLog = []
+
+        let remainingEmptyLines = 0
+        const emptyResult = { type: '', file: '', text: '', line: undefined }
+        let currentResult: { type: string, file: string, text: string, line: number | undefined } = emptyResult
```
James-Yu: You can remove `emptyResult` definition, and change `currentResult` like this:
```javascript
let currentResult: { type: string, file: string, text: string, line: number | undefined } = { type: '', file: '', text: '', line: undefined }
```
Then in subsequent `currentResult !== emptyResult` checks, change them to `currentResult.type !== ''`.

This is my observation, you can freely select adopting it or not.

src/parser.ts
```diff
@@ -88,42 +91,70 @@ export class Parser {
     }
 
     parseLaTeX(log: string) {
-        log = log.replace(/(.{78}(\w|\s|\d|\\|\/))(\r\n|\n)/g, '$1')
         this.buildLogRaw = log
-        const lines = log.replace(/(\r\n)|\r/g, '\n').split('\n')
+        const lines = log.split('\n')
         this.buildLog = []
+
+        let remainingEmptyLines = 0
```
James-Yu: I checked the code, and found that `remainingEmptyLines` only has two possible values: 0 and 1. Is it very familiar?

What about changing the variable to `let reachedEmptyLine = false` to make the statement more intuitive?

src/parser.ts
Vogel612: I do not know, but that doesn't matter much, given the adjustments I made in [line 50](https://github.com/James-Yu/LaTeX-Workshop/pull/205/files/b1db393f45414370fcce405fa364d29d20f90d86#diff-6518fffc4e430fcc63cd30eb07131a95R50), where the repeated replaces were centralized

PR #267, opened 2017-09-08
Effective in inline math mode and with macros contatining "@" letter.

YiHoze:


James-Yu:
Thanks for your contribution!

PR #283, opened 2017-10-02
Usage for disable auto build

### LINKED ISSUE #282: Option to disable building on save?
Can I request a feature that we have an optional which disable building on save? My XeTex is a hog which I would prefer use a shortcut key to build.

fantasticfears:
Fixes #282.


James-Yu:
I would prefer leave this PR open since I am not very sure if this should be highlighted in README as the extension is meant to be customizable via configurations. Any comments and discussion from the crowd are welcomed!

danielepanozzo:
I found this comment very useful. I am new to vscode and I did not realize that this was an available option until I found this PR.

PR #284, opened 2017-10-02
Fix CMap URL

maruta:
Fix CMap URL to display CJK languages correctly.


James-Yu:
Thank you very much for your contribution!

PR #308, opened 2017-11-15
support multi root workspace

innerlee:
this pr:

pr

master:

master

Notice that with this pr, latex outline changes when you click on files from different project.
And preview pdf button can correctly show the right pdf.


James-Yu:
Thanks for your contribution!

James-Yu:
I will need to test a bit on this change before pushing it to the public channel btw.

innerlee:
Okay 👍  I haven't test the code too much, a good review is great.

PR #309, opened 2017-11-16
fix tsc error

innerlee:
This error prevent package publish


innerlee:
ref https://github.com/Microsoft/TypeScript/issues/10642

PR #310, opened 2017-11-16
Kill all tslint warns

innerlee:
No essential change except removes all annoying warns by tslint, happy coding!


James-Yu:
Thanks for the contribution. Meanwhile, can you please branch and develop the subsequent fixes for PR? I'm always a bit hesitated to merge PRs with massive files involved, and in this case it is the base for the rest ones. What do you think?

innerlee:
I understand the sense of hesitation. However, the changes are just formatting, mostly. A careful review may be feel safer.

innerlee:
I will separate the two later prs from this one.

innerlee:
the latter two prs have been decoupled from this one, and this pr is also no longer depend on the previous one.

PR #311, opened 2017-11-16
Jump to location of current file when click view pdf button

innerlee:
It is good to let this configurable. However, make this behavior default is good in the short run.

### DIFF
--- a/src/components/viewer.ts
+++ b/src/components/viewer.ts
@@ -114,6 +114,7 @@ export class Viewer {
                         client.type = client.prevType
                     }
                 }
+                this.extension.locator.syncTeX()
                 break
             case 'close':
                 for (const key in this.clients) {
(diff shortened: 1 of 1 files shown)


innerlee:
I used this functionality these days, it goes generally good. 
However, there are times that latex sync cannot locate well, then this function will cause trouble.

So this pr can hold on, serves an issue, and wait for a better implementation.

innerlee:
Thought on new implementation: relocate the page only if the current cursor position is not visible.

innerlee:
I do not have time to implement a good version of this (may be one can ref https://github.com/izuzak/atom-pdf-view/blob/7922066915be352540e1d5eb0ac9542dd244325e/lib/pdf-editor-view.js#L224), and the current implemenation is ugly. So, it does not have meaning to keep it open :)

PR #314, opened 2017-11-16
ignore empty bib warn

innerlee:
The `` Empty `thebibliography' environment `` warning is the least informative and most annoying warn that deserves a rule to filter it out. Hope you agree :P


James-Yu:
Thanks for all your contrbutions!

PR #321, opened 2017-11-20
latex inline math completion

innerlee:
pr:
pr2

master:
master2


James-Yu:
Thanks for the contribution. Can you please introduce back the original "press enter to insert \" item as it was a request by a previous issue? I would figure this item quite necessary when ppl somehow wants a standalone `\`.

PR #348, opened 2017-12-14
Clean up dot2texi generated files

Atry:
See https://ctan.org/pkg/dot2texi


James-Yu:
There is no need to change the default inclusion list.

Issue #306, opened 2017-11-14
Can't view PDF


yzlnew:
I'm on the latest version with macOS 10.13.1.
Viewing in new tap or browser don't work for me. It shows nothing in VSCode and opens about:blank in browser.
It seems that I can't view the localhost page.
Full log:
```
[00:55:34] Found root file from active editor: /Users/r0brew/Desktop/cv-master/eivindml-cv.tex
[00:55:34] Root file remains unchanged from: /Users/r0brew/Desktop/cv-master/eivindml-cv.tex.
[00:55:34] No PDF viewer connected for /Users/r0brew/Desktop/cv-master/eivindml-cv.pdf
[00:55:34] Serving PDF file at http://::1:50851/viewer.html?file=/pdf:%2FUsers%2Fr0brew%2FDesktop%2Fcv-master%2Feivindml-cv.pdf
[00:55:34] Open PDF tab for /Users/r0brew/Desktop/cv-master/eivindml-cv.pdf
```

yzlnew:
It seems that I've found the issue.
For ipv6 address, actually the server is on http://[::1]:50851/viewer.html?file=/pdf:%2FUsers%2Fr0brew%2FDesktop%2Fcv-master%2Feivindml-cv.pdf

James-Yu:
I do not have access to a simulated environment with ipv6 support to test this out. I'll leave it as a community-help-needed bug here for now.

Claire-Jiaqi:
I also have the same problem. Have you solved it?

yzlnew:
@Claire-Jiaqi You need to edit your hosts file to manually map localhost to 127.0.0.1

Claire-Jiaqi:
Oh I find I can see the pdf in web brower but in the new tab it is black and nothing to show. If I can see in the brower, should I edit my hosts file?

yzlnew:
@Claire-Jiaqi Are you using any custom hosts settings before? And which address are you visiting through web browser?

Claire-Jiaqi:
The log:
<img width="856" alt="2018-01-01 19 22 58" src="https://user-images.githubusercontent.com/23518139/34467329-3c3d73de-ef29-11e7-8456-5e9a43f7394d.png">
I don't know about the custom hosts setting.

yzlnew:
@Claire-Jiaqi I believe your problem is nothing to do with the localhost issue.
You can create another issue.

Claire-Jiaqi:
OK,thank you!

PR #359, opened 2018-01-01
Fix missing brackets in ipv6 address for macOS

caidao22:
I had exactly the same problem as described in #306. Without the brackets, pdf file cannot be displayed on macOS. This commit fixed the problem. But I have not tested it on other platforms.


James-Yu:
Unfortunately, I cannot merge this PR since it breaks ipv4 compatibility. Can you please fix it?

caidao22:
@James-Yu OK, I have just updated the PR. Does it work now?

PR #361, opened 2018-01-02
Comments may contain latex commands that should not be read 

jsinglet:
Regex fix to ignore comments containing latex.


James-Yu:
It seems that this regex cannot recognize `(space)(space)\input{something}` with indentations. Any idea?

jsinglet:
Updated to be this: https://www.regexpal.com/?fam=99847

Not ideal, but works.

PR #416, opened 2018-02-15
[WIP] Clean auxillary files and retry once on build errors

### LINKED ISSUE #414: Suggestion: Clean auxillary files and retry on toolchain error
When the toolchain fails this can be resolved by deleting the auxillary files pretty often. Currently this is a two step process (three steps if the build is triggered manually).

It would be very useful if the workshop automatically deletes the auxillary files and tries to build once again when there is an error within the toolchain.

It should only retry once. There should also be a config option to enable (or disable) the feature.
The feature should also automatically disable when the `latex.clean.enabled` is set to true as the errors can't be caused by the auxillary files then.

schrej:
This basically works and retries one time when the build fails. The issue I'm having is that the `commander.clean()` method is asynchronous and therefore cleaning and rebuilding starts at the same time. What would be the best approach to implement a callback/promise into that function to continue when the clean is completed?

resolves #414

### DIFF
--- a/package.json
+++ b/package.json
@@ -282,6 +282,11 @@
           "default": false,
           "description": "Build LaTeX after a LaTeX source file has changed in the workspace.\nThis property defines whether LaTeX Workshop will execute the LaTeX toolchain command(s) to build the project after any LaTeX file in the workspace is saved on the disk which is not open in the active editor window."
         },
+        "latex-workshop.latex.autoBuild.cleanAndRetry.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "Delete LaTeX auxillary files when errors occur during build and retry.\nThis property defines whether LaTeX Workshop will try to clean and build the project once again after errors happen in the build toolchain."
+        },
         "latex-workshop.latex.clean.enabled": {
           "type": "boolean",
           "default": false,
--- a/package.json
+++ b/package.json
@@ -498,7 +498,6 @@
     "workspaceSymbolProvider": "true"
   },
   "dependencies": {
-    "@types/fs-extra": "^5.0.0",
     "chokidar": "^1.6.1",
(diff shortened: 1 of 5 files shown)


James-Yu:
Thanks for your contribution. Please @ me when you feel save to merge the PR.

schrej:
@James-Yu I'm currently looking for an elegant way to describe the problem I have with not knowing when `commander.clean()` is finished. I'd like to replace the `fs` Package with [`fs-extra`](https://www.npmjs.com/package/fs-extra) as that transforms all the asynchronous functions to return promises. Would you be ok with adding that package?

When all of the actions on the filesystem are promises we can collect them and create a single promise using `Promise.all()`. Whenever you don't care when it's finished you can always ignore the promise.

James-Yu:
Yes I'm perfectly fine with `fs-extra`. IIRC it is an drop-in replacement of `fs` and the original code requires no changes?

schrej:
Yes, it adds some functionality, one of which is Promises which makes it easier to wait for all the deletions.

schrej:
The `clean()` method is now returning a Promise, I tested it a few times and it seems to work fine, I wasn't able to test all the cases as I was unable to manually produce some of the errors.
The rebuild now gets triggered after the cleaning is done.

In the process I rewrote large parts of the `clean()` method and turned into a Promise chain and using the `map` and `reduce` functions of JavaScript arrays.

package.json
```diff
@@ -498,7 +498,9 @@
     "workspaceSymbolProvider": "true"
   },
   "dependencies": {
+    "@types/fs-extra": "^5.0.0",
```
James-Yu: Please move type def to devDependencies.

schrej: Of course, sorry about that

Issue #426, opened 2018-02-22
Proposal: Intellisense additions to \usepackage


Swoy:
Hello,

I've looked at the syntax plist files, but I cannot wrap my head around it. So I was hoping to get some help here;

After I have expanded \usepackage{$1} I would like to add a _second_ layer to the code suggestion with help text for all CTAN available packages. My problem is that I don't know how to do this.

I can provide *name* and *help description*/link to CTAN/usage for all packages in the format you desire, as long as I know how to write the data :)

James-Yu:
Instead of `syntax`, intellisense are coded in the extension. You may refer to `src/providers/completer` for how `\cite{}` works.

In the meantime, it is already good enough to provide the data in the form of `data/unimathsymbols.json`, which contains the symbols suggested by the extension.

Swoy:
Alright, I've made a packages.json now with 27298 lines. It has the following format:
```
    "zwpagelayout": {
	    "command": "zwpagelayout",
	    "detail": "https://ctan.org/pkg//pkg/zwpagelayout",
	    "documentation": "Page layout and crop-marks."
      },
```
Does this work?

James-Yu:
That's good enough. Can you please rename the file and place it in the corresponding folder? I am more than happy to merge your PR.

Swoy:
I'm not sure which folder and name you refer to, I apologize, I am quite new with extensions ^^
I will send you a PR as soon as I know this :)

Swoy:
Pull request added as #429

PR #429, opened 2018-02-23
Added new datafile for packagenames from #426 

Swoy:
Package names and descriptions (with links to their corresponding CTAN page).

### DIFF
--- a/data/packagenames.json
+++ b/data/packagenames.json
@@ -1,0 +1,27298 @@
+{
+"a0poster": {
+	"command": "a0poster",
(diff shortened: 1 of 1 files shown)


James-Yu:
Unfortunately, the parsed data is not a valid json file. When possible, can you please vet the content? For example, line 270 `	"documentation": "A "book" style for Addison-Wesley house style."` is not valid due to the double quotes.

James-Yu:
Currently I will remove the documentation for packages.

Swoy:
Ah, shoot! How could I miss that! I will fix this shortly, and I apologize!

PR #457, opened 2018-03-08
Add .dtx files support 

jlelong:
This PR is a proposition to add support for .dtx files. 
Basically it redefines to comment character to be `^^A` instead of `%`.
It highlights major elements in .dtx files such as
 - `%<label>`/`%</label>` lines
 - `%    \(begin|end){macrocode}` : the highlighting makes sure there are exactly four spaces between `%`and  `\` as required by `docstrip`. Anything inside the `macrocode` environment is highlighted as pure TeX content.

Any suggestion is welcome.


Stone-Zeng:
Something like this will be parsed incorrectly:

```tex
% Some text
% $a+b+c
% x+y+z$
% Some text
```

default

By the way, is it possible to support expl3-style at the same time? Maybe we need several things:

- LaTeX
- LaTeX-Expl3
- DTX
- DTX-Expl3

It's really strange...

jlelong:
I think I have managed to fix the different problems noticed by @Stone-Zeng.
I do not know yet how to implement his suggestions.

Stone-Zeng:
Some guards are still not perfect:

```tex
%    \begin{macrocode}
\def\aaa{
%<label>xyz}
abc}
%    \end{macrocode}

%    \begin{macrocode}
%<label>xyz
%    \end{macrocode}
```

default

jlelong:
Just to be sure. Your point is that a guard inside a `macrocode` environment should be ignored as it is actually a comment, right?

You wrote
> Support LaTeX outline in dtx files.

I guess this is already the case. Anything not inside a `macrocode` environment is considered as normal LaTeX without the leading `%`.

Stone-Zeng:
Actually, a dtx file has two functions:

1. Generating files. When dtx files are compiled with **tex** (or pdftex, xetex etc), all the comments (begin with `%`) will be removed, except those "guards" (`%<label>`). Guards will be detected by `docstrip` package, so they will tell tex how to generate different files.

2. Typesetting source code. When compiled with **latex** (or pdflatex, xelatex etc), the dtx file will be considered as a normal LaTeX file. With the help of `doc` package, there are several things to be noticed:

    - `%` in normal text will be ignored and `^^A` becomes the new comment sign.
    - Codes between `%    \begin{macrocode}` and `%    \end{macrocode}` (exact *four* spaces are required here) are considered as verbatim.

Therefore, in my opinion, these guards have nothing to do with the `macrocode` environment. So my idea is to highlight them uniformly, no matter they are inside or outside `macrocode` (and should *not* be considered as normal comments, of course).

Stone-Zeng:
For LaTeX outline, the panel is not available for dtx format at present.

jlelong:
Thanks for your detailed comment. I have fixed guards highlighting, at least hope so.

So far, I have decided to consider `%` as part of the matching pattern in `%<label>` and `\\(begin|end){macrocode}`, which makes `%` highlighted as the rest of the pattern. I am wondering whether this is smart? I could choose to highlight all the `%` uniformly in the whole file.
Any suggestion on that?

jlelong:
> For LaTeX outline, the panel is not available for dtx format at present.

I cannot figure out how to launch the extension for `.dtx` files. However, if I also open a `.tex` file, which actually launches the extension, then I can access the panel from my `dtx` file and use it to compile and view the documentation part of the `dtx` file.

Any suggestion?

Stone-Zeng:
@jlelong If comments in `macrocode` were put in group, then the highlight is still no correct:

```tex

%    \begin{macrocode}
\begin{equation}
abc
% abc
^^A abc
\end{equation}
%    \end{macrocode}

%    \begin{macrocode}
{abc
% abc
^^A abc
}
%    \end{macrocode}

%    \begin{macrocode}
$
abc
% abc
^^A abc
$
%    \end{macrocode}
```

default

Here, text with `%` should be considered as comment, while text with `^^A` should not.

jlelong:
@James-Yu thanks for the tip. Now, the extension is activated for `.dtx` files.

jlelong:
@Stone-Zeng, I have managed to fix highlighting inside groups (the problem was not restricted to comments). I am confident it will finally make it 😝

Finally, outside `macrocode`, `%` is only highlighted when in first position.

Stone-Zeng:
Consider the following things: (:satisfied:)

1. `%` that is not at the beginning:
```tex
%    \begin{macrocode}
{abc
% xyz
abc %xyz
\abc
\abc %xyz
}
%    \end{macrocode}
```



2. Containing `\chapter` etc may break all the syntax:

```tex
%    \begin{macrocode}
  {
    \chapter*{
      {\abc}}%
  }
%    \end{macrocode}

%    \begin{macrocode}
Something afterwards
%    \end{macrocode}
^^A abc
% xyz
```

jlelong:
@Stone-Zeng, you are an impressive beta tester 😛
It has been fixed by considering everything inside `macrocode` as pure TeX.

Stone-Zeng:
This time let's turn to something outside `macrocode`:

```tex
%^^A help%
```

image

jlelong:
I have refactored a bit how to handle `^^A` and `%`. Here is an overview of the choices I implemented
capture d ecran 2018-03-13 a 15 33 45

Is this what you expected?

Stone-Zeng:
It seems that everything goes right now!

jlelong:
Thanks for your help identifying the problems!

James-Yu:
You guys really set a benchmark for community collaborations! More than happy to merge the PR.

jlelong:
I agree with @Stone-Zeng:`.ltx` is  standard LaTeX. 
I am making a new PR #466 as I cannot reopen this one.

PR #459, opened 2018-03-10
make arguments for latexindent configurable

project-pp:
Hello, I'm using LaTeX-Workshop everyday!! 
I wrote new option ”latex-workshop.latexindent.args”, which make arguments for latexindent configurable. I think it is minor problem but it would improve this extension.
regards,

### DIFF
--- a/package.json
+++ b/package.json
@@ -442,11 +442,11 @@
           "type": "array",
           "default": [
               "-c",
-              "%DIR%",
+              "%DIR%/",
               "%TMPFILE%",
               "-y=\"defaultIndent: '%INDENT%'\""
           ],
-          "description": "Define the command line arguments for latexindent."
+          "description": "Define the command line arguments for latexindent.\nNote: At this moment -c option requires trailing slash."
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -437,6 +437,11 @@
           "type": "string",
           "default": "latexindent",
           "description": "Define the location of the latexindent executable file."
+        },
+        "latex-workshop.latexindent.args": {
+          "type": "string",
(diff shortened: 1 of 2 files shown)


James-Yu:
Thanks for your contribution. It will absolutely improve the extension.

In the meantime, can you please modify it to obey the common `args` pattern for other tools, i.e., list of string instead of a plain string? This is more a consistency consideration. In addition, `latexindent` should be `spawn`-ed instead of `exec`-ed by `child_process`, on which I will work later (PR appreciated!).

project-pp:
I agree, it mus be `string[]`, but how is the `stdio` option of `spwan`?
I mean, at this moment `latexformatter.ts` ignores stderr, I thought it would be better to handle and post it via `addLogMessage`.

project-pp:
I agree, it should be `string[]`.
The next commit implements both stdout and stderr handler.
It seems to work for me:smile:

project-pp:
Note for about 966ee31:
At this moment `-c` option of `latexindent ` requires a path with trailing slash.
c.f. https://github.com/cmhughes/latexindent.pl/blob/2cf5ae5cb57017fe383ebe1a8420f95abb63f95c/LatexIndent/LogFile.pm#L97
I just add it to `latex-workshop.latexindent.args` (so anyone can remove when it is no longer needed).

James-Yu:
Nice job! One last request: can you please add a brief description to `%TMPFILE%` and `%INDENT%` to the description of the config item? They are nowhere else defined.

project-pp:
Exactly.
I think it would be better to add information about `latex-workshop.latexindent.*` into README (but i'm not good at english...pls can you or can anyone later...?:innocent:)

James-Yu:
Nor am I a native speaker but I'd like to try. Thank you again for your contribution!

Issue #102, opened 2017-04-10
Proposal - help users keep magic comments in order?  Or set root file config? Or both?


jabooth:
Following the discussion in #101 I got thinking about the best way latex workshop can help users choose root files.

I can see a few ways to take this:

#### Provide tools to help users manage magic comments

Magic comment is really the 'standard' way to choose the root file - it works across all latex tools. latex workshop could focus on helping users set magic comments and keep them up to date. This would be nice as LW would be encouraging a best practice, and users would be able to benefit from this when they use their tex files anywhere outside LW.

Some basic ideas of what could be done here:

1. If there is an ambiguous root, LW could show a warning, and ask users to choose the correct root file **for the current document**. After choosing, LW could insert the relevant magic comment at the top of the current file.

2. We could periodically check that all .tex files referenced from the current root document have a correct magic comment. LW could automatically show a problem when magic comments are missing or incorrect, and offer a code action to fix them. As they are fairly simple, we could have a command to run to fix all magic comments for the current root. Optionally, users could ask LW to automatically provide these fixes, so as they move files around LW works to ensure the magic file is always correctly set.

3. Of course, all this could be disabled by config.

#### Add a workspace setting for setting the root.

This is fine too, and is very easy to understand. However it adds another way the root file can be chosen (Another thing for users to debug if LW isn't behaving as they expect with root files) and is pretty hopeless if there are two latex documents in one workspace (which can be quite common with papers + supplimentary material).

skamoen:
If the magic comment is the 'standard way' (I really wouldn't know), I think that's a better option than a custom config setting. 

I didn't know about the magic comment until I ran into trouble which proved not very trivial to debug, so the magic comment proposal ideas would make the whole root document selection a lot more transparent :+1:

jabooth:
Hmm it's funny @skamoen because I've seen a number of tools use these comments over the years and just figured it was 'official' - I don't think it is, but they seem to have quite broad support, more info here:

http://tex.stackexchange.com/questions/78101/when-and-why-should-i-use-tex-ts-program-and-tex-encoding

jabooth:
Key bit from above link:

**Editors that understand % ! TeX directives**
```
              E P R S
TeXShop       x x x x
TeXStudio     x x x x
TextMate      ? x x ?
TeXworks      x x x x
SublimeText   o x x x
Atom          o x x o
Vim (vimtex)  o x x o
              | | | |
x = yes       | | | Spellcheck
[...]

PR #467, opened 2018-03-14
Added "add root document" command

oerpli:
I've implemented a command that allows to add a root document to the current document.
This is done by specifying another .tex file from the current workspace, calculating the relative path and inserting the proper command in the first line of the active document. 

If there is already a tex root command there this is ignored and the new command gets added in the line above (always in line 0).

I think this additional feature would not work for some users (depending on the toolchain, etc), but they can just ignore this command and live happily until a more elaborate system is in place (related  issue:  #102 ).
If a more feature complete system is planned/implemented in the future it might be advisable to refactor this code to a component. I thought it would be fine to just add the function in commander.ts until that is the case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -213,7 +213,7 @@
         "category": "LaTeX Workshop"
       },
       {
-        "command": "latex-workshop.addroot",
+        "command": "latex-workshop.addtexroot",
         "title": "Insert %!TeX root magic command",
         "category": "LaTeX Workshop"
       },
--- a/package.json
+++ b/package.json
@@ -214,7 +214,7 @@
       },
       {
         "command": "latex-workshop.addroot",
-        "title": "Insert root document",
+        "title": "Insert %!TeX root magic command",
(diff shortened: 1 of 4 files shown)


oerpli:
I've addressed your two demands and also added a isTeX check to only execute the command if the current file is a TeX file (copied from `build()` command)

Additionally I would like to raise a small concern: I'm not entirely sure, if it is possible to have filepaths that break my algorithm for creating the relative path. Possible problems I can currently think of:

- Adding TeX Root to document that is not part of the current project (no idea if it is necessary to handle this case)
- Differences between windows/linux filepaths (I assumed that folders/files don't contain `/` and `\` and have a length of at least 1.  I don't know if this always holds.

oerpli:
I think I've now implemented and tested all things that I would expect from a feature like this. 

If other users/developers have use cases that should also be supported I could implement them in the next days.

Issue #461, opened 2018-03-12
Enhancement: unify some of the formats


Stone-Zeng:
There are now several formats supported in this extension:

- tex
- latex
- latex-expl3
- latex-beamer
- latex-memoir
- dtx (under development)

and some other things (bib and log).

Personally, I don't think `beamer` and `memoir` should be consider as separate formats, since they still use the normal LaTeX syntax. But dtx should be separated, since the syntax is almost totally different and the extension is different as well (`.dtx` rather than `.tex`).

However, things are a little bit complex for expl3. Basically, LaTeX3/expl3 is a "new" language. But it is compatible with the traditional LaTeX2e. In one words, I can write LaTeX2e materials in a LaTeX3 file, and using LaTeX3 in LaTeX2e is also OK if put in the `\ExplSyntaxOn` ... `\ExplSyntaxOff`.

Therefore, what I want to say is that, we can merely use two format: `latex` and `dtx`, but they should support LaTeX3/expl3 syntax at the same time. `tex`, `latex`, `latex-beamer` and `latex-memoir`, in my opinion, can be merged into a single format.

Thank you very much!

jlelong:
I agree that `latex-memoire` and `latex-beamer` should be part of `latex`.

Because of all the `\begin ... \end` constructions in `latex`, I would rather keep a pure `tex` syntax separated from `latex` to make `tex` reusable.  
For instance in `dtx`, you can have unterminated groups ended by `\end{macrocode}`, whereas overlapping groups are LaTeX errors.  In order to make the `latex` syntax compatible with this,  everything dealing with groups has to be redefined, which means a lot.

jlelong:
@James-Yu, if you think it is a good idea  I can merge `latex-beamer` into `latex` as a first step.

James-Yu:
@jlelong Yes it would be fine to combine some, though I'm hesitated on `tex` and `latex`.

jlelong:
I do think we need to separate `tex` and `latex`. This is the case for most editors.
Vim through its `vim-latex` provides a `plaintex` mode and a `latex` mode. The same holds for Emacs with Auctex. Actually, plain `tex` files can contain unbalanced `latex` stuff inside `\def` for instance , which the main reason why both must keep separated.

Stone-Zeng:
Maybe the format `TeX` should be called `plain TeX`, to be precise.

[...]

PR #468, opened 2018-03-15
Match all braces delimiting arguments uniformly

jlelong:
Wherever '{...}' is used as an argument of a command but not as a group,
the braces are matched as
    `punctuation.definition.arguments.begin.latex`
    `punctuation.definition.arguments.end.latex`
We get rid of the scope `punctuation.definition.arguments.latex`

PR related to https://github.com/James-Yu/LaTeX-Workshop/issues/461#issuecomment-373165202


James-Yu:
What a syntax guru!

Issue #464, opened 2018-03-12
Additional tool chain process: latex + dvisvgm


sabicalija:
### Description

I would like to add an additional build process to latex workshop, that I could use globally when working on latex files.

Today I figured out, that I can create svg with latex + dvisvgm.
I want those two steps when working with TikZ, globally available in vscode.

For instance, suppose following tex file:

`drawing.tex`
```
\documentclass{standalone}
\usepackage{tikz}

\begin{document}
   \begin{tikzpicture}
      \draw (0,0) rectangle ++(3,2);
   \end{tikzpicture}
\end{document}
```
With following two steps, one can create an svg out of this tex.

```
$ latex drawing.tex
$ dvisvgm drawing.dvi
```

The result is an svg named `drawing.svg`.

Is it possible to add additional build processes and instruct Latex Workshop to select one of those available, maybe per file extension, or with "Magic comments"??

Would be great if one could customize the steps necessary.

James-Yu:
It's a popular request. However, there's no current plan to implement it. When you have two toolchains, people will ask for more.

A good alternative is [vscode tasks](https://code.visualstudio.com/docs/editor/tasks). You can define a series of command in a task. I have no previous experience in it but I believe it fits your request.

sabicalija:
Thanks for the tips. Anyway, I do understand your decision.
I was hoping someone will know how to deal with it, or it would inspire someone to implement such a feature. :crossed_fingers: 

However, I've setup a tasks which does exactly what I was describing above.
The main problem with it was, that those tasks can't be configured globally, unfortunately.

Actually, I had following approach in mind:

![image](https://user-images.githubusercontent.com/13609768/37375469-0ea0ab60-2720-11e8-8bfc-3a13be7949e4.png)

My suggestion is to create an additional "recipes" entry, which allows to select from a list of such "recipes". A user would than configure them similarly to the "toolchain" customization, but within an _array_.

For instance:

```
"latex-workshop.latex.recipes" : [
  {
    "name": "latex",
    "commands": [
      {
        "command": "latex",
        "args": [
          "-synctex=1",
          "-interaction=nonstopmode",
          "-output-directory=",
          "%DOC%"
        ]
      }
    ]
  },
  {
    "name": "latex + bibtex + latex",
    "commands": [
      {
        "command": "latex",
        "args": [
          "-synctex=1",
          "-interaction=nonstopmode",
          "-output-directory=",
          "%DOC%"
        ]
      },
      {
[...]

PR #477, opened 2018-03-20
New build toolchain system; Recipes

project-pp:
From #464 
The current implementation is a prototype. I'd like to hear opinions from all of you.
Possible improvements:

- Check the backward-compatibility
- More maintainability (the current implementation is hardly mixed with old toolchain system)
- Extensibility (I proposed `do`; recipe reference. But there may be more flexible/useful feauture)

Thanks.

### DIFF
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -120,10 +120,10 @@ export class Builder {
         }
     }
 
+    static readonly RECIPE_KEY_IMPORT: string = 'Invoke latex.toolchain'  // special key; it would import toolchain from 'latex-workship.latex.toolchain'
+
     getRecipeKeys() : string[] {
-        const result = [
-            'latex-workshop' // special key; it would import toolchain from 'latex-workship.latex.toolchain'
-        ]
+        const result = [ Builder.RECIPE_KEY_IMPORT ]
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
         const recipes = configuration.get('latex.recipes')
         if (typeof recipes !== 'undefined') {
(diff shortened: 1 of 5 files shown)


James-Yu:
I've checked the implementation. It is a good attempt, though some improvements can be made. Do you mind my committing to the PR and merge it later?

project-pp:
Sure. At the moment my code is not a production level. :pray:

James-Yu:
Some preliminary tests yield good results. Please have a test and see if things work well.

src/components/builder.ts
```diff
@@ -126,7 +126,7 @@ export class Builder {
                 args: configuration.get('latex.magiccommand.args') as string[]
             }]
         } else {
-            const recipes: {name: string, tools: string[]}[] = configuration.get('latex.recipes') as {name: string, tools: string[]}[]
+            const recipes = configuration.get('latex.recipes') as {name: string, tools: (string | StepCommand)[]}[]
```
James-Yu: Can you please elaborate on why `tools` field can be a string?

project-pp: I think it would be better to allow *raw* StepCommand as `latex-workshop.latex.recipes[n].tools` element (it is allowed in my original implementation). If you don't want to do this, I'll rollback it (and also delete line.156).

James-Yu: I would suggest a uniformed definition of `tools`. This can be somewhat annoying at first, but multi-recipe users should be happy to see a tidy configuration file.

James-Yu: Maybe let it be a hidden undocumented feature would work well.

project-pp: Of course I think it is not a clean solution.
Pls remove it later when you think it is unnecessary or someone implements good alternatives.

project-pp:
It seems good. At this moment I have no more commits.
For non-expert user, you might want to add a little migration guide of toolchain->recipes.:thumbsup:

PR #505, opened 2018-04-01
Fix #504 Recipe being ignored in favor of magic comment // wrong behavior for magic comment

### LINKED ISSUE #504: Recipe being ignored in favor of magic comment // wrong behavior for magic comment
I've setup a toolchain and recipe to build my projects using xelatex->biber->xelatex->xelatex.
However, due to the presence of % !TeX program = xelatex, the toolchain is ignored.
There are two issues here, at the same time:

1) Even if I run the "Build with recipe" task, my recipe is not run due to the magic comment.
2) The behavior of the magic comment toolchain is, I think, wrong. It should do a 4-pass compilation, as most tools need that. First to generate aux files, second to build bibliography, third to get cross-ref right and fourth to generate the PDF with everything in place. Or am I wrong here? It seems to work like that for xelatex, lualatex and pdflatex.

The way it is now I'm not getting references nor bibliography in the output pdf.
All works if I remove the magic comment (but I need it).

Proposal: make the default toolchain for magic comments be
(!TeX program) -> (!BIB program) -> (!TeX program) -> (!TeX program)

acristoffers:
This should fix both issues, so selecting a recipe in menu should use that recipe and ignore the magic comment, while the build command should now use the magic command present making a 4 pass using also the !BIB program magic comment. I had to insert a new configuration for this, for the bib command args. Default bib program, in case no magic comment is found, is bibtex, as it seems to be the default in other tex applications.

### DIFF
--- a/package.json
+++ b/package.json
@@ -294,6 +294,13 @@
           ],
           "description": "Define the arguments to be input to magic command executable."
         },
+        "latex-workshop.latex.magic.bib.args": {
+          "type": "array",
+          "default": [
+            "%DOCFILE%"
+          ],
+          "description": "Define the arguments to be input to BIB magic command executable."
+        },
         "latex-workshop.latex.outputDir": {
           "type": "string",
           "default": "./",
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -31,12 +31,12 @@ export class Builder {
     }
 
     buildInitiator(rootFile: string, recipe: string | undefined = undefined) {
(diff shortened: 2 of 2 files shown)


acristoffers:
Duplicating the find-magic method was stupid, as it loads the file twice. But now I can't make another pull request until you accept this one, so I'll just wait and send the other one right after.

James-Yu:
More than happy to merge the PR!

PR #521, opened 2018-04-07
Fix #519

### LINKED ISSUE #519: Latest release has broken my compile
### Description

OK, this is weird and I'd _really_ appreciate some help figuring this out.

My document compiled fine with the previous release, but the 5.0.4 release now results in errors like this:

```
This is BibTeX, Version 0.99d (MiKTeX 2.9.6500 64-bit)
The top-level auxiliary file: main.aux
A level-1 auxiliary file: Bla.aux
A level-1 auxiliary file: Bla2.aux
I found no \citation commands---while reading file main.aux
I found no \bibdata command---while reading file main.aux
I found no \bibstyle command---while reading file main.aux
```

### How to reproduce

I reduced my document down to this and could still reproduce, even after a clean:

```
% !TEX program = xelatex

\documentclass{book}

\begin{document}

Foo

\end{document}
```

So I copied this to a completely fresh `TEMP` directory, and it compiled fine - I cited the PDF and everything. I therefore tried a `git clean -xdf` in my original project to be certain there were no residual generated files, and it _still_ fails to compile.

On a hunch, I tried copying the subdirectories of my project into my `TEMP` directory. I cleaned and compiled and...it failed! I immediately removed the directories I'd just copied, did a clean so I had only the original tex file, then compiled again. It _STILL_ failed!!

Utterly confounded at this point, I tried the whole thing again, this time in a `TEMP2` directory. It failed right off the bat this time. But I check the output and the PDF _is_ there.

I went back to my original project and rebuilt it. It failed with the same error, but I found the PDF in the output directory. However, it was really messed up - stuff missing or clipped/not sized correctly.

I tried one more clean+build, but the result was the same :(

### Expected behavior

Document should still compile without error.

### Actual behavior

Document fails to compile with above error, and it's not at all clear what triggers it or how to rectify.

#### LaTeX Workshop Output

```
This is BibTeX, Version 0.99d (MiKTeX 2.9.6500 64-bit)
The top-level auxiliary file: main.aux
I found no \citation commands---while reading file main.aux
I found no \bibdata command---while reading file main.aux
I found no \bibstyle command---while reading file main.aux
(There were 3 error messages)
```

#### Developer Tools Console

I can't see anything suspicious.

### Additional Information

Running on Windows 10.

acristoffers:
Only run the bib phase if there's a bib magic command. I didn't think about simple non-bib projects when writting, as most of my documents are academic and require references. Sorry about that.

### DIFF
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -142,7 +142,11 @@ export class Builder {
                 command: magicBib,
                 args: configuration.get('latex.magic.bib.args') as string[]
             }
-            steps = [magicTexStep, magicBibStep, magicTexStep, magicTexStep]
+            if (magicBib === '') {
+                steps = [magicTexStep, magicTexStep, magicTexStep]
+            } else {
+                steps = [magicTexStep, magicBibStep, magicTexStep, magicTexStep]
+            }
         } else {
             const recipes = configuration.get('latex.recipes') as {name: string, tools: (string | StepCommand)[]}[]
             const tools = configuration.get('latex.tools') as StepCommand[]
@@ -191,7 +195,7 @@ export class Builder {
         const tex = content.match(regexTex)
         const bib = content.match(regexBib)
         let texProgram = ''
-        let bibProgram = 'bibtex'
+        let bibProgram = ''
 
         if (tex) {
             texProgram = tex[1]
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -142,10 +142,10 @@ export class Builder {
                 command: magicBib,
                 args: configuration.get('latex.magic.bib.args') as string[]
             }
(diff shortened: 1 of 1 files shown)


src/components/builder.ts
```diff
@@ -142,7 +142,11 @@ export class Builder {
                 command: magicBib,
                 args: configuration.get('latex.magic.bib.args') as string[]
             }
-            steps = [magicTexStep, magicBibStep, magicTexStep, magicTexStep]
+            if (magicBib === '') {
```
wopian: This is more readable if the logic is flipped to:

```js
if (magicBib) {
	steps = [magicTexStep, magicBibStep...]
} else {
	steps = [magicTexStep, magicTexStep...]
}
```

Issue #569, opened 2018-04-23
Adding language support for LaTeX Expl3 and DocTeX


Liam0205:
Issue Type: <b>Feature Request</b>

I'm a TeX/LaTeX macro developer.

When developing with LaTeX Expl3 and the DocTeX util, I find that LaTeX workshop for VSCode could not display code highlighting well. Please consider adding the support for the combination of them.

Extension version: 5.3.0
VS Code version: Code 1.22.2 (3aeede733d9a3098f7b4bdc1f66b63b0f48c1ef9, 2018-04-12T16:32:53.389Z)
OS version: Darwin x64 16.7.0

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-4278U CPU @ 2.60GHz (4 x 2600)|
|Load (avg)|2, 2, 2|
|Memory (System)|8.00GB (0.03GB free)|
|Process Argv|/Applications/Visual Studio Code.app/Contents/MacOS/Electron|
|Screen Reader|no|
|VM|0%|

</details>

jlelong:
Can you be more precise on what you expect or even give us an example?

Liam0205:
Sorry for anything that is not precise enough.

In LaTeX, `%` is used for comment. It works like `#` in Python. However, in DocTeX mode, the leading `%`s will be striped for literate programming. That is to say, in DocTeX mode, comments behind the leading `%`s forms the documentation of those code are not being commented. This is why VSCode highlights contents behind leading `%`s in DocTeX mode.

<img width="1280" alt="doctex mode" src="https://user-images.githubusercontent.com/3348485/39160885-783b2960-47a0-11e8-94e5-f57c9df0fe4f.png">

The commands in LaTeX are begun with the escape char `\` and followed by a uninterrupted sequence of English letters (called control word) or followed by a single sign (called control sign). In LaTeX2e mode, LaTeX expl3 commands like `\@@_alternate_cs:n` are invalid in most time, hence, in most LaTeX highlighting extensions, including DocTeX mode in VSCode, this kind of commands will not be highlighted correctly. In short, LaTeX expl3 allows `@`, `_`, and `:` in valid commands, while LaTeX2e does not.

[...]

PR #582, opened 2018-04-26
Enable latex3 syntax in doctex mode

jlelong:
This is related to the original demand in #569.

### DIFF
--- a/syntax/DocTeX.plist
+++ b/syntax/DocTeX.plist
@@ -64,7 +64,7 @@
 				</dict>
 				<dict>
 					<key>include</key>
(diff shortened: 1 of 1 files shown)


syntax/DocTeX.plist
```diff
@@ -64,7 +64,7 @@
 				</dict>
 				<dict>
 					<key>include</key>
-					<string>text.tex</string>
+					<string>text.tex.latex.expl3</string>
```
Liam0205: It seems to be a clever workaround. Thanks @jlelong .

PR #604, opened 2018-05-06
Ignore " in Latex syntax highlighting.

### LINKED ISSUE #593: Syntax highlight for quotation marks
### Description

In TeX, a quotation mark can be used to represent a hex number, so it should not be highlighted as an emphasis environment.

For example, in this figure, `"24B6`, `"24D0` and `"2460` should not be highlighted.

image

### Additional Information

This behavior exists in LaTeX, LaTeX expl3 and DocTeX mode.

jlelong:
Along with 1dcb0d7, it closes #593.


jlelong:
I have just reverted part of the commit to allow `"` to close ` `` `.

PR #618, opened 2018-05-10
Add snippets with no leading \

jlelong:
Comments are welcome


James-Yu:
Let's do it and see the response!

PR #684, opened 2018-07-01
support multiple root file

innerlee:
This is somewhat cpu intensive. Should test for performance before merge.

With this pr, you can have multiple root files in one project. The root file do not have to be in project root folder.


innerlee:
actually, after some optimization, it does not add that much burden.

PR #693, opened 2018-07-09
fix open pdf view tab

innerlee:


innerlee:
@James-Yu this time, it should work :P

innerlee:
just in case, pls test locally before publish...

Issue #658, opened 2018-06-06
Support package `import` in the outline


florianrusch:
### Description
It would be nice, if LaTeX-Workshop would support the package `import`. With import you have the ability to include graphics in subfiles via there relative path.

Example folder structure:
- `main.tex`
- `chapter01/`
  - `subfile.tex`
  - `graphics/`
    - `graphic.jpg`

Example `main.tex`:
```tex
...
\begin{document}
	\section{Section 01}
	\subimport{./chapter01/}{subfile}
\end{document}
```

Example `subfile.tex`:
```tex
\subsection{Subfile}
\begin{figure}[ht]
	\includegraphics{./graphics/graphic.jpg}
\end{figure}
```

**Concrete** would it be nice if the sections from a subfile, which is included via `subimport`, will appear in the outline.

Currently only `Section 01` will appear in the outline.

jlelong:
If we do not aim at supporting nested subimport, I think it can be done fairly easily.

Supporting nested subimport is far more difficult and requires to redesign how files are discovered inside a LaTeX project.

I am not using the `subimport` package, is nesting subimport a typical usage?

florianrusch:
I think the first layer would be enough. I haven't seen yet any example where an `\import or \subimport` was used in an imported file.

jlelong:
Just added. Let me know if there is any trouble.

florianrusch:
Awesome! I'll try it later this day or tomorrow.
Thank you for you fast support! 👍

vogler:
For me this does not work.
Subsections via `\include` show up in the 'structure' view, but via `\subimport` and `\subimport*` don't.
`main.tex`:
~~~
\section{Foo} % shown
\subimport{sub/}{intro}  % not shown
\include{sub/intro} % shown
~~~
`sub/intro.tex`:
~~~
\subsection{Introduction}
~~~

jlelong:
See the new issue #1827.

PR #705, opened 2018-07-15
Change subimport command to import for the outline

florianrusch:
The command subimport should be just used to import on the second level, which isn't supported as in #658 discussed.


jlelong:
The `import` package provides the two commands: `\import{<full_path>}{<file>}` and `\subimport{<rel_path>}{<file>}`. From the package manual, the only difference between the `\import` and `\subimport` commands seems to be absolute vs relative path.

As of now, only `\subimport` is understood by the outline feature. The next step is to do the same for the `\import` command. Please modify your PR in order to add support for both commands.

Issue #668, opened 2018-06-15
Issues with Syntax Highlighting in Math Mode


bnord01:
### Description

Inside math mode there are several issues with syntax highlighting. 

- [x] Subscripts break the keyword highlighting (see `\sigma_{i}`below, superscripts work).
- [x] Opening curly braces followd by punctuation characters like `(),.` are highlighted differently (see `\{(a,b)\}` below).
- [x] Closing textmath `\( \)` paranthesis followed by punctuation characters like `(),.` inside `\text{}` are highlighted differently (see `\text{Have \(\sigma\),}` below).
- [x] Newlines (`\\`) inside math tables are highligned differently depending upon beeing followed by brackets for optional parameters. In general I'd prefer layout affecting macros like `\\, \newline, \linebreak, \pagebreak, \clearpage` to have their own distinctive markup like `&`.  (see `\\[1ex]` and `\\` in the `aligned` environment below)

### How to reproduce

Example
![image](https://user-images.githubusercontent.com/1318553/41468298-ce5ad41c-70a9-11e8-8d9f-b98dff5218da.png)

```latex
Hello \\ World! \\[1ex] Have \(\{(a,b)\}\), \(y\).
\[
	\begin{aligned}		
		Foo   & Bar, \\[1ex]
		42    & \text{Have \(\sigma\),}\\
		\sigma_{i} & \sigma^{i}
	\end{aligned}	
\]
```

### Expected behavior

- Consistent keyword highlighting with and without subscripts
- Consistent highlighting of curly braces
- Consistent hightlighting of math mode delimters indepentent of following punctuation characters
- Consistent hightlighting of newline macros

### Actual behavior

- Incosistent highlighting.

### Additional Information

Code Version 1.24.1 
LaTeX Workshop Version 5.5.1
macOS

jlelong:
Syntax highlighting is quite complicated. The syntax file for LaTeX is very large and tricky, which makes it very hard to test. Therefore, I do not feel like changing it too much in depth. Anyway, I will do my best to try to fix the inconsistencies you pointed out if it does not involve too many changes.

jlelong:
I think I have managed to fix everything although it is still not fully satisfying for `\text{...}` and `\mbox{...}` whose content should not be highlighted as maths. This would require to capture the content of the curly braces, which caused too many troubles so we decided to drop it a while ago.

Let me know if there is still something wrong.

PR #822, opened 2018-09-25
highlighting \} followed by non-whitespace.

tamuratak:
Hi,

Highlighting \\} followed by non-whitespace does not work well in the following example:
```
$\{\}$
$\{a\}+b$
```

This patch fixes the problem. Even without the removed part of regex, an example in #668 is highlighted well. I cannot find conditions under which the removed part is required. Please check.

Regards,

### DIFF
--- a/syntax/TeX.plist
+++ b/syntax/TeX.plist
@@ -332,7 +332,7 @@
 						</dict>
 					</dict>
 					<key>match</key>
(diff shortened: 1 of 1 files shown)


jlelong:
I guess the following regex would do the job
```
<string>(\\)(?!begin\{)([A-Za-z]+)</string>
```
No need of `[^a-zA-Z]`. Do you confirm?

tamuratak:
Yes, I confirm that. With your regex,
```
<string>(\\)(?!begin\{)([A-Za-z]+)</string>
```
\\}, \\%, and others are highlighted as constant.character.escape.tex everywhere, which, I agree, is more consistent.

PR #824, opened 2018-09-26
refactor regex.

tamuratak:
Hi,
this is minor refactoring.
\\s* matches zero-length string immediately after the beginning of a line. So we can drop ^ .

Regards,

### DIFF
--- a/syntax/LaTeX.plist
+++ b/syntax/LaTeX.plist
@@ -715,7 +715,7 @@
 		<dict>
 			<key>begin</key>
 			<string>(?x)
-					(?:^|(?:\s*))										# Optional whitespace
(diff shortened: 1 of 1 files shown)


jlelong:
I agree with you, but I think there might be a slight difference. In one case the beginning of line character is included in the match, which may be important as there is a notion a precedence in the order of matching regex in syntax files. I have spent so much time working on the syntax files to understand why highlighting was not working properly in some very specific cases that I am always a bit suspicious with all this.

If you do not mind, I will not merge the PR now but keep it in mind and do so more advanced testing.

tamuratak:
No, I don't mind. Thanks.

Issue #809, opened 2018-09-19
Keybindings


tecosaur:
# Background

Keybindings are great, we all love them, and LaTeX workshop currently doesn't have any. While this may not be as significant a change as other additions, after discovering [this](https://ftp.gnu.org/gnu/auctex/11.88-extra/tex-ref.pdf) since it seemed reasonably easy to implement (for the most part) I thought why not add it to the LaTeX workshop!
I found that `Ctrl`+`L` seems to be unused, why not use that instead of `C-c` from the AUCTEX reference card.

# Implementation Progress

This is a modified version of the AUCTEX reference card, some things removed, some things added.

### Issue encountered

- VS Code only take 2 part key bindings, not 3.

### _Command Insertion_

| Shortcut      | Does                     | Implemented |
| ------------- | ------------------------ | :---------: |
| `C-l C-s`     | Insert Section           |  Need Help  |
| `C-l C-e`     | Insert LaTeX Environment |  Need Help  |
| `C-l C-Enter` | Insert `\item`           |      ✓      |
| `M-Enter` | Insert `\item`           |      ✗      |
| `C-l Escape`  | Close LaTeX Environment  |  Need Help  |
| `C-l C-.` | Insert `\@.` | ✗ |
| `C-l C-space` | Insert <code>\\&nbsp;</code> | ✗ |

### _Font Selection_

| Shortcut Opt#1 | Shortcut Opt#2 | Does            | Implemented |
| -------------- | -------------- | --------------- | :---------: |
| `C-b`          | `C-l C-b`      | Bold text       |      ✓      |
| `C-i`          | `C-l C-i`      | Italic Text     |      ✓      |
| `C-u`          | `C-l C-u`      | Underline       |      ✓      |
| `C-r`          | `C-l C-r`      | Roman Text      |      ✓      |
| `C-e`          | `C-l C-e`      | Emphasised Text |      ✓      |
| `C-t`          | `C-l C-t`      | Typewriter Text |      ✓      |
| `C-s`          | `C-l C-s`      | Slanted Text    |      ✓      |
|                | `C-l C-c`      | Small Caps Text |      ✓      |
|                | `C-l C-l`      | Lowercase       |      ✓      |
|                | `C-l C-u`      | Uppercase       |      ✓      |
|                |                | Delete Font     |  Need help  |
|                |                | Replace Font    |  Need help  |

I'm currently using Opt#1 for BIU and emphasis when text is selected, then Opt#2 for everything else

[...]

PR #831, opened 2018-09-29
The afformentioned keybindings

tecosaur:
Here it is at last! Have some more ideas but here's V1 at least. 
Edit: For the sake of referencing this stuff comes from #809.

### DIFF
--- a/package.json
+++ b/package.json
@@ -285,189 +285,212 @@
         "key": "ctrl+alt+x"
       },
       {
-				"key": "ctrl+l ctrl+enter",
-				"command": "type",
-				"when": "editorTextFocus && editorLangId == 'latex'",
-				"args": {
-					"text": "\n\\item "
-				}
-			},
-			{
-				"key": "ctrl+b",
-				"when": "editorHasSelection && editorLangId == 'latex'",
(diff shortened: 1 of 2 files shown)


Astrantia:
@tecosaur I would suggest to add readonly's too:
https://github.com/neilsustc/vscode-markdown/blob/master/package.json#L60

tecosaur:
That seems like a decent idea. I'll do that.

jlelong:
The following keybindings `ctrl+e`, `ctrl+i`, `ctrl+u`, `ctrl+b` are already defined by vscode, see the _Keyboard shortcuts_ menu. We should only keep the version with the prefix `ctrl+l`.

tecosaur:
I did consider that. However the `ctrl+b` etc. keybindings have the following `when` condition: `editorHasSelection`. This means that it will only take effect if text has been selected then they keybinding pressed. In all other situations, the default actions will be unaffected. 
Sinc only in that situation will that shortcut actually be active I felt it was relatively safe adding `ctrl+b` etc. as keybindings. From my use so far this has not caused any issues.

tecosaur:
Forgot about those taking selected text. I'll get on that :)

PR #833, opened 2018-09-30
Keybindings Patch 1

tecosaur:
PR in response to @jlelong in #809.
Noticed the first commit message isn't clear, it's an edit of `snippets-doc.md`


jlelong:
Some shortcuts are mentioned in the documentation but not defined in your PR and cannot be actually : `ctrl+r`, `ctrl+t`, `ctrl+s`. They should be removed from the documentation. I am in favor of only keeping keybindings prefixed by `ctrl+l` to avoid clashes with vscode itself or other extensions.

Thanks for fixing typos in my first version of the documentation 👍

tecosaur:
So what are you suggesting should be done? Remove the Extension Commands section of Shortcuts-doc.md?

jlelong:
No, I was making reference to your other PR, in which you add some new keybindings and only talking about these new ones. I am just saying that `ctrl+letter` cannot be used by the extension because of too many clashes with vscode shortcuts. For theses commands, just keep the `ctrl+l ctrl+letter` shortcut. Can you modify your two PRs accordingly?

tecosaur:
That should do it

Issue #847, opened 2018-10-04
Reintroducing the item-on-enter feature. Testers are welcome!


jlelong:
I have started reintroducing the feature of automatically inserting `\item` while hitting enter and the current line already starts in the same way. This was the prupose of #742, but it unfortunately introduced many bugs. I think I have fixed many.

I need help to test all this in the [item branch](https://github.com/James-Yu/LaTeX-Workshop/tree/item).

tecosaur:
What would I need to do to test this?

viktorstrate:
This is only apparent on the _item_ branch, and I don't know if you are already aware of this.

When I am inside a begin tag and press enter the cursor is not automatically indented, exactly as described here https://github.com/James-Yu/LaTeX-Workshop/issues/742#issuecomment-413203153.

jlelong:
Strange, because I thought I had fixed this in my last commit. At least it works fine for me.
Sorry to ask but are you sure you are up to date with the item branch? Are you using Windows? That might be related to eol encoding but I used the same approach as in [vscode-markdown](https://github.com/neilsustc/vscode-markdown/blob/master/src/listEditing.ts).

viktorstrate:
Oops, i pulled the branch from my fork.

jlelong:
So do you confirm that the indentation bug has been fixed for you?

viktorstrate:
I have just tested it, and the indentation works correctly. 👍

jlelong:
I have just added the variable `latex-workshop.bind.enter.key` to activate or not the feature, see c891f30. Default is true.

If no bug is found, I plan to merge it in master at the beginning of next week.

Issue #841, opened 2018-10-03
Merge vscode-latex-utilities into this project?


viktorstrate:
> This is a continuation of the end of the discussion in issue #809, but i've made a issue, because this isn't only related to shortcuts.

## Backstory
Hi, i'm the creator of another vscode extension called [vscode-latex-utilities](https://github.com/viktorstrate/vscode-latex-utilities). It is a small extension which aim is to extend this plugin, with new features. Because I wanted to add some bigger features, I decided to make my own extension, instead of submitting pull requests to this project.

## Merge request
As discussed in this issue viktorstrate/vscode-latex-utilities#5, @tecosaur asked me if I wanted to merge vscode-latex-utilities into this project, which I think is a good idea. In that way, users would not have to install multiple extensions to work with LaTeX in vscode.

Below is some of the reasons to merge the projects.

### Toggle tags (improvement)
In vscode-latex-utilities, tags can be toggled. When the cursor is inside an already wrapped sentence, and the command to toggle the tag is called, the tag will be removed. If it doesn't find the wrapping tag, it will add the snippet.

This currently only works for `textbf`, `textit`, `emph`, `underline`, but it is written in a way that is easily extendable.

The code for this can be found [here](https://github.com/viktorstrate/vscode-latex-utilities/blob/master/src/textManipulation.ts#L7-L68)

### \item is automatically added (addition)
When enter is pressed inside an `itemize` block, `\item` will automatically be added on the new line, this can be disabled in the settings, because I think that not everyone would want this.

## Your opinion
@tecosaur and I would like to hear what the maintainers @James-Yu and @jlelong think about this?
Please also read @tecosaur's issue viktorstrate/vscode-latex-utilities#5

I would be happy to make the initial merge on a new branch on vscode-latex-utilities, and send a pull request.

jlelong:
@viktorstrate Great if you are ready to contribute!

### Keybindings

I did not know your extension, so I have implemented a mechanism to easily assign a keybinding to any snippet. For now, toggling is not possible in what I have done but I am pretty sure we could adapt your approach to make it work with text transformation snippet. However, I am not sure people toggle `\textxx` commands that much. Is the extra effort worth it? 

### \item

I was about to create a new branch to reintroduce all the stuff around automatically adding `\item`.  It would nice if you could help us debug this. Here are the features I implemented
- When on a line starting in `\item([])?`, <kbd>Enter</kbd> creates a newline and inserts `\item([])?`
[...]

PR #848, opened 2018-10-04
Merge latex-utilities with LaTeX-Workshop

viktorstrate:
> This pull request is a result of the discussions in issue #841

This pull request currently merges the code from [latex-utilities](https://github.com/viktorstrate/vscode-latex-utilities) which handles shortcuts for keywords like `\textbf`, `\textit`, `\underline`.

For latex-utilities to be fully merged, we should find a solution to automatically add `\item` when enter is pressed. But I think @jlelong is working on that, on the 'item' branch, see #847.

### I've updated the yarn.lock
I have also regenerated the yarn.lock, as I could not download all the packages, because my version of node was too new. Below is the output before I updated the file. I haven't thoroughly tested the extension though.

```
yarn install v1.10.1
[1/5] 🔍  Validating package.json...
warning latex-workshop@5.10.2: The engine "vscode" appears to be invalid.
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
error upath@1.0.4: The engine "node" is incompatible with this module. Expected version ">=4 <=9". Got "10.11.0"
error Found incompatible module
```


jlelong:
Great! It seems that the toggle feature is not robust to nested `{...}`.

For instance, if you have
```
\underline{Some text with {\em braces} and more}
```
capture d ecran 2018-10-04 a 22 12 59

and call `toggleSelectedKeyword('underline')`, you get

capture d ecran 2018-10-04 a 22 11 27

It reminds me of bad headaches!

tecosaur:
I think this can be solved by changing the regex pattern match to require curly brackets to be matched.
This is what's happening at the moment with `\\keyword{(.*?)}`:
image

Here's a different pattern: `\\keyword(\{)(.*)(\})`
image
_NB: you want match group 2 from this_

Hope that helps

jlelong:
Thinking of it again, you can probably use the function [getLongestBalancedString](https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/providers/structure.ts#L11) to find the matching `}` once `\keyword{` is localized. Hope that helps.
You may need to move `getLongestBalancedString` somewhere in the extension.

tecosaur:
Perhaps use the regex match `\\keyword{(.*)}` then apply `getLongestBalancedString`? In my mind that should work.

viktorstrate:
@tecosaur Yes, that should not be that hard to implement, when we already have the `getLongestBalancedString`.

viktorstrate:
I have merged with master and fixed removing of tags with nested tags inside.

The code doesn't search further than the current line. If it finds a tag that is not closed off on the same line, then the closing bracket will not be removed. And if the cursor is inside a tag that is opened on a previous line, then it will not register it, and theirfore add a new tag.

I do not want to search the whole file for tags as it might be slow, but maybe an incremental approach could be used to check previous lines, only if it sees a dangling end bracket, or check the next lines, if `getLongestBalancedString` doesn't find the matching bracket.

jlelong:
The single line approach for tags toggling looks fine too me. I guess very few people use hard line wrapping.

Issue #844, opened 2018-10-04
I have some ideas, opinions?


tecosaur:
## I've had some ideas, I'd appreciate feedback on them
_i.e. are they worth pursuing?_

### Math Wrap Keybinding
 - I've been trying out the shortcut `ctrl+m ctrl+w` for _'wrap with math_ i.e. `\( … \)` 
 - have found it very handy so far
 - not sure about a corresponding snippet - perhaps `@mw`??
 - Can implement myself

### Section promotion / demotion
 - Idea: ability to promote and demote section commands
    - i.e. `chapter` ↔ `section` ↔ `subsection` ↔ `subsubsection` ↔ `paragraph` ↔ `subparagraph`
 - Idea came about because of how freqently I change the 'priority' of various parts of the documents I work on
 - I'm currently thinking `ctrl+l ctrl+p` (for promote) and `ctrl+l ctrl+d` (for demote) may not be a bad idea
 - May be able to implement myself

### Auto-rename matching tag
 - Basically [this](https://marketplace.visualstudio.com/items?itemName=formulahendry.auto-rename-tag) but for LaTeX
 - Probably wouldn't be able to implement myself without help

jlelong:
As I am not a fan of keybindings, I will let others respond. Just a small comment: math Wrapping already exists. Select text, hit `\`, the first entry in the list does what you want.

If what you mean by tag renaming is changing the name in a `\begin{...} / \end{...}` accordingly. This can be easily achieved by using _LaTeX Workshop: Select the current environment name_ and enter the new name.

tecosaur:
Selecting text then hitting `\` just overwrites the text for me - I feel like I'm missing something.

Re: environment setting - Never noticed 😅 thanks for the tip (I should really read through the commands list before having ideas like these)

One another note, regarding keybindings vs. snippets, the thing I have about snippets, particularly within maths is that (at least without me working out some keybinding modifications) I can't do snippet-in-snippet. 
I.e. Pressing tab after typing `@a` within a `\frac` from `@/` takes me to the second part of the fraction, and doesn't expand `@a`  —  do you know any way around this? It's why I prefer keybindings at the moment.

jlelong:
It does not seem possible to nest snippets when the first one contains placeholders. We have already come across that, see #609.

jlelong:
> Selecting text then hitting `\` just overwrites the text for me - I feel like I'm missing something.

[...]

Issue #609, opened 2018-05-08
Feature Request: Auto-Brackets in math-mode


maaaaadn:
### Feature Request Auto-Brackets in math-mode
since moving from emacs/auctex to VScode/LaTeX-Workshop I'm only missing a small but very comfortable feature. 
In math-mode if you type ^ or _ for super/subscript curly brackets where automatically added and the cursor placed inside (for super/subscripts with more than one character).

Is it possible to get this behavior in LaTeX-Workshop as well?

James-Yu:
Unfortunately, it's not possible for the extension to detect math-mode. An alternative is to provide an intellisense item upon typing `^` or `_`, which is however can be disturbing in non-math-mode.

Discussions welcomed.

jlelong:
I have a few extra snippets for such really cool features. 
First, `^` is a very tricky key for snippets as in most keyboard layouts, it is a dead key. To avoid interference with non math mode, I use the `vim-latex` approach, which is to trigger the snippet for `_ _` (double `_`) and `^^`.  To circumvent the dead key trouble, I actually have to type `^` + `space` twice.

maaaaadn:
@jlelong how do these snippets look like?
(sorry but I'm really just starting with VSCode and I'm not a real programmer)
a behavior like double ^ opens up a curly bracket would be sufficient for me. 
also if the brackets would appear in non-math mode wouldn't be a problem if there is an easy solution for that.

chantisnake:
@jlelong  I am also very curious how the snippet would look like.

chantisnake:
@James-Yu  I don't believe latex allow `_` or `^` outside of math mode. 
`_` or `^` outside of mathe modes will raise a `missing $ inserted` error. 

Therefore I believe when people trying to type `_` or `^` (without backslash `\` in front of them),
we can safely assume that they are in math mode.

jlelong:
In the menu _Preferences / User Snippets_, I choose the `latex` and simply added
```
	"subtext" : {
		"prefix": "__",
		"body": "_{$1}",
		"description": "Add a subtext"
	},
	"supertext" : {
		"prefix": "^^",
		"body": "^{$1}",
		"description": "Add a supertext"
	},
```
To use the snippet, just enter `_` twice and hit `TAB`.

@James-Yu, I could make a PR (one more 😜)

@chantisnake, `_` can appear on its own without a leading `\` outside math mode if one uses the `underscore` package.

[...]

PR #850, opened 2018-10-07
Change $0 in ^{} and _{} snippets to $1

tecosaur:
This is pretty minor, and there may well be a good reason for this, in which case please dismiss this.

However, I find myself repeatedly typing something along the lines of (`**`+`<tab>` → `^{ <cursor> }`) + `-1` + `<tab>` → <code>^{-1&nbsp;&nbsp;&nbsp;&nbsp;}</code>. As a result I'm wondering why the shortcuts aren't of the form `^{$1}` instead of `^{$0}`.

In case there's no significant reason for this I have made this pull request since I find it likely that I'm not the only one that gets tripped up by this.

### DIFF
--- a/snippets/latex.json
+++ b/snippets/latex.json
@@ -6,12 +6,12 @@
 	},
 	"subscript": {
 		"prefix": "__",
-		"body": "_{$0}",
+		"body": "_{$1}",
 		"description": "Add a subscript"
 	},
 	"superscript": {
 		"prefix": "**",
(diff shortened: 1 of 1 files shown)


jlelong:
May maths snippets used to use `$1` but  there was indeed a very good reason to have it changed and to use `$0` instead of `$1`: to enable nesting snippets.
This has been discussed in #609, see also your last comment in #844.

Issue #854, opened 2018-10-08
We can no longer tab out \( \)


innerlee:
I would blame https://github.com/James-Yu/LaTeX-Workshop/pull/617

tecosaur:
See #850

jlelong:
The benefits introduced by #617 (see explanations in #850) overcome by far the slight inconvenience of not being able to tab out of some maths snippets.

innerlee:
What was the benefit? If it was for auto-brackets, i think that maybe auto-brackets needs a better solution rather than sacrifice this.

jlelong:
This is discussed in details in #609 and applies too many more snippets than `\( .... \)`. We could not find any better approach at the time and nobody has complained for 5 months. If you have a better idea, you are welcome to expose it and eventually make a PR.

innerlee:
maybe relevant link https://github.com/Microsoft/vscode/pull/52634

PR #855, opened 2018-10-08
Zz/bringmybracketback

innerlee:
Fix https://github.com/James-Yu/LaTeX-Workshop/issues/854
Also re-fix https://github.com/James-Yu/LaTeX-Workshop/issues/609

### DIFF
--- a/data/commands.json
+++ b/data/commands.json
@@ -7,17 +7,17 @@
   },
   "lrparen":{
     "command":"left(",
-    "snippet":"left($0\\right)",
+    "snippet":"left(${1}\\right)${0}",
     "detail":"\\left( ... \\right)"
   },
   "lrbrack":{
     "command":"left[",
-    "snippet":"left[$0\\right]",
+    "snippet":"left[${1}\\right]${0}",
     "detail":"\\left[ ... \\right]"
   },
   "lrcurly":{
     "command":"left\\{",
-    "snippet":"left\\{$0\\right\\\\}",
+    "snippet":"left\\{${1}\\right\\\\}${0}",
     "detail":"\\left\\{ ... \\right\\}"
   },
   "frame":{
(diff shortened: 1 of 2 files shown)


jlelong:
This breaks [snippet nesting](https://github.com/James-Yu/LaTeX-Workshop/issues/609#issuecomment-387475664), which was the main issue about snippets and using `${1}`.

Without your PR: `\(@a\)` expands to `\(\alpha\)`.
With it: `\(@a\)` does not expand. `TAB` just jumps out of the `\(...\)`.

tecosaur:
Do you know if there might be any way to force vs code to expand a snippet if it can, then only if it can't move on to the next $ region.

jlelong:
Indeed, I have search a lot for this and never found anything interesting. So I guess it a no.

innerlee:
I do not understand the concept of `nesting`.

For the link https://github.com/James-Yu/LaTeX-Workshop/issues/609#issuecomment-387475664, we can auto complete brackets in math mode with this pr, so no need to type `__<tab>`-like things.

For the `@a` example, I cannot make it work on my machine. Do I need to turn on any option to get the `@something` nesting?

innerlee:
Cannot get it working with the version on marketplace

abc

jlelong:
Have checked that you have `"editor.tabCompletion": true`?

innerlee:
Both `true` or `false` do not work.
latexworkshop version 5.11.1
vscode version 1.27.2
win10

jlelong:
Have you tried playing a bit with `"editor.wordBasedSuggestions": false/true` and `"editor.quickSuggestions": false/true`?
A few people have already reported this but we have never managed to understand what was happening. Maybe this time...

Astrantia:
@jlelong i'll jump in. having both quicksuggestions and wordBasedSuggestions set to true, I still don't get any suggestions when I type `@a`

jlelong:
As @Astrantia pointed out, the completion list for `\a` is so huge, `\alpha` is only on 18th position and you have to type `\al` to have it on the first completion screen. The `@` snippets are designed such that two prefixes cannot start with a common substring. Moreover, `\` requires more keystrokes (3 on my layout) than `@` on many keyboard layouts.

Astrantia:
@jlelong Why does this happen though?
https://i.imgur.com/JhXLMHY.gifv

jlelong:
This is related to [vscode#35651](https://github.com/Microsoft/vscode/issues/35651)
Intellisense does not take into account non alphabetical characters, hence the long list you get. But, no matter the list, the snippet you expect is always the first one!

Astrantia:
@jlelong You misunderstood the point of the gif. I expect to get suggestions for `@8`, `@/`, `@%` as they are also defined

Astrantia:
@jlelong https://github.com/James-Yu/LaTeX-Workshop/blob/master/snippets/latex.json#L42
Notice that it doesn't appear in my gif.

Astrantia:
Update: I had to set `"editor.tabCompletion": "onlySnippets"` for `@%` + Tab to expand the snippet. I still don't get suggestion pop-up, but tabbing works now.

innerlee:
okay, I will wait for https://github.com/Microsoft/vscode/issues/45385

PR #862, opened 2018-10-10
Fixed invalid json as per instruction in James-Yu/LaTeX-Workshop#429

Swoy:
Before I embarked on my thesis, I PR'd according to the propsal in  James-Yu/LaTeX-Workshop#426. This is now fixed and the file is ready for vetting. I've linted the file and everything seems to work well here.

### DIFF
--- a/data/packagenames.json
+++ b/data/packagenames.json
@@ -267,7 +267,7 @@
 "add-wes": {
 	"command": "add-wes",
 	"detail": "https://ctan.org/pkg/add-wes",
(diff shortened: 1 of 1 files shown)


Swoy:
I really don't understand where rebase & merge is gone? Since I already made the change before I remembered to update from the original repo - so I guess this is why this file has the wrong filename?

I will delete my local repo now and re-fork it if nobody knows how to pull from main without using arcane gitcommands.

PR #863, opened 2018-10-11
Fixed invalid json as per instruction in James-Yu/LaTeX-Workshop#429

Swoy:
James-Yu/LaTeX-Workshop#426 is the ref here. @jlelong closed the other PR.


Swoy:
@jlelong This should be perfectly fine ( the previous PR was fine too btw).

@James-Yu: You mentioned in feb (yeah, who would remember) that you would take care of the descriptions and integrate them in LaTeX-Workshop once I had fixed the file.

jlelong:
I have updated the completion mechanism to take into account the new format of the `packagenames.json` file but I cannot figure out how to make package urls clickable...

Swoy:
Thanks! Yeah I tried a bunch of terms, I cannot find anything worth a mention either. I can see there is a read more icon on some completion listed (some have this blue circular blue shape with an "!" inside) If you ctrl-space once more, there is a small dropdown that is displayed. Maybe it can work there?

PR #864, opened 2018-10-11
Bundle LaTeX-Workshop with webpack

NavyAdmiral:
This is too heavily borrowed from the author of the other great extensions @fabiospampinato so all credits go to him. From what I'm seeing the activation times are reduced by 80%.

In order to test, just type `yarn compile` after cloning the branch.

@James-Yu @jlelong You can take a look at this now. Workflow after this PR should still be the same: `npm run compile/yarn compile`. I don't get any errors after testing and the startup times are better with less dist files.


package.json
```diff
@@ -831,23 +832,28 @@
     "workspaceSymbolProvider": "true"
   },
   "dependencies": {
+    "bufferutil": "^4.0.0",
     "chokidar": "^2.0.3",
     "fs-extra": "^5.0.0",
+    "fsevents": "^1.2.4",
```
fabiospampinato: `fsevents` should be an optional dependency, I think it only works on macOS and its installation fails in other OSs.

package.json
```diff
@@ -831,23 +832,28 @@
...
     "glob": "^7.1.1",
     "micromatch": "^3.0.0",
     "pdfjs-dist": "^1.8.609",
+    "ts-loader": "^5.2.1",
```
fabiospampinato: `ts-loader` is only a development dependency.

James-Yu:
Thanks for the contribution!

James-Yu:
Have to comment again. It's such a great improvement of the startup performance of the extension! Brilliant!

James-Yu:
@NavyAdmiral do you mind have a check on the latest PR about hover preview? It seems that the module mathjax-node does not work with webpack, but is an essential module to the feature. Thanks in advance.

PR #867, opened 2018-10-11
Hover preview

tamuratak:
Now, I believe this patch get good quality to be merged.

### Note
Because VSCode supports SVG format as image in markdown, without WebView, we can implement hover preview generating SVG with  mathjax-node. However, we cannot know actual colors used in each theme in extension process. To work around this problem, we need WebView anyway.


James-Yu:
Have to say it's just awesome. Will do some test and merge ASAP.

tamuratak:
Thank you. I hope you enjoy it.

tecosaur:
@tamuratak regarding colours and webview — would it be possible that when webview is used the colour scheme info is stored so that in subsequent sessions previews are possible without webview by using the stored colours?

In part, I mention this because I have a low-resolution smallish screen laptop and often I can't really afford the real estate of the side panel, and so being able to use the preview without that would be quite beneficial.

James-Yu:
The webview serves a significantly more important role than getting the color. I have checked the implementation and cannot think of a working solution without a visible webview. @tamuratak cmiiw.

tamuratak:
Without WebView, using mathjax-node, we can generate SVG of equations, c.f. use_mathjax_node. So if we know colors used in a current theme in an extension process, or if we know whether the theme is light or dark, we don't need WebView. 

The main issue is the following. For LaTeX-Workshop users, it is difficult to understand, intuitively, that they have to open a pdf viewer to save color information if math preview in hover does not work, or when they switch a dark theme to a light one. There are lots of ways when and how to tell users such things. In documents,  in an information message above status bar, or in hover. We have to choose a right way. I think that is the main issue.

Anyway, we'd better make a new issue on this topic.

sloth3toe:
@tamuratak , sorry for being ignorant, but I've been using the most updated extension (5.13.0 right now) and still can't figure out this hovering feature. I do not see any hovering capability, even though I have a lot of both in-line math and align environments. Do you have any tips?

PR #872, opened 2018-10-13
CTAN package list script (completion)

Swoy:
Added a script that will fetch an updated list of all packages from the CTAN repo, should make it easier to refresh this list on new releases.

### DIFF
--- a/dev/ctanpkglist.py
+++ b/dev/ctanpkglist.py
@@ -1,0 +1,20 @@
+'''
+This script fetch latest list of packages and their descriptions
+from CTAN at https://ctan.org/pkg and save the result as a json file.
+The result is used to generate command intellisense for LaTeX Workshop.
+'''
+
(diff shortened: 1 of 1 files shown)


Swoy:
If I by chance also find it in me to compile package examples over the next months, how would you like me to structure this?
Does anyone know if the completion popup (in lack of better words) support html or Markdown from JSON?

I could add it as a prop; 'sample': '#a0poster \n this is how you use a0....'?

Swoy:
@James-Yu I made sure it won't invalidate the json 🥇

jlelong:
The auto-completion menu des not seem to have html neither Markdown support. Unfortunately, we are left with text only.

James-Yu:
Any data structure you consider appropriate would be fine, though it seems that we have to push vscode for providing rich API support.

PR #878, opened 2018-10-14
Fix ctrl+l ctrl+e keybinding

tecosaur:
### Issue
Due to silly naming I did when creating the `\emph` snippet (called it `"textem"`) when the implementation was changed, since it was referred to as `textem`  — <kbd>ctrl</kbd>+<kbd>l</kbd> <kbd>ctrl</kbd>+<kbd>e</kbd> now toggles the nonexistant `\textem` command.
### What's in this PR
- Renamed all instances of `textem` to `emph`
- Removed `\item` inserting keybinding while I was at it
- Removed toggle `textem` function, since it isn't a thing

### DIFF
--- a/package.json
+++ b/package.json
@@ -285,12 +285,6 @@
         "key": "ctrl+alt+x"
       },
       {
-        "key": "ctrl+l ctrl+enter",
-        "mac": "cmd+l cmd+enter",
-        "when": "editorTextFocus && !editorReadonly && editorLangId == 'latex'",
-        "command": "latex-workshop.shortcut.item"
-      },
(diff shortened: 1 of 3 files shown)


jlelong:
I think it makes sense to keep the shortcut to insert `\item`
- for inserting the first `\item` in custom environment
- for people using hard wrapping (insert line breaks) instead of soft wrapping as the insert `\item` on enter feature only checks the current line.

I will modify your PR in that sense before merging.

jlelong:
This `\textem` to `\emph` conversion has been merged.

PR #880, opened 2018-10-15
Display user-defined control sequence as is.

tamuratak:
Hi,
this is a patch to display user-defined control sequence as is. Related to link.


tecosaur:
@James-Yu Would it be worth patching this out as `5.12.2`? In my opinion it has a big impact on the hover on preview feature.

Issue #876, opened 2018-10-14
[Bug?] Long mathjax render gets cropped


tecosaur:
The below relates to the following LaTeX code,
```
\begin{align*}
	\Lap[t H(t-1)] + \Lap[(3-t)H(t-2)] &= \Lap[\underbrace{((t-1)+1)}_{f(t)=t+1}H(t-1)] + \Lap[\underbrace{(1-(t-2))}_{f(t)=1-t}H(t-2)] \\
	&= e^{-s}\left( \frac{1}{s^2} = \frac{1}{s} \right) + e^{-2s}\left( \frac{1}{s} - \frac{1}{s^2} \right)
\end{align*}
```
For this aligns the preview produced is as below
![image](https://user-images.githubusercontent.com/20903656/46912456-91a89500-cfa8-11e8-8cd1-ae098890ecac.png)
When I expect something more like this,
![image](https://user-images.githubusercontent.com/20903656/46912468-d7655d80-cfa8-11e8-9464-69165643ab65.png)
This is from the mathjax online demo. With #875 fixed I would get the following,
![image](https://user-images.githubusercontent.com/20903656/46912481-fcf26700-cfa8-11e8-81fa-50bf4027fe47.png)
Sorry for the issue-spam.

tamuratak:
Related to Microsoft/vscode/issues/14165. To work aroud this problem, please set  `latex-workshop.hoverPreview.scale` to less than 1.0. I have set it to 0.9.

PR #890, opened 2018-10-16
WIP. Offline hover preview.

James-Yu:
This is a work in progress. The only TODO is to handle the conflict between `mathjax-node` and `webpack`. In debug mode, things should work. I'm just too tired to go on today. Leaving it to following days.


James-Yu:
@tamuratak We might join force on fine tuning the preview things. I will handle the webpack.

NavyAdmiral:
What exactly is the problem with compiling mathjax-node with webpack? I tried to compile it on my machine and this was the output:

image


This might be relevant too: https://github.com/mathjax/MathJax-node/issues/332

James-Yu:
@NavyAdmiral It won't load due to the dirty implementation https://github.com/mathjax/MathJax-node/blob/master/lib/main.js#L84 . MathJax-v3 is not being considered due to other reasons. I am thinking of disabling webpack bundling if the problem cannot be solved. All in all the extension activation is not causing great trouble.

NavyAdmiral:
@James-Yu Whats the problem of having online mathjax implementation?

James-Yu:
The key point is that the implementation allows hover preview without webview, which is a great improvement. Offline is a bonus.

NavyAdmiral:
https://github.com/mathjax/mathjax-v3/releases/tag/3.0.0-beta.2 Can this be used instead of mathjax-node?

James-Yu:
Almost not as indicated in https://github.com/mathjax/mj3-demos-node . Hard to translate TeX to SVG.

NavyAdmiral:
According to the creator of mj-node, he's going to deprecate the package when mj3 lands, ref https://github.com/mathjax/MathJax-node/issues/375

James-Yu:
The information is not helping resolve the problem.

James-Yu:
MathJax v3:
```
The current TeX input processor has all the core functionality of the MathJax v2 TeX input, and several of the extensions built in, but some extensions are still to come. For example, \unicode, \bbox, and the color extension are not yet available.

The CommonHTML and SVG output implement all the MathML elements that they do in v2, but do not yet include support for line breaking (neither automatic nor explicit ones); this will be implemented in a later beta version. Both output renderers currently only support the MathJax TeX font; other fonts will be added in the future.
```
These are among the critical reasons why it cannot be used for now.

James-Yu:
It's a hard decision to decide scrap webpack in this branch for NOW. I will go on looking into ways of resolving the issue.
Here is a demo vsix of this branch up to now for testing.
[latex-workshop-5.12.2.vsix.zip](https://github.com/James-Yu/LaTeX-Workshop/files/2485398/latex-workshop-5.12.2.vsix.zip)

NavyAdmiral:
One quick thing I noticed that I think @tamuratak fixed earlier is `undefined custom sequence` 
Try defining:
```
\newcommand*{\R}{\ensuremath{\mathds{R}}}
```

Then, try to preview `$\R \sin{x}...$`

NavyAdmiral:
Since mj-node is not updated too frequently, why don't we make our own fork with a patch for https://github.com/mathjax/MathJax-node/blob/master/lib/main.js#L84 ?

James-Yu:
That is desired. However, its not as simple as locate the correct file location. Trust me, I've tried that, and more problems arise.

James-Yu:
Is there a way for webpack to exclude `node_modules` or more desirably selected modules in the folder? I tried google a solution, but nothing worked.

NavyAdmiral:
@James-Yu Did you take a look at https://www.npmjs.com/package/webpack-node-externals ?

James-Yu:
Because I didn't know about `externals` 🤣 It seems that this one works. Will do some testing and push commits.

NavyAdmiral:
One thing to note, there's no horizontal scrollbar when the svg exceeds hover width.

James-Yu:
Long svg bug see here #876

NavyAdmiral:
That solution is really hacky and not working for even longer SVGs. 
What I'm looking for is the horizontal scrollbar that vscode randomly(?) provides on some hovers.

James-Yu:
Here's what I got. You may notice the scroll.
image

NavyAdmiral:
@James-Yu I edited my comment. That is exactly what I'm looking for, but try reloading vscode and hovering over it for the first time. You will notice that horizontall scrollbar isn't there on the first hover. Is there any reason why this would behave like this?

James-Yu:
I also noticed your observation. I suspect this would be an upstream issue of vscode. All in all there is no difference from the extension side on how mathjax was called. Actually I inspected the generated svg xml and base64 encoded strings, they are identical for the first run and subsequent ones.

tamuratak:
Hi, should I make a PR of my branch to offline-mathjax, or to master?

NavyAdmiral:
There's not too much that can be done on our end until vscode makes max-width of tooltips configurable. This PR looks good to me. @tamuratak has come up with a good solution for colors for now too.

James-Yu:
@tamuratak This is the said color-reading commit. You may directly contribute to this PR, or make PR to this PR/master. We will see which one is better for the feature.

tamuratak:
I prefer my PR. The followings are reasons.
 
1. On Default Light+ theme, this PR does not work well. To parse JSON files for VSCode themes, we support "include" properly, which is not in JSON standard. [link](https://github.com/Microsoft/vscode/blob/master/extensions/theme-defaults/themes/light_plus.json).
```
{
        "$schema": "vscode://schemas/color-theme",
        "name": "Light+ (default light)",
        "include": "./light_vs.json",
        "tokenColors": [
                {
```
2. Default Light+ theme does not define `activityBar.foreground`.  Which color each theme defines depends on each author. One theme defines `editor.foreground`, but another one doesn't. What we really need is [window.getComputedStyle](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) although VSCode doesn't allow it.
3. An appropriate color for math preview in hover depends on each theme. For example, on Ayu Dark theme, this PR causes no error, but math preview is almost invisible. So, for [Aryu](https://marketplace.visualstudio.com/items?itemName=teabyii.ayu) Dark, we must choose a color different from `activityBar.foreground`.


In short, because VSCode themes are in chaos, we had better avoid to touch them directly. This is the reason that I prefer my PR.

Regards,

James-Yu:
@tamuratak Please have a test on this new commit. The W3C way of calculating BW color is used based on the background color of editor, which "should" be defined in any serious themes. Fallbacks to 'type === 'dark', then to black.

tamuratak:
Great. This works really well. I close my PR.

tamuratak:
Please don't forget `rm -rf viewer/mathjax/`.

PR #908, opened 2018-10-20
Add gifs to readme

tecosaur:
I think this makes the readme a bit prettier and better showcases the extension's capabilities. 

Added files: **3× .png + 15× .gif**
Total size: **2.3 MB**
Theme used: **Light+**


James-Yu:
Can you please move the media to a root folder with name e.g. `screencasts` or `media`? This reduces problem dealing with paths of spaces, which might emerge in some future time.

tecosaur:
Sure, I'll do that later tonight.

>

tecosaur:
On second thoughts I can't see any easy way to do that (at least via Githib web), any suggestions?

tecosaur:
Time for plan B

James-Yu:
It seems that README.md was not changed

tecosaur:
???
image
Why do you say that?

tecosaur:
Ahh, I think I might have found why you would think that - I forgot to charge the `src` attribute of the images when I moved the images.
I also noticed a sentence that was somewhat dodgy (I think written by me originally), thought I might as well fix that.

PR #909, opened 2018-10-21
Add [View on pdf] link  in hover on \ref

tamuratak:
Hi,
this is a patch to add  [View on pdf] link in hover on \ref. When clicking the link, SyncTeX command is executed for \\label{...} corresponding to \\ref{...}. We can do the same thing by clicking [Go to definition] or [Peek definition], moving a cursor to the label, then executing SyncTeX. But, the advantage of this patch is that SyncTeX command can be executed for other than the current PDF viewer. We can use the current viewer to view a sentence we are editing now, and use tab viewer or browser to view a part \\ref{...} refers to, and vice versa. With this patch, we overcome the imperfect navigation of PDF viewers in a measure.

---
oct-21-2018 16-32-19

### DIFF
--- a/package.json
+++ b/package.json
@@ -593,6 +593,16 @@
             "[Experimental] Open PDF with the external viewer set in \"View > Pdf > External: command\"."
           ]
         },
+        "latex-workshop.view.pdf.ref.viewer": {
+          "type": "string",
+          "default": "auto",
+          "enum": [
+            "auto",
+            "tabOrBrowser",
+            "external"
+          ],
+          "description": "PDF viewer used for [View on PDF] link on \\ref."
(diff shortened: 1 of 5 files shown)


tecosaur:
Nice!

realjoenguyen:
@tamuratak @James-Yu hi, what is this command? I want to set the shortcut for "view on pdf". Thank you!

jlelong:
Unfortunately, there is no easy access to the real command behind "view on pdf". It calls `latex-workshop.synctexto` with arguments coming from the reference database internally computed by the extension. Binding this call to a keyboard shortcut would require specific development.

PR #926, opened 2018-10-29
add animation for pdf-to-tex SyncTeX

tamuratak:
Hi,
this is a patch to add animated notification for pdf-to-tex SyncTeX on pdf viewer. 

oct-29-2018 13-47-30

Please notice that this patch does not add animation for pdf-to-tex SyncTeX on  external pdf viewer.


tecosaur:
I notice that this seems to only 'highlight' the first word. I'd personally consider it to be slightly preferable if it 'highlighted' the line, and started with non-solid red, something like opacity=0.8. Any thoughts?

PR #928, opened 2018-10-29
Clean output directory recursively

zoehneto:
This PR implements recursive cleaning when invoking the "Clean up auxiliary files" action. This is needed for projects where some latex files are in a sub directory e.g.:
```
main.tex
chapters/chapter1.tex
chapters/chapter2.tex
```

### DIFF
--- a/src/components/cleaner.ts
+++ b/src/components/cleaner.ts
@@ -23,7 +23,9 @@ export class Cleaner {
             outdir += path.sep
         }
         if (outdir !== './') {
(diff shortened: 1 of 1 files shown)


James-Yu:
Sorry to revert this PR. Please see #930 for the concern.

zoehneto:
@James-Yu @tamuratak Thanks for the comment, I didn't think of this issue. I think the solution would be to only clean directories recursively if the output directory isn't `./` i.e. when the user has specified an output directory. I would assume that there is no important content (content that should not be deleted) in the output directory.

Issue #930, opened 2018-10-30
PR  #928 must not be merged. 


tamuratak:
@James-Yu. I think PR #928 must not be merged. This PR deletes files too aggressively. Subdirectories do not necessarily contain TeX-related files only.  For example, it also deletes files with the suffix `.log`  in subdirectories.  When someone has data files with the suffix `.log` in a subdirectory to write a paper, the result will be disastrous.

James-Yu:
Hmm fair enough. Will revert it.

tamuratak:
Thank you for your quick response.

PR #933, opened 2018-10-30
Clean output directory recursively

zoehneto:
This PR implements recursive cleaning for the output directory (only when an output directory other than `./` is specified) when invoking the "Clean up auxiliary files" action. This is needed for projects where some latex files are in a sub directory e.g.:

```
main.tex
chapters/chapter1.tex
chapters/chapter2.tex
```

@James-Yu @tamuratak This is a new version of PR #928 with the concerns from issue #930 in mind. The concerns are handled by only cleaning in the specified output directory, which shouldn't contain any files apart from those generated by the build.

### DIFF
--- a/src/components/cleaner.ts
+++ b/src/components/cleaner.ts
@@ -23,7 +23,7 @@ export class Cleaner {
             outdir += path.sep
         }
         if (outdir !== './') {
-            globs = globs.concat(globs.map(globType => outdir + globType))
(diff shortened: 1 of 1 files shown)


James-Yu:
@tamuratak if you are good with the PR, I will merge it. I did not find niche cases in which the PR breaks things.

tamuratak:
I have not found any problem. I hope we are right.

PR #938, opened 2018-11-01
fix toolbar in pdf viewer

tamuratak:
Hi,
this is a patch to fix the toolbar in pdf viewer broken for a certain width. This patch hides toolbar buttons in the following order.

1. next and previous buttons  
2. scaleSelect
3. trimSelect


tamuratak:
I close this PR. I wil submit a new one later.

Issue #263, opened 2017-09-06
Proposal: Folding of sections and subsections


StepanSushko:
### Description

Currently, there is folding feature for begin/end constructions and some other.

I'd like to fold sections/subsections too, it is convinient when a text is long.

Thx in advance!

SigurdAngenent:
I would like to second this proposal

James-Yu:
https://github.com/Microsoft/vscode/issues/105

```
@ColacX The current folding only depends on indentation and is not language aware.
The request for language aware folding is https://github.com/Microsoft/vscode/issues/3422
```

smagill:
It looks like an API for language-aware folding (not based on indentation) has been recently added:  
https://code.visualstudio.com/updates/v1_21#_folding

I too would love to have section-based folding in LaTeX!  Thanks for the great extension!

steveeJ:
@James-Yu please revive this issue :-) It seems to be possible to implement this as @smagil pointed out.

neilsustc:
This is the top request now 😉

![image](https://user-images.githubusercontent.com/7588612/47622784-0e875300-db01-11e8-86c6-c861341bff91.png)

PR #944, opened 2018-11-04
Folding range provider

neilsustc:
Provide section-based and environment-based folding (#263)

demo

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -25,6 +25,7 @@ import {ProjectSymbolProvider} from './providers/projectsymbol'
 import {SectionNodeProvider} from './providers/structure'
 import {DefinitionProvider} from './providers/definition'
 import {LatexFormatterProvider} from './providers/latexformatter'
+import {FoldingProvider} from './providers/folding';
 
 function renameConfig(originalConfig: string, newConfig: string) {
     const configuration = vscode.workspace.getConfiguration('latex-workshop')
@@ -259,6 +260,7 @@ export async function activate(context: vscode.ExtensionContext) {
     context.subscriptions.push(vscode.languages.registerCompletionItemProvider({ scheme: 'file', language: 'latex'}, extension.completer, '\\', '{', ',', '(', '['))
     context.subscriptions.push(vscode.languages.registerCompletionItemProvider({ scheme: 'file', language: 'doctex'}, extension.completer, '\\', '{', ',', '(', '['))
     context.subscriptions.push(vscode.languages.registerCodeActionsProvider({ scheme: 'file', language: 'latex'}, extension.codeActions))
+    context.subscriptions.push(vscode.languages.registerFoldingRangeProvider({ scheme: 'file', language: 'latex'}, new FoldingProvider()))
 
     extension.linter.lintRootFileIfEnabled()
     obsoleteConfigCheck()
--- a/src/providers/folding.ts
+++ b/src/providers/folding.ts
@@ -1,0 +1,82 @@
+import * as vscode from 'vscode'
+
+export class FoldingProvider implements vscode.FoldingRangeProvider {
+
+    public provideFoldingRanges(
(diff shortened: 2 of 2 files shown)


Astrantia:
@neilsustc Thanks!!!
@James-Yu Please push a new release in marketplace right away once this is merged :D

tecosaur:
This is fantastic. Thanks @neilsustc

Astrantia:
@neilsustc 

Actually I find some bugs with multiple environments:

```
\section*{Bug1}

\subsection{Intro} 
\begin{enumerate} % BUG 2
	\item  a
	\item  d
	\item  s
	\item  q
	\item  q
	\item  a
\end{enumerate}

\subsection{Now}
\begin{enumerate}
		\item d
\end{enumerate}

\begin{enumerate}
		\item
\end{enumerate}
```
Bug 1: The uppermost Section doesn't get fold option
Bug 2: folding enumerate folds both enumerates

Astrantia:
Yes, the current implementation is quite buggy 😞. At least, we've got something to work on.

neilsustc:
You are right. Some further improvements are needed.

Astrantia:
@neilsustc no longer working on this?
@James-Yu this implementation is not working for 99% of tex files, why don't we revert to the indentation folding?

neilsustc:
Bug 1 is easy to fix
Bug 3 might be fixed with #959 (I guess)
As for bug 2, it is a little bit tricky. It seems there is no simple way if we stick to the current "regex matching" method

I might have some time during the weekend

If the current stage is unusable for you, you can always set `editor.foldingStrategy` to `indentation` ([link](https://github.com/James-Yu/LaTeX-Workshop/issues/953#issuecomment-436583427))

Astrantia:
@neilsustc Unfortunately changing that setting doesn't fallback to the default indentation. I already tried it, but extension still used the Provider implemented by you.

neilsustc:
Then it is a vscode issue if you did it right. It works for me and at least another people in above link.

iamhyc:
I am not familiar with regex, but I think there is a bug with `envRegex`. You should at least set the regex flag from `g` to `gU`.

when the flag is `g`, wrongly folded.
image


when the flag is `gU`, works correctly.
image

(regex tested on [https://regex101.com](https://regex101.com/))

neilsustc:
There is no good solution (in regex) since both following cases are possible

```
\begin{env}
    ...
\end{env}

\begin{env}
    ...
\end{env}
```

and 

```
\begin{env}
    \begin{env}
        ...
    \end{env}
\end{env}
```

No single unified regex can handle both cases (seems to me).

iamhyc:
> There is no good solution (in regex) since both following cases are possible
> 
> ```
> \begin{env}
>     ...
> \end{env}
> 
> \begin{env}
>     ...
> \end{env}
> ```
> and
> 
> ```
> \begin{env}
>     \begin{env}
>         ...
>     \end{env}
> \end{env}
> ```
> No single unified regex can handle both cases (seems to me).

Oh, it seems like a operation stack. but anyway, I have disabled this provider and waiting for the bug fixed :)

neilsustc:
No

jlelong:
Just a piece of advice, you can have a look at the function [locateMatchingPair](https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/components/envpair.ts#L79), which deals with the problem of finding matching `\begin/\end` pairs.

neilsustc:
Thanks for the information

James-Yu:
I will revisit the folding logic.

iamhyc:
Hi everyone, I make a PR to solve the folding problem with env.

Could you help check on #984 ? I think my implementation may be naive without thorough test.

Thank you!

James-Yu:
@iamhyc It looks good for me, will merge and make minor edits.

Issue #953, opened 2018-11-05
Version 5.15.0 removes folding in *.tex files


DrRuhe:
### Description

Since the Update to version 5.15.0 of Latex Workshop the code folding dissappeared in .tex files.

### Log messages

Although i doubt that the Logs will give any meaningful info about the Issue here they are:

I replaced all paths in the log with this text: "PATH TO A FILE"
It was the same path all the time btw.

[15:34:28] Initializing LaTeX Workshop.
[15:34:28] Creating LaTeX Workshop http and websocket server.
[15:34:28] LaTeX Workshop initialized.
[15:34:29] Server created on 127.0.0.1:58210
[15:34:29] LaTeX Workshop version: 5.15.0
[15:34:29] Snippet data loaded.
[15:34:29] LOG command invoked.
[15:34:29] Default data loaded.
[15:34:29] Try root file in root directory: PATH TO A FILE
[15:34:29] Parsing PATH TO A FILE
[15:34:29] Try root file in root directory: PATH TO A FILE
[15:34:29] Parsing PATH TO A FILE
[15:34:29] Try root file in root directory: PATH TO A FILE
[15:34:29] Parsing PATH TO A FILE
[15:34:29] Try root file in root directory: PATH TO A FILE
[15:34:29] Parsing PATH TO A FILE
[15:34:52] Found root file from active editor: PATH TO A FILE
[15:34:52] Root file changed from: undefined. Find all dependencies.
[15:34:52] Instatiating new file watcher for PATH TO A FILE
[15:34:52] Parsing PATH TO A FILE
[15:36:20] LOG command invoked.
[15:36:22] LOG command invoked.

Nothing Suspicoius in either the Developer Console nor the Ouput tab of VS Code
### How to reproduce

1. Update Latex Workshop to 5.15.0
2. 
3. 

### Expected behavior

Able to Fold Code

### Actual behavior

Folding icons are there for a brief moment after reloading VScode but then dissappear

### Additional Information

I did double check if folding was disabled in the settings.json but it was active

DrRuhe:
I did some more testing and it seem as the extension only removes folding options where there is no direct latex function it could fold. Maybe adding an option in the settings to only show folding points where there is an actual latex function would help.

James-Yu:
Can you please give an example on the part of a document which was originally and should be folded, but cannot in the current version?

tecosaur:
Here's an example from me:

```
\makeatletter
[...]

PR #959, opened 2018-11-06
Fix section folding

zoehneto:
The new folding implementation has two issues which are addressed in this PR:
1. To determine if a section is the last one, it checks whether there are any more sections instead of checking if there are any more section on the same level or on a higher level (line 32).
2. For the last section the endLine variable is decremented in the loop, but the value of endLine should only be changed if there is actually a `\end{document}` present (not the case for typical includes, line 41).

This should resolve issue #953

### DIFF
--- a/src/providers/folding.ts
+++ b/src/providers/folding.ts
@@ -29,7 +29,9 @@ export class FoldingProvider implements vscode.FoldingRangeProvider {
         return sections.filter(section => section['level']).map((section, index, allSections) => {
             const startLine = section['lineNumber']
             let endLine
-            if (index < allSections.length - 1) { // Not the last section
+
+            // Not the last section
+            if (allSections.filter((element, elementIndex) => index < elementIndex && element['level'] <= section['level']).length > 0) {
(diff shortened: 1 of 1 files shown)


neilsustc:
> To determine if a section is the last one, it checks whether there are any more sections instead of checking if there are any more section on the same level or on a higher level (line 32).

It is not true since we have [this line](https://github.com/James-Yu/LaTeX-Workshop/pull/959/files#diff-5b3381ac03051642feca3883eb701cedR36).

Now they are redundant. We need to remove one of it 😂

zoehneto:
Say we have a structure like this:
```
section
    subsection
section
    subsection
```

Thinking some more about the code, the line you point to does't check the right thing for the first section, since it should check for sections with same or higher level than the current section (so the folding includes the subsections).

For the second section we have to use the else path, since the current code would take the start of the second subsection as the end, instead of taking the end of the file. The else path still has to be used even with my suggested change, since no section with same or higher level follows.

neilsustc:
makes sense 👍

neilsustc:
I mean the second paragraph is true. As for the first paragraph, I don't think so... since you can see `section 1` does fold `subsection 1.1` in (see GIF shown in the original PR)

PR #964, opened 2018-11-06
delete Spell Right

tamuratak:
As seeing recent issues, we had better not to recommend a problematic solution.


tecosaur:
I'd go a bit further than that. I'd add a small section listing packages that are known not to play well with this extension.

tamuratak:
That's good 👍

PR #969, opened 2018-11-08
Add testing

tamuratak:
Hi,
this is a patch to add minimal tests. I heavily use `sleep` to wait for build commands and others to finish.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,8 @@ node_modules/*
 dev/unimathsymbols.txt
 VERSION
 .idea/
+test/texfiles/**/*.aux
+test/texfiles/**/*.dvi
+test/texfiles/**/*.gz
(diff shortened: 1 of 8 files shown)


tecosaur:
If this is added, could we also integrate a CI testing? something like Travis or Circle.

zoehneto:
At least on travis ci you can install packages as described [here](https://docs.travis-ci.com/user/installing-dependencies/), though the obvious question is how long it would take to install texlive :)

BreakBB:
@tamuratak I am curious why you're using `sleep` inside the tests since you wait for the Promises to resolve right before?

tamuratak:
Rejected in https://github.com/James-Yu/LaTeX-Workshop/pull/1284#issuecomment-480711400.

PR #972, opened 2018-11-10
Item shortcut should have space after item

tecosaur:
This makes it Ctrl+L Ctrl+Enter:
 (a) slightly better to use
 (b) consistent with item-on-enter


jlelong:
I have just fixed it.

PR #979, opened 2018-11-13
fix the style of trimSelectContainer

tamuratak:
This is a patch to undefine the max-width of trimSelectContainer.  We don't have to do this for scaleSelectContainer because viewer.js dynamically sets an appropriate width. Without this patch, the select element is not appropriately displayed as follows.

### DIFF
--- a/viewer/latexworkshop.css
+++ b/viewer/latexworkshop.css
@@ -126,3 +126,14 @@ html #outerContainer.sidebarOpen > #sidebarContainer {
       display: none;
     }
 }
+
+#trimSelectContainer {
+    max-width: none !important;
+    width: auto !important;
+}
+
+#trimSelectContainer > select {
+    -webkit-appearance: button;
+    -moz-appearance: button;
+    appearance: button;
+}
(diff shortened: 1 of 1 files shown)


James-Yu:
You may notice that the height of this selector is different from the zooming one.

tamuratak:
I have set the height of selects. Please check.

tamuratak:
Another way of fixing this issue is setting a slightly larger width to trimSelectContainer, [link](https://github.com/James-Yu/LaTeX-Workshop/compare/master...tamuratak:fix_trim_select_style_b).  Please check.

James-Yu:
Actually this is because that the select items are set to `button`s. Why is that?

tamuratak:
I set them to `button` to remove the arrow of the select element provided by browsers. Otherwise, we see two arrows in the trim select as follows. Another arrow is provided by pdf.js.

It doesn't have to be `button`. It can be `none`.

James-Yu:
It is not consistent with the other selector, which set the width of select items to be ~20px wider. This should be a proper solution.

tamuratak:
pdf.js dynamically does that while taking locales into account. It is difficult for me to do that.

James-Yu:
Please see the commit above

tamuratak:
Great. 👍

PR #983, opened 2018-11-14
Change and unify some shortcuts

### LINKED ISSUE #975: Don't use Ctrl+Alt+[character] as a shortcut
### Description

Ctrl + Alt (same as Alt Gr) is used to write less-common characters on most keyboard layouts (https://en.wikipedia.org/wiki/AltGr_key). Using it as a shortcut is guaranteed to break those bindings for at least some users, which is why it should be strictly avoided.

### Log messages

N/A

### How to reproduce

N/A

### Expected behavior

User is able to normally type less common characters using their keyboard after installing this extension.

### Actual behavior

User has to disable / change keyboard shortcuts after installing this extension in order to type certain less common characters.

### Additional Information

N/A

James-Yu:
Resolve #975

@jlelong how do you think of the issue as well as the new plan? Maybe a notification is required to tell users that keybinds are changed.

### DIFF
--- a/SHORTCUT.md
+++ b/SHORTCUT.md
@@ -6,45 +6,51 @@ All commands are listed as <kbd>ctrl</kbd>+<kbd>some key</kbd>. If you're on Mac
 
 ## Extension Commands
 
+The `l` in <kbd>ctrl</kbd>+<kbd>l</kbd> stands for `LaTeX`.
+
 | Shortcut                                    | Does                               |
 | ------------------------------------------- | ---------------------------------- |
-| <kbd>ctrl</kbd>+<kbd>alt</kbd>+<kbd>b</kbd> | Builds current file                |
-| <kbd>ctrl</kbd>+<kbd>alt</kbd>+<kbd>c</kbd> | clean latex build files            |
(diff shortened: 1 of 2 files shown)


jlelong:
My first remark is that <kbd>ctrl</kbd>+<kbd>t</kbd> is already bound to _Show all Symbols_. 

I am really unsure of what to do with #975. There are some many extensions available for vscode defining tons of shortcuts that finding free combinations is really hard. Other extensions such as gitlens heavily use the <kbd>ctrl</kbd>+<kbd>alt</kbd> combination but they offer an alternate keymap. 

My idea would be
1. to introduce a new configuration variable `keymap-alternate`, with default value `true`
1. to only define the 5 <kbd>ctrl</kbd>+<kbd>alt</kbd> keybindings when `latex-workshop:keymap-alternate==true`. For most users, this would be fine and this will undoubtedly save us tons of bug reports.
1. to define more exotic keyboards such as <kbd>ctrl</kbd>+<kbd>l</kbd>, <kbd>alt</kbd>+<kbd>letter</kbd> for `letter` in  `b,c,v,j,x` when `latex-workshop:keymap-alternate==false`.

What do you think of it? I feel everybody should be happy with this.

Issue #980, opened 2018-11-13
Problem with code folding


Saphrosit:
### Problem with code folding

Code folding does not work as expected in certain situations.

### Log messages

No log needed to reproduce the issue. 

### How to reproduce

I am attaching a minimal working example (GitHub did not let me upload the file in .tex extension).
[mwe.txt](https://github.com/James-Yu/LaTeX-Workshop/files/2576243/mwe.txt)

### Expected behavior

All proof environments should be foldable.
Folding the first proof environment should result in the collapse of the lines 18-22.

### Actual behavior

Trying to fold the first proof environment results in the lines 18-43 to be collapsed. Also some proof environments are not foldable at all (third and fourth).

### Additional Information

VScode 1.29.0
LaTeX extension version 5.13.3
macOS 10.12.6

James-Yu:
Please check #984 #985

PR #984, opened 2018-11-14
re-implement env folding provider with stack

iamhyc:
Hi, I think this PR could solve the issue mentioned here #980 , which is introduced by folding provider with environment range.


neilsustc:
Smart idea! Then I believe the section-based folding can also be reimplemented with stack 😅

PR #985, opened 2018-11-14
Better folding

James-Yu:
This aims to resolve some new minor issues when the indentation-based folding was replaced by a handicraft folding logic @neilsustc. The new PR makes use of stacks to store env folding, and a line number list to handle section nestings. The env stack part was contributed by @iamhyc #984.

@neilsustc @Astrantia Please take a look.


neilsustc:
(I will do it if it is still not taken by someone else by this weekend 😂)

Issue #322, opened 2017-11-20
Add a dark theme to the pdf viewer


ismasou:
### Description

I just decided to start using VS code for my TeX and I'm in love with this plugin, the only reason I'm actually switching is because of the dark theme of VS code, and I was wondering if you could implement a dark theme for the viewer too.

### How to reproduce

The easiest way would be to just switch dark text into white and the white background black. The other colors should stay the same.

James-Yu:
I'm afraid that this won't be implemented. The proposed plan may only work well with b/w documents, but it may cause great problems for color ones.

henrikdahlberg:
What exactly is the issue? Colored documents will have their colors inverted as well, that's the entire point. I usually inverted everything on my screen, webpages for instance, and there are smart inverters that don't invert images. Color inversion is a feature that essentially every PDF viewer has implemented and it's annoying to have to use an external one when working with LaTeX Workshop.

James-Yu:
While I am still finding what you said is a must for PDF viewer in my installation of Adobe Acrobat Reader DC, I'd unfortunately say this feature is currently not on the list to implement.

bvibhav:
I usually use Adobe reader on Linux tiled side by side with VS Code. Ctrl+I and you have inverted colors.

tamuratak:
Related to mozilla/pdf.js/issues/2071.

fuzihaofzh:
I also hope this feature could be considered. I just wonder whether using the gray scale mode could simplify the development? Many pdf viewers (e.g. pdf expert) provide many background colors mode. In these modes, the document is converted into grayscale and then render the user-defined colors. I think this could make things simpler?

tamuratak:
See https://github.com/James-Yu/LaTeX-Workshop/wiki/View#latex-workshopviewpdfinvert

PR #986, opened 2018-11-14
add an invert color option for pdf viewer

tamuratak:
Hi,
this is a patch to add an invert color option for pdf viewer. Requested in #322. We can implement it by using the [filter CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/filter). 

Although I am not sure this is what users want, I hope this will help users who need this feature.

nov-14-2018 22-30-41

### DIFF
--- a/viewer/latexworkshop.js
+++ b/viewer/latexworkshop.js
@@ -277,3 +277,17 @@ window.addEventListener("pagerendered", () => {
     observer.observe(page, {attributes: true, childList: true});
   }
 }, {once: true});
+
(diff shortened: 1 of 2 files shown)


James-Yu:
I liked the feature very much, so there was a configuration item called `latex-workshop.view.pdf.invert` which did the same thing.

James-Yu:
Can you please check if the config serves the same functionality?

tamuratak:
I missed that. Thanks.

Issue #989, opened 2018-11-15
Intellisense Does Not Behave as Expected


SophtwhereDev:
This is not a bug but simply that Intellisense doesn't function as I would've liked it to. I'm trying to migrate from TeXstudio so some of my issues are based on the differences between how TeXstudio handles things and how LaTeX-Workshop handles things, but some restrict workflow.

# Issue 1
The first thing being that when I start an environment I prefer typing as far as `\begin{enu` and then depend on intellisense and press enter and what I get is simply `\begin{enumerate}` although I expect:

```latex
\begin{enumerate}
    \item Cursor Here
\end{enumerate}
```

Or at least both the begin and end blocks. This is what I'm used to and prefer from TeXstudio. I realised that to get both begin and end I have to type just `\begin` and press enter which I find restrictive.

# Issue 2
Next, I realised that Intellisense doesn't provide any information on the arguments required by functions, and no awareness of functions that support multiple arguments. For example, Intellisense only contains `\includegraphics{}` while I would prefer a second option for `\includegraphics{size}{file}`.

# Issue 3
I believe a lot of packages don't seem to have Intellisense, I say a lot but I just tested the packages cancel and tabularx. I'm not sure if this is due to me doing something wrong or due to shortcomings of LaTeX-Workshop.

# Issue 4
I realised there's no Intellisense option for `\section*{}` and the lower levels of the same command.

jlelong:
### Issue 1

Type BEN and `TAB` to complete, you will get exactly what you want. See [the snippets](https://github.com/James-Yu/LaTeX-Workshop/blob/master/snippets/snippets-doc.md).

### Issue 3

Contributions are welcome!

### Issue 4

You probably mean snippets.

SophtwhereDev:
## Issue 1
That's convenient but only works for the specific environments that already have snippets, it doesn't work if I want to create an empheq environment for example.

## Issue 3
It seems that TeXstudio handles auto-completion with .cwl files for packages, can't they be utilised by LaTeX-Workshop as well since they already exist? It's mentioned [here](http://texstudio.sourceforge.net/manual/current/usermanual_en.html#CWLDESCRIPTION).

I believe this would solve my Issue 2 as well.

## Issue 4
What I mean is that if I type `\sec`, I get the option to auto-complete to `\section` but not `\section*`

SophtwhereDev:
[...]

PR #992, opened 2018-11-17
Provide command intellisense for selected packages

James-Yu:
#989 

This PR creates a new python package to obtain cwl files from https://github.com/LaTeXing/LaTeX-cwl and parse the contents into `commands.json`, which is used to provide command intellisense.

@SophtwhereDev @tamuratak Can you please have a check?


James-Yu:
Also there is a new problem. It seems that `\` intellisense is now providing way too many entries that the purpose of intellisense is voided. Is there a good solution to that? Discussion welcomed!

jlelong:
@James-Yu I was about to post the same remark about far too many suggestions by intellisense. 

In particular, many commands have a common prefix, for instance there are more than 120 commands starting with `\cite`. More importantly I have the impression that intellisense is not case sensitive. Typing `\cite` actually suggests `\Cite` instead.

SophtwhereDev:
Apologies but I do not know how to get this PR into vscode to try testing it, I'd appreciate some guidance.

SophtwhereDev:
@James-Yu 
> Also there is a new problem. It seems that `\` intellisense is now providing way too many entries that the purpose of intellisense is voided.

I'm not sure how operation is given that I haven't been able to test this, but I assume that the .cwl files for all the packages populate Intellisense now? This would be counterproductive, I think it would be better if the preamble of the .tex file being edited could be checked for all `\usepackage{}` lines and only populate Intellisense entries for the packages that are currently being used in the document that's being edited.

Of course this is only valid if my assumption is valid.

James-Yu:
Even with this commit, I am still hesitated to merge the PR as the feature is really slowing down the intellisense. Maybe a config item to disable all? @jlelong

jlelong:
First @James-Yu what an impressive work 👍.

I have done very little testing but I do not really experience slow down in intellisense. 

However, I feel the list of completions is still too long. For instance, `natbib` adds 20 more items to the intellisense completion for `\cite` and `\cite` itself only appears in 11th position. This is related to the numerous issues in vscode repo on how the intellisense suggestions are sorted. In particular, some people argue that an exact match should appear first but this does not seem be the case yet and I have not found any conclusion on completion ranking.

I am wondering if aiming at having a complete suggestion list is worth it considering we already have a feature to automatically discover commands already used in a `.tex` file. The main issue is that some packages provide a lot of commands and we usually use very few of them so we do not want intellisense to be polluted with all the others.

Maybe we can wait a bit before merging and do some more tests. Someone may also come up with a new idea.

James-Yu:
I've come up with a better implementation idea. Will try next and see if things are better.

James-Yu:
Please have a check on the new implementation @SophtwhereDev @jlelong

SophtwhereDev:
@James-Yu 

I don't notice any drop in performance at all, the first .tex file I opened took about a 5 second delay but the next file I opened was instant. And everything seems to be working great.

Although intellisense still doesn't seem to help with the arguments, when there's multiple arguments.

Astrantia:
Great work @James-Yu Startup time indeed increases at least 10 times. Why does it still load if it's disabled in settings? You can see that after this PR, no matter if this is enabled in settings or not the startup times are still increased by a lot. 

If I'm not mistaken this only work on-demand right? If a package is not imported then intellisense for that package is not displayed right?

I would also love to get an added functionality of displaying Unicode versions in the intellisense tooltip and a little bit more info such as what package it comes from and etc. Can we discuss how to add this functionality too?

James-Yu:
Can you please confirm that you are on the latest commit? I did not observe increase in startup time, nor is there significant change on startup loading logic. If so, please also provide the startup performance information via `startup performance` command of vscode.

Astrantia:
@James-Yu That last commit broke unimath intellisense for me altogether. Startup is still 500ms from 80ms.

James-Yu:
The package auto-completion works. You need to enable the new config item `latex-workshop.intellisense.package.enabled`.

James-Yu:
5.16.0 live:
image

224530f:
image

It is not a significant increase on my modern potato. The increase is partially due to minor pre-loading changes. If others may have similar observations with mine, I'd merge the PR and release.

Astrantia:
@James-Yu Im suggesting
```
        data[segments[3]] = {
            'command': segments[3],
            'detail': segments[1],
            'documentation': segments[7].strip().capitalize()
        }
```

James-Yu:
Please PR. It is off-topic.

tamuratak:
Because I don't catch up the whole discussion, I give a minor comment only. I think commands starting with lowercase should come before ones starting with capital letters since we use the former ones more frequently in general. The following is a screenshot with the `natbib` package.

jlelong:
@tamuratak This is related to the long intellisense list problem we were mentioning in previous posts. Unfortunately, I do not think we can alter the order of intellisense suggestions from the extension. In particular, it seems impossible to rank an exact match as first.

jlelong:
@James-Yu It does not seem to change anything for the order of `\cite` list. Apparently, vscode reorders the list of suggestions based on some internals algorithms + some guidance from the user. See [here](https://code.visualstudio.com/docs/editor/intellisense#_locality-bonus) and [there](https://code.visualstudio.com/docs/editor/intellisense#_suggestion-selection).

This is partly good news anyway. If you set `editor.suggestSelection: recentlyUsedByPrefix` and you type `\cite` and explicitly choose `\cite` item as the completion, it will do it automatically on any following completions of `\cite`.

tamuratak:
@James-Yu, @jlelong, I missed that point. Sorry.

PR #1001, opened 2018-11-19
Update unimathsymbols.json

Astrantia:


Astrantia:
@James-Yu I don't know if these PRs make the cut, but I think these are very nice to have at least as a feature request.

PR #1005, opened 2018-11-21
(discuss) adding a support for an online changelog

NavyAdmiral:
My idea is to add a support for a webview changelog that users will be able to view within vscode. 

@James-Yu @jlelong @tamuratak et al, what are your thoughts?
https://inspiring-bohr-ceec26.netlify.com/

### DIFF
--- a/src/commander.ts
+++ b/src/commander.ts
@@ -357,7 +357,7 @@ export class Commander {
                                 break
                             case 'View LaTeX-Workshop extension change log':
                                 vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(
-                                    'https://github.com/James-Yu/LaTeX-Workshop/blob/master/CHANGELOG.md'))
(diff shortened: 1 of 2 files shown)


James-Yu:
I'm quite neutral to an external changelog site for better styling. Any ideas?

NavyAdmiral:
To be clear since this PR is misleading, I don't want the changelog button to open an external changelog website. I want to see webview inside vscode that will render the html, so we won't need to get a domain (another error in PR description)

jlelong:
Actually, the Visual Studio marketplace already proposes its own rendering of the [Changelog](https://marketplace.visualstudio.com/items/James-Yu.latex-workshop/changelog). Maybe, we could just put this [link](https://marketplace.visualstudio.com/items/James-Yu.latex-workshop/changelog) instead behind the _Open ChangeLog_ button.

Issue #366, opened 2018-01-05
Formatter does not respect settings for tabs as spaces


WorkingDevel:
Thank you for the integration of the formatter! But there is a little flaw with the intention.

If you format a *.tex file the indention from the line beginning is done always with real tabs. If you have set `"editor.insertSpaces": true` it should be indented by `"editor.tabSize"` spaces instead of the real tab stop character.

### How to reproduce

Using Windows 10 x64 with VSCode 1.19.1, latexindent 3.3 from texlive (2017-12-01),

0. You have following settings set:
```json
"editor.tabSize": 4,
"editor.insertSpaces": true,
"editor.renderWhitespace": "all",
```
1. Make this the solely content of a *.tex file:
```tex
\documentclass[paper=A4,paper=landscape,BCOR=0mm,9pt,twocolumn]{scrartcl}
\usepackage{fetamont}
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}

\begin{document}
\begin{tabular}{rl}
test1 & test2\\
\end{tabular}

  \begin{tabular}{rl}
  test1 & test2\\
  \end{tabular}
  
\end{document}
```
2. Run the formatter from context menu: `Format Document`
3. Take a look at the lines `test1 & test2\\`

### Expected behavior

lines indented with 4 spaces

### Actual behavior

lines indented with 1 tab stop character (`\t`)

### Log messages

nothing directly related to formatting, but started build afterwards...

```
[12:59:45] Parsing w:\Users\Christoph\nextCloud\_shared\Notizen\keymaps\test.tex for outline
[12:59:50] Root file remains unchanged from: w:\Users\Christoph\nextCloud\_shared\Notizen\keymaps\test.tex.
[13:00:01] Root file remains unchanged from: w:\Users\Christoph\nextCloud\_shared\Notizen\keymaps\test.tex.
[13:00:01] Root file remains unchanged from: w:\Users\Christoph\nextCloud\_shared\Notizen\keymaps\test.tex.
[13:00:04] Root file remains unchanged from: w:\Users\Christoph\nextCloud\_shared\Notizen\keymaps\test.tex.
[13:00:04] Root file remains unchanged from: w:\Users\Christoph\nextCloud\_shared\Notizen\keymaps\test.tex.
[13:00:07] Found root file from active editor: w:\Users\Christoph\nextCloud\_shared\Notizen\keymaps\test.tex
[...]

PR #1015, opened 2018-11-24
Formatter is not respecting tab and space settings due to wrong option

tete1030:
The formatter is still not respecting vscode's tab and space settings, despite it should have been fixed as in #366.

After digging, I found that the default formatting arguments for `latexindent` are 
```
['-c', '%DIR%/', '%TMPFILE%', '-y="defaultIndent: \'%INDENT%\'"']
```
However the double quotes in the last argument are preventing this option from working. Double quotes are only necessary when used in shell to indicate that the argument should not be separated by the space inside it, and are removed by shell when arguments are passed to process.

In this case the arguments are directly passed to process instead of through a shell, so double quotes will not be removed. I guess `latexindent` will get an option with key `"defaultIndent` and value `'%INDENT%'"`, and take it as an illegal argument. So I removed double quotes in the last argument. 
```
['-c', '%DIR%/', '%TMPFILE%', '-y=defaultIndent: \'%INDENT%\'']
```
The formatter is now working correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -848,7 +848,7 @@
             "-c",
             "%DIR%/",
             "%TMPFILE%",
-            "-y=\"defaultIndent: '%INDENT%'\""
+            "-y=defaultIndent: '%INDENT%'"
           ],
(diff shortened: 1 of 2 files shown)


tete1030:
@James-Yu Please consider reviewing this. It's a simple fix. If you need any further information or verification, please tell me.

Issue #1008, opened 2018-11-21
Bundle windows binary of synctex with the extension


James-Yu:
**Is your feature request related to a problem? Please describe.**
MikTeX does not ship synctex, which leads to many issue reports in the repository.

**Describe the solution you'd like**
It seems that only two small binary files, i.e., an `exe` and a `dll` is required for the functionality. We may bundle them in the package and include the directory in `cp.spawn` path env.

**Describe alternatives you've considered**
Hmm... Letting LaTeX newcomers to know that MikTeX does not have synctex is actually unnecessary.

**Additional context**
However, all my testing PCs are equipped with TeXLive. I may need some volunteer to test a potential implementation/PR. Also, whether we should bind them is still an open question, though I am a bit inclined to do so. Discussions welcomed.

tamuratak:
There is a parser [tdurieux/synctex-js](https://github.com/tdurieux/synctex-js) for *.synctex files written in JavaScript by @tdurieux. So, we might be able to implement the synctex command as a function in TypeScript. This allows us to execute the synctex command synchronously. By using this function, we might be able to implement a synchronous scroll although I am not a fan of the synchronous scroll.

tdurieux:
Technically it is completely possible I implemented this feature for BlueLatex.
I am also not a big fan of synchronous scroll.
What I did that I found pretty useful is to highlight the current line in the PDF.

tecosaur:
Jumping in on synchronous scroll, simply put I think some like it some don't. This seems like the sort of thing where some sort of setting would be a good idea ;)

jlelong:
I guess this issue has been outdated by @tamurak's [PR](#1030).
Set  `latex-workshop.synctex.synctexjs.enabled` to `true`.

Issue #936, opened 2018-11-01
Reverse Synctex dosen't work when directory contains Chinese characters


xlf605401969:
### Description

When directory name only contains ASCII characters, both direct and reverse synctex works fine. But when directory name contains Chinese characters, reverse synctex (from pdf to .tex) stops work wihle direct synctex still works.

### Log messages

<details>
<summary>LaTeX Workshop Output</summary>

Direct synctex logs (works fine)
[10:24:13] SYNCTEX command invoked.
[10:24:13] Found root file from active editor: c:\测试\test.tex
[10:24:13] Root file remains unchanged from: c:\测试\test.tex.
[10:24:13] Executing synctex with args view,-i,47:18:c:\测试\test.tex,-o,c:\测试\test.pdf
[10:24:13] Try to synctex c:\测试\test.pdf

Reverse synctex logs (does not work)
[10:25:42] Executing synctex with args edit,-o,1:62.42128603104212:537.3392392177541:c:\测试\test.pdf
[10:25:43] SyncTeX to file c:\����\test.tex
</details>

<details>
<summary>Developer Tools Console</summary>

Reverse synctex logs (does not work)
There is one error in the console., It seems that it is caused by encoding?

 ERR cannot open file:///c%3A/%EF%BF%BD%EF%BF%BD%EF%BF%BD%EF%BF%BD/test.tex. Detail: 找不到文件(file:///c:/����/test.tex): Error: cannot open file:///c%3A/%EF%BF%BD%EF%BF%BD%EF%BF%BD%EF%BF%BD/test.tex. Detail: 找不到文件(file:///c:/����/test.tex)
</details>

### How to reproduce

1. Put a .tex file under a directory with Chinese characters.
2. Build the .tex file.
3. Try reverse synctex from the pdf.

### Additional Information
Latex Version: TexLive 2017
System Version: Win10 x64 Home 1803
Windows Display Language: Simplified Chinese [中文(中华人民共和国)]

James-Yu:
It's a bug related to synctex. We cannot fix it from our side.

xlf605401969:
>           It's a bug related to synctex. We cannot fix it from our side.

Thanks for you reply. But I think synctex.exe works fine with Chinese characters. There are some tests. The terminal encoding is set to windows defaut (GB2313) here.
```
[...]

Issue #727, opened 2018-07-31
SyncTex doesn't work when file path contains Chinese character.


nathanatgit:
### Description

I intended to sync pdf to code from cursor but ctrl+click doesn't work when there is a Chinese charactor in path.
I don't know if it's my own setting issue or a common issue.

When the file path contain Chinese character, Latex workshop output shows garbled path when executing Synctex.

```
[12:35:33] Open PDF tab for c:\Users\Nathan\OneDrive\Lab\研二\SingleNodeAF\LatexTemplate\bare_jrnl.pdf
[12:35:33] Found root file from active editor: c:\Users\Nathan\OneDrive\Lab\研二\SingleNodeAF\LatexTemplate\bare_jrnl.tex
[12:35:33] Root file remains unchanged from: c:\Users\Nathan\OneDrive\Lab\研二\SingleNodeAF\LatexTemplate\bare_jrnl.tex.
[12:35:35] Preview PDF file: c:\Users\Nathan\OneDrive\Lab\研二\SingleNodeAF\LatexTemplate\bare_jrnl.pdf
[12:35:47] Executing synctex with args edit,-o,1:378.2123893805309:464.33628318584067:c:\Users\Nathan\OneDrive\Lab\研二\SingleNodeAF\LatexTemplate\bare_jrnl.pdf
[12:35:47] SyncTeX to file c:\Users\Nathan\OneDrive\Lab\�ж�\SingleNodeAF\LatexTemplate\bare_jrnl.tex
[12:35:47] Executing synctex with args edit,-o,1:379.1150442477876:467.0442477876106:c:\Users\Nathan\OneDrive\Lab\研二\SingleNodeAF\LatexTemplate\bare_jrnl.pdf
[12:35:48] SyncTeX to file `c:\Users\Nathan\OneDrive\Lab\�ж�\SingleNodeAF\LatexTemplate\bare_jrnl.tex
```

Ctrl+click then code editor remains at the same place.
It seems there is a wrong recorded path which is already garbled in .synctex.gz.

After i move the file to desktop and rebuild .tex file, Ctrl+click works.

### How to reproduce
Move .tex to path containing chinese charactor.

### Additional Information
.synctex file content with issue
```
SyncTeX Version:1
Input:1:c:/Users/Nathan/OneDrive/Lab/�ж�/SingleNodeAF/LatexTemplate/bare_jrnl.tex
[...]

Issue #495, opened 2018-03-26
Ctrl+Click not work when file path contains CJK characters


LukeLike:
### Description

When I press ctrl+click in the pdf file whose path contains CJK characters, it does not jump to .tex file. It seems that the CJK characters in the path is not recognized because in the log for LaTeX Workshop the CJK characters are garbled. However sync from .tex file to pdf works fine.

Is there a way to fix this?

James-Yu:
It's not related to the extension, since LaTeX and its tools generally do not respect CJK chars. I'm not aware of solutions.

LukeLike:
Well, I don't think this is an issue for synctex tool. Because when I run the synctex command directly in the command line, the output seems correct. I guess this is due to some encoding issues when reading the stdout.

![screenshot](https://user-images.githubusercontent.com/8063808/37946836-2616fac0-31ba-11e8-8177-16debe57091a.PNG)

PR #1030, opened 2018-11-30
Support a builtin synctex functionality.

tamuratak:
This is a patch to support a builtin synctex functionality, which solves #1008. By setting latex-workshop.synctex.synctexjs.enabled to true, a builtin synctex function will be used instead of the synctex command. Although I don't think we should use the function as a default, it works somehow.


James-Yu:
Awesome! Really look forward to use this one as default and remove all synctex binary related code.

Is it possible to move the synctexjs source to a new folder `src/lib` since it is a 3rd-party contribution? Also, is this javascript implementation fast enough to replace the binaries shipped with LaTeX distributions?

tamuratak:
I have moved `synctexjs.ts` to `src/lib`. As far as I tried, it is fast enough. The main issue is the accuracy of the backward (pdf to source) synctex. We need to improve it.

tamuratak:
Yes, we still need synctex files. We must compile tex files with `-synctex=1`.

tamuratak:
Now, this PR works well with various latex engines, latex, pdflatex, xelatex, lualatex, and uplatex, at least for `\documentclass{article}` as far as I tested. Testing for other classes and packages are welcome.

tamuratak:
Now, using `iconv-lite`, we can invoke synctex for a tex file whose name containing Chinese characters. This solves #495, #727, and #936.

James-Yu:
Awesome work. Will merge and subsequent fixes if any may go another branch.

James-Yu:
@tamuratak When you think ready, please make this feature default in the configuration. After that we can work on removing synctex binary codes from the extension.

Astrantia:
Is it me or is this version of synctex faster ?

tamuratak:
@James-Yu, thanks.

@Astrantia, please try. I have not measure it. If you find a case in which the built-in synctex is slow, please tell me.

Astrantia:
@tamuratak I tried it and I think that javascript version is faster.

tdurieux:
I'm really happy to see that my code is used it this really good plugin :)

Well done!

tamuratak:
@tdurieux, I am glad to hear you say that. Thank you for your great coding.

Issue #661, opened 2018-06-07
Reference autocompletion should replace whole reference with found reference


chbndrhnns:
### Description

When I use the autocomplete feature for references, I often start typing what I think the reference is called and hit enter once it appears in the dropdown list. Then, what I already typed is replaced except if I enter something like `tab:table-bla` which is autocompleted to `tab:tab:table-bla` and leads to a reference error.

### Log messages

None

### How to reproduce

1. Write `\label{tab:reqs-lang-status}` in latex document
2. write `\autoref{tab:re` to see matching autocomplete entries
3. Hit Enter to accept it

### Expected behavior

The correct reference is put after \autoref{

### Actual behavior 

The reference looks like `tab:tab:reqs-lang-status`

James-Yu:
This is a known issue. Unfortunately, labels with `-` or `:` are interpreted as multiple labels by vscode. This seems not to be a programmable behavior from the extension side. You are encouraged to submit an issue on vscode's repo.

tl,dr; sorry we cannot fix it on our side.

jlelong:
It does actually work as expected on my installation even with labels containing `:` or `-`.

The only way I managed to reproduce the problem is 
1. start typing `\autoref{tab:re` 
1. Delete at least one character
1. Trigger intellisense using `^space`
1. Select `tab:reqs-lang-status` in the list
You end up with `\autoref{tab:tab:reqs-lang-status}`.

Can you tell use what is your configuration related to `suggestions`?

chbndrhnns:
I noticed the same and was sure that the issue existed for months. I upgraded to VS Code 1.24 today. Maybe this resolved the issue?

chbndrhnns:
I could trigger the behaviour, as well, by putting "moving the cursor left or right" as the second step in your reproduction step list.

Issue #121, opened 2017-04-20
Minor issue with autocompletion of reference


angan06:
Dear James, 

When I want to use the auto-completion with a reference, for instance:  \ref{s:Title section}, 
I usually start writing "\ref{s:T " then select my reference in the list appearing and press enter. 

However, when doing like that, code studio completes by \ref{s: s:Title section}. I guess that it is an issue with the " : ".  
Is there a workaround that allow to keep the same notation for my reference ?

Best regards.

James-Yu:
This should be an upstream problem with vscode. The extension cannot control or modify the intellisense behavior. I suppose the only "workaround" is to avoid using characters that has special meanings, e.g., `,.:'`. Or just type `\ref{sT` for the said entry.

Hope that can answer your question. Sorry no solution we can add to the extension.

angan06:
Thanks for the answer.

PR #1032, opened 2018-12-02
Fix the completion of refs.

tamuratak:
This is a patch to solve issues on the completion of refs reported in #121 and #661.  Now the completion works well even if labels include `:`, `-`, `.`, and `_`. To pass arguments to Reference.provide, we have to rewrite Completer.completion.

---

Before.
dec-02-2018 11-21-35

---

After.
dec-02-2018 11-12-48

### DIFF
--- a/src/providers/completer/input.ts
+++ b/src/providers/completer/input.ts
@@ -16,7 +16,7 @@ export class Input {
         this.extension = extension
     }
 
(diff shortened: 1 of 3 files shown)


jlelong:
Thanks for the PR. I will have a look in detail.

PR #1037, opened 2018-12-03
automatically remove %TMPDIR%

P4SSER8Y:
The generated temporary folders are not removed currently. I added a method to auto-remove them when the extension is deactivated.


P4SSER8Y:
This is not safe enough. I should use the method provided by tmp.

PR #1039, opened 2018-12-03
add: reveal output folder in OS; auto-remove %TMPDIR% even if not empty

P4SSER8Y:
Add a shortcut to reveal the output folder in OS. Using vscode command `revealFileInOS`.


P4SSER8Y:
The %TMPDIR% should be auto-removed even if it is not empty, because it is always re-created.

Issue #1020, opened 2018-11-26
Allow multiple pdf viewer tabs of same document


edgardozoppi:
Right now there is no way to open more than one pdf viewer tab. I would like to split the tab to see different pages of the same pdf document at the same time.

Thanks!

James-Yu:
Actually you can now. Click the split pane button for the effect.

edgardozoppi:
I cannot make it work for me. I right-click on the pdf viewer tab and select one of the split options of the context menu like Split Up, but the new tab (in this case above of the original one) is empty.

I think you should be able to reproduce this issue. I have the latest version of both VS Code and your extension.

zoehneto:
@James-Yu I have the same problem when using the latex workshop viewer in the browser. Currently the root file is used as the key for the registered clients so even if i create a second viewer in the browser (by duplicating the tab of an existing viewer) this will register the viewer and since it shows the same file the websocket is overwritten. So now I have a second viewer but the first one doesn't get notified about changes.

Apart from that it is also impossible to launch more than one viewer per root file, since latex workshop checks if there is already a client registered and then just refreshes that client.

PR #1063, opened 2018-12-11
Allow multiple viewers for the same document

zoehneto:
This PR fixes issue #1020 by creating an additional viewer every time the `View LaTeX PDF File` command is invoked or the respective button is pressed.


James-Yu:
Thanks for the contribution. As this might be a potential ground breaking change, I may need some time to test it. Will do so soon after I dealt with my in real life matters (which has taken all my time for weeks).

James-Yu:
Actually no 😂😂 Will do so in the next day or two. Trust me (TM).

Issue #1050, opened 2018-12-08
PDFs not generated from latex do not render


jerlich:
**Describe the bug**
When I try to open a PDF not in a latex folder or generated from latex workshop, it does not render.
**To Reproduce**
Download a random PDF from wherever. Try to open. 

**Expected behavior**
I expect the PDF browser to open.

**Logs**

[2018-12-08 12:39:41.495] [renderer1] [error] Illegal argument: Error: Illegal argument
    at Object.t.illegalArgument (file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:174:917)
    at file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:1634:281
    at file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:1582:790
    at t.e.invokeFunction (file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:1422:693)
    at t._tryExecuteCommand (file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:3041:736)
    at file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:3041:458
    at n.Class.define.cancel.then (file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:169:517)
    at t.executeCommand (file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:3041:432)
    at e.$executeCommand (file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:3204:528)
    at t._doInvokeHandler (file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:3278:1009)
    at t._invokeHandler (file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:3278:675)
    at t._receiveRequest (file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:3277:262)
    at t._receiveOneMessage (file:///Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.main.js:3276:126)
[...]

PR #1067, opened 2018-12-13
add a delay to open pdf files as a work-around.

tamuratak:
This is a patch to solve #1050. @jerlich,  @jlelong, please try.


jlelong:
Unfortunately, it does not help for me. vscode still displays a binary of the file.

PR #1069, opened 2018-12-14
add selinux label for docker volume mounts

tobikris:
On Fedora the docker build scripts fail because of selinux. This PR adds the necessary selinux labels for the docker volume mounts.

### DIFF
--- a/scripts/latexindent
+++ b/scripts/latexindent
@@ -1,2 +1,2 @@
 #!/bin/sh
-docker run -i --rm -w "$(pwd)" -v "$(pwd):$(pwd)" $LATEXWORKSHOP_DOCKER_LATEX latexindent "$@"
+docker run -i --rm -w "$(pwd)" -v "$(pwd):$(pwd):Z" $LATEXWORKSHOP_DOCKER_LATEX latexindent "$@"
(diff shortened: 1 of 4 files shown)


James-Yu:
I am not familiar with linux commands. Will this change break the compatibility on other distributions?

tobikris:
Ok, I will try to test this when I find the time.

scripts/latexindent
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-docker run -i --rm -w "$(pwd)" -v "$(pwd):$(pwd)" $LATEXWORKSHOP_DOCKER_LATEX latexindent "$@"
+docker run -i --rm -w "$(pwd)" -v "$(pwd):$(pwd):Z" $LATEXWORKSHOP_DOCKER_LATEX latexindent "$@"
```
dminca: what's the point of `:Z` when mapping the Docker volume?

The `-i` passed is just for executing the `latexindent "$@"` ? Questioning here if we need `-t` also (terminal)

tamuratak:
VS Code supports Docker with [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) now. Please try it.

Issue #1077, opened 2018-12-20
External pdf viewer default command is wrong for linux/macos


L0g4n:
**Describe the bug**
Trying to open the PDF with an external viewer on linux/macos does not open it.
**To Reproduce**
Click the "Open PDF" button in the editor with the tex file opened.
**Expected behavior**
The pdf is opened in the user's preferred PDF viewer.
**Additional context**
I do not provide logs since I think I know what the problem is. The specified command for opening the file in the external pdf viewer tries to open a specific windows PDF viewer with the file.
To fix this on linux and macos you have to overwrite this command:
`"latex-workshop.view.pdf.external.command": {  
        "command": "xdg-open", // for macOS you have to use open  
        "args": [  
            "%PDF%"  
        ]  
    }`  
So you could change the default setting for the command in the source code to "xdg-open" on linux or "open" on macOS respectively.

L0g4n:
I offer to fix that myself, but since I am not a JS/TS guy and not  familar with writing VSCode extensions I did not know where to look.

jlelong:
The variable `latex-workshop.view.pdf.external.command` is defined in [package.json](https://github.com/James-Yu/LaTeX-Workshop/blob/master/package.json#L654), but unfortunately I do not think you can have the default value depend on the OS. 

I guess the best we can do is to test if `latex-workshop.view.pdf.external.command` is non empty in [viewer.ts](https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/components/viewer.ts#L140) and then tweak the command depending on the os.

L0g4n:
@jlelong if that is the case you can ignore my PR #1078. I thought since you can use those labels for OS specific tasks you could use them in package.json.

L0g4n:
@jlelong The thing is, how do other VSCode extensions realize OS specific default commands?

jlelong:
Do you have an example of such an extension?

L0g4n:
https://github.com/Microsoft/vscode-python/blob/master/package.json

PR #1078, opened 2018-12-20
Add first try for OS specific commands

L0g4n:
See #1077.
Please check if that would work, since I do not know if "windows", "linux" and "osx" are valid in these cases here, I know you can use those labels for **tasks** in vscode but I do not know if you can use them in normal commands.

### DIFF
--- a/package.json
+++ b/package.json
@@ -654,10 +654,24 @@
         "latex-workshop.view.pdf.external.command": {
           "type": "object",
           "default": {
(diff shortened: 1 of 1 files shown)


jlelong:
No it cannot work as it is without adapting the `ts` code reading the variable `latex-workshop.view.pdf.external.command`, which is defined as a JSON object itself.

L0g4n:
As I suspected.

PR #1079, opened 2018-12-21
Improve outdir handling

zoehneto:
This PR improves the handling of output directories by making 3 changes:
- Documenting the availability of placeholders when setting `latex-workshop.latex.outputDir`
- Adding a %OUTDIR% placeholder for the toolchain steps
- Adding the `-outdir` option to the latexmk tool configuration so that output directories work out of the box if a user configures `latex-workshop.latex.outputDir`

I also considered adding `-output-directory` to the pdflatex tool configuration but since bibtex doesn't support it I did not make that change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,7 @@ As you may notice, there is a mystic `%DOC%` in the arguments. Symbols surrounde
 | `%DOCFILE%` | The LaTeX root file name without `.tex` extension                                   |
 | `%DIR%`     | The LaTeX root file path                                                            |
 | `%TMPDIR%`  | A temporary folder for storing ancillary files, and will be auto-removed when exit. |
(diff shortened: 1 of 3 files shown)


jlelong:
Adding the option `-outdir=%OUTDIR%` to `latexmk` in [package.json#L506](https://github.com/James-Yu/LaTeX-Workshop/blob/master/package.json#L506) will break the recipe when `%OUTDIR%` is empty, why is the case whenever `latex-workshop.latex.outputDir` is not set.

zoehneto:
Actually the default is `./` (see [package.json#L548](https://github.com/James-Yu/LaTeX-Workshop/blob/1128dc9ed85dc9be3f58f235b966f507ff8760af/package.json#L548)), I have tested it with that default and it works for me (using texlive 2018.4 on linux). It would only break if someone were to manually set the outdir to be empty (which would probably break other checks in the code).

PR #1083, opened 2018-12-23
Use wiki sidebar to list features

### LINKED ISSUE #1072: Merge README and Wiki
As the README is used by Windows marketplace and most people landing on the github page probably only look at the README, it would be worth trying to merge the wiki in the README.

I imagine that the README could be a concatenation of the wiki pages or so. Any thoughts on that?

jlelong:
This closes #1072.
Any comments / suggestions?


tecosaur:
One comment from me. I'd be tempted to leave the 'features' bit so that people browsing and looking at the README.md on the vs code marketplace can get a good idea / preview of the sorts of things this extension is capable of.

jlelong:
@tecosaur I would be tempted to agree but the _Features_ list is rather incomplete and maintaining an up to date list as well as the README is quite a pain.

A middle of the road approach would be to keep the existing visual entries from the list (the ones with screenshots) and we may rely on PRs for further update.

I will give it a try.

PR #1128, opened 2019-01-20
New multi-viewer implementation

James-Yu:
@zoehneto Please have a check if this implementation provides the same functionalities compared with #1063, except for the open-in-new default behavior which still needs some discussion.

### DIFF
--- a/src/components/viewer.ts
+++ b/src/components/viewer.ts
@@ -143,7 +143,22 @@ export class Viewer {
     openExternal(sourceFile: string) {
         const pdfFile = this.extension.manager.tex2pdf(sourceFile)
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
-        const command = JSON.parse(JSON.stringify(configuration.get('view.pdf.external.command'))) as ExternalCommand
+        let command = JSON.parse(JSON.stringify(configuration.get('view.pdf.external.command'))) as ExternalCommand
+        if (!command.command) {
+            switch (process.platform) {
+                case 'win32':
+                    command = {'command': 'SumatraPDF.exe', 'args': ['%PDF%'] }
+                    break
+                case 'linux':
+                    command = {'command': 'xdg-open', 'args': ['%PDF%'] }
+                    break
+                case 'darwin':
+                    command = {'command': 'open', 'args': ['%PDF%'] }
+                    break
+                default:
+                    break
+            }
+        }
         if (command.args) {
             command.args = command.args.map(arg => arg.replace('%PDF%', pdfFile))
         }
--- a/src/components/viewer.ts
+++ b/src/components/viewer.ts
@@ -24,7 +24,15 @@ export class Viewer {
         this.extension = extension
     }
 
-    refreshExistingViewer(sourceFile: string, viewer?: string) : boolean {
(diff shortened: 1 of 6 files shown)


src/components/viewer.ts
```diff
@@ -180,22 +143,7 @@ export class Viewer {
     openExternal(sourceFile: string) {
         const pdfFile = this.extension.manager.tex2pdf(sourceFile)
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
-        let command = JSON.parse(JSON.stringify(configuration.get('view.pdf.external.command'))) as ExternalCommand
```
zoehneto: This should probably not be removed. It was added in commit d9bb923506a987021b47bb45cde2d27f6f7bbed4 in order to have a good default value on all platforms because specifying a platform specific default configuration isn't supported by vscode.

jlelong: I confirm this should be kept along with the following `switch (process.platform)` to set a default value for the external viewer depending on the os.

src/components/viewer.ts
```diff
@@ -109,14 +72,14 @@ export class Viewer {
         return url
     }
 
-    openViewer(sourceFile: string) {
-        const url = this.checkViewer(sourceFile, 'viewer')
+    openBrowser(sourceFile: string, newViewer: boolean = false) {
+        const url = this.checkViewer(sourceFile, 'browser', newViewer = newViewer)
```
zoehneto: You are missing the `respectOutDir` parameter before `newViewer` here. Without that change `newViewer` actually sets the `respectOutDir` parameter of the function, meaning that you can't open a viewer in the browser if you have an outDir specified.

zoehneto:
Apart from the comments above this implementation seems to work and be functionally equivalent to my PR. 

For me the question is whether we want a specific command for opening in a new viewer or whether it would be more convenient to have a setting for it (which then raises the question what the default behavior should be).

src/components/viewer.ts
James-Yu: Never checked that. I copied the code from a too old commit.

James-Yu:
Any discussion and suggestion on the issue raised above?

James-Yu:
Another small issue I can come up with is that setting the default behavior to new tab prevents users from manually refreshing the existing viewers except for building the project. 

I had an interesting workflow which can suffer from the issue: two root files in the same folder, `a.tex` and `b.tex`. I made edits in `a.tex`, and `b.tex`'s content can be updated by a script, then `b.tex` is compiled.

Before the change, I may switch to `b.tex`, click the icon, get the new pdf displayed in the same viewer. After the change, it seems not possible to refresh the same viewer. Either I need to close the existing one, or make changes and save `b.tex`, or find the build command from palette/side bar. This behavior can be sometimes annoying, though I admitted that 1) the workflow is quite strange 2) the previous actions are also not elegant. I'm yet to find a solution to it.

zoehneto:
Maybe we just need a `Refresh Viewers` command / button and make that independent of the 'open pdf in new viewer' behavior.

James-Yu:
I agree. This looks better. Any other potential issues w.r.t. the PR?

zoehneto:
No, looks good and seems to work as intended (both the multi viewer and refreshing all viewers).

PR #1131, opened 2019-01-21
Resolve #1117 New linter config

### LINKED ISSUE #1117: Make chktex config on save/on demand instead of an interval
As far as the integration of ChkTex  (#39) is greatly appreciated, I was thinking that executing a linter should be an on-demand action, not based on a time interval (especially since the default config sets a 300 ms interval, which is IMHO way overkill). It is not high priority factor, but it could improve overall performance (not that the extension is suffering from it, but any improvement should be welcome).

I think linter config in other extensions are an example: 

`"python.linting.lintOnSave": true,`
`"php.validate.run": "onSave",`

James-Yu:
Is this one good? @jlelong

### DIFF
--- a/package.json
+++ b/package.json
@@ -727,10 +727,11 @@
           "default": true,
           "markdownDescription": "Enable using a builtin synctex function. The command set in latex-workshop.synctex.path will not be used."
         },
-        "latex-workshop.chktex.enabled": {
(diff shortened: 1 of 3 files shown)


jlelong:
Looks great 👍

PR #1144, opened 2019-01-25
Combine config items

James-Yu:
This is an initial PR to merge the two configs to one. This may reduce the clutter in the config of this extension.

I also have a plan to merge `latex.autoBuild.onSave.enabled` with `latex-workshop.latex.autoBuild.onTexChange.enabled`, but use the experience here.

One more thing that makes me hesitate is on whether a `latex.clean.enabled` is required, just like what we did for `chktex.enabled` and `chktex.run`. Also, whether `latex.clean.enabled` should be renamed to something like `latex.autoClean.enabled`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -569,15 +569,11 @@
           "default": [],
           "markdownDescription": "List of directories where to look for `.bib` files.\nAbsolute paths are required. This setting is only used by the intellisense feature, you may also need to set the environment variable `BIBINPUTS` properly for the LaTeX compiler to find the `.bib` files."
         },
-        "latex-workshop.latex.autoBuild.onSave.enabled": {
-          "type": "boolean",
-          "default": true,
-          "markdownDescription": "Build LaTeX after saving LaTeX source file.\nThis property defines whether LaTeX Workshop will execute the LaTeX toolchain command(s) to build the project after new LaTeX contents are saved."
-        },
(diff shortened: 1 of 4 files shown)


jlelong:
Although having both `.enabled` and `.run` seems canonical for linting, this two way approach does not apply to other features.

Renaming `latex.clean.enabled` to `latex.autoClean.enabled` would be more meaningful. I guess it would be worth integrating this in your PR.

I also vote for merging `latex.autoBuild.onSave.enabled` with `latex-workshop.latex.autoBuild.onTexChange.enabled` in a single variable using an enum.

zoehneto:
If I understand the way the vs code settings work correctly, your change would mean that one can set `latex-workshop.latex.clean.run` either to `onBuilt` or to on `onFailed` but one can't have both, which some users may want.

James-Yu:
What do you mean by having both? In the current setting, onBuilt will cover cases in onFailed. I suppose no one would clean the files on successful builds but not on failed ones, which cannot be covered.

James-Yu:
I think it's good to be merged. Any issues undiscovered?

jlelong:
Once the PR is merged, can you also propagate these changes to the wiki?

James-Yu:
Surely will do so😂

PR #1146, opened 2019-01-25
Add ability to increment/decrement sectioning

tecosaur:
### Case One

Every so often I want to change the 'sectioning level' is a section of my document. Sometimes I start writing content thinking it will be a section, then half an hour later realise that it should really be a subsection.
So, within a certain region, I want to downgrade all `sections` to `subsections`, `subsections` to `subsubsections` etc. This is usually done via a series of 3-5 scoped text replacements. Not the easiest.

This is something I infrequently do, but every time I do I wish it were a less involved process.

### Case Two

Quite frequently I create a `paragraph`, only to think it should be a `subparagraph`; or a `subsubsection`, only to change it to a `subsection`.
This is quick to change, move cursor, either type `sub` or press <kbd>del</kbd>/<kbd>backspace</kbd> three times. However, this adds up, and is what I currently consider to be mildly irritating.

## What this PR does

This PR adds a function to `commander.ts` - `shiftSectioningLevel` which operates on a line/selection and takes one argument, indicating whether the selection/line should have all sectioning within it 'incremented' or 'decremented'; significantly simplifying the processes outlined previously.

Likewise, there are two commands added to `main.ts`:
 - `latex-workshop.increment-sectioning`
 - `latex-workshop.decrement-sectioning`

I have also added keyboard shortcuts for this:
 - <kbd>ctrl</kbd>+<kbd>alt</kbd>+<kbd>]</kbd> - Increments Sectioning
 - <kbd>ctrl</kbd>+<kbd>alt</kbd>+<kbd>[</kbd> - Decrements Sectioning

## What I Haven't Added

At the moment when this command is applied to a selection, the selection disappears when the contents is replaced. I think this feature would be better if that selection would remain, or if a line had this applied

### DIFF
--- a/package.json
+++ b/package.json
@@ -337,13 +337,25 @@
       {
         "key": "ctrl+alt+]",
         "mac": "cmd+alt+]",
-        "when": "editorTextFocus && !editorReadonly && editorLangId == 'latex'",
+        "when": "editorTextFocus && !editorReadonly && editorLangId == 'latex' && !latex-workshop:altkeymap",
         "command": "latex-workshop.increment-sectioning"
       },
       {
         "key": "ctrl+alt+[",
         "mac": "cmd+alt+[",
-        "when": "editorTextFocus && !editorReadonly && editorLangId == 'latex'",
+        "when": "editorTextFocus && !editorReadonly && editorLangId == 'latex'&& !latex-workshop:altkeymap",
+        "command": "latex-workshop.decrement-sectioning"
+      },
+      {
+        "key": "ctrl+l ctrl+]",
+        "mac": "cmd+l cmd+]",
+        "when": "editorTextFocus && !editorReadonly && editorLangId == 'latex' && latex-workshop:altkeymap",
+        "command": "latex-workshop.increment-sectioning"
+      },
+      {
+        "key": "ctrl+l ctrl+[",
+        "mac": "cmd+l cmd+[",
+        "when": "editorTextFocus && !editorReadonly && editorLangId == 'latex'&& latex-workshop:altkeymap",
         "command": "latex-workshop.decrement-sectioning"
       },
       {
@@ -693,7 +705,9 @@
           "type": "object",
           "default": {
             "command": "",
-            "args": [ "" ]
+            "args": [
+              ""
+            ]
           },
           "markdownDescription": "The command to execute when using external viewer.\nThis function is not officially supported. %PDF% is the placeholder for the absolute path to the generated PDF file."
         },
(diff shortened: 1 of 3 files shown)


jlelong:
Nice feature! I am afraid that the shortcut `ctrl+alt+]` is not very convenient. Actually, on a French Mac keyboard, you cannot enter it all, see [the wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ#I-cannot-use-ctrlalt-in-a-shortcut). I would be good to also provide an alternative shortcut as `ctrl+l -` / `ctrl+l shift+-`.

tecosaur:
I keep on forgetting about those other keyboard layouts!
I've added some alternative keybindings.
 - <kbd>ctrl</kbd>+<kbd>l</kbd>  <kbd>ctrl</kbd>+<kbd>]</kbd>
 - <kbd>ctrl</kbd>+<kbd>l</kbd>  <kbd>ctrl</kbd>+<kbd>[</kbd>

tecosaur:
I've gone with <kbd>[</kbd> and <kbd>]</kbd> because to me it made sense for it to be similar to indenting; just instead of changing the indent level, the sectioning level.

Also, any ideas regarding the last part of my commit -
Do you have any idea how to make a text selection with the vscode API?

jlelong:
For the alternate shortcuts, I suggest to use either `ctrl+l [`but not `ctrl+l ctrl+[` because `[` is typically accessed via `ctrl+alt` on some layouts.

You have an example in [envpair.tex#L182](https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/components/envpair.ts#L182) of how to use vscode api to select text.

tecosaur:
I see. I kept the <kbd>ctrl</kbd> because I know I typically find shortcuts that have me keep the modifier keys easier to execute quickly and accurately. If you think it's better I'd be happy to switch it over.

tecosaur:
That makes this feature a bit nicer :grin:  —  thanks for the example.

Please do confirm if you want me to change to keybinding or leave as is, I'm happy either way.

jlelong:
Considering we define two shortcuts and the first one uses the <kbd>ctrl</kbd>+<kbd>alt</kbd> combination, I think we should make sure the alternate one works for every keyboard layout. In that respect, <kbd>ctrl</kbd>+<kbd>l</kbd> <kbd>[</kbd> is safer.
Anyway, we can change it anytime if we feel we have made a bad choice.

tecosaur:
There we go, now that's what happens.

That should do it :smiley:

Issue #1152, opened 2019-01-27
[Idea] Progress Bar


tecosaur:
## Progress

 - [x] Simple Regex
 - [ ] Robust Regex
 - [x] Intercept stdout and get page no.
 - [x] Adding status bar
 - [x] Get total pages
 - [x] Visual refinement
 - [ ] Track speed of page-generation
 - [ ] Generate ETA

## Background

There's something deeply unsatisfying about having no idea when my latex compilation is soon to finish.

Should I make myself a cup of coffee or just wait an extra 20s?
Who knows?

I had previously searched for such a thing at not had much luck, and as such dismissed the idea as two-hard-to-do.
However I recently came across this page stackexchange answer: https://tex.stackexchange.com/a/211572/167605 that made me realise that as pages are completed, that info is sent to the stdout!

## Idea

1. Get total page count from compiled PDF
2. When compiling pass output through a function that identifies page-complete-info (I think it's something like `[[5]]` from memory, need to check)
3. Keep track of last created page
4. Set progress = last created page / total pages
5. Update progress bar

## Issues

This only works when:
. (a) There's a PDF
. (b) The change doesn't intoduce too many page numbers

Also page compilation time varies a fair bit...

### However

I think this is better than nothing, and for documents with > 20 pages I imagine it could be relatively handy.

## Need to do

1. Find more spare time
2. Locate example method of:
  (a) Intercepting the latexmk stdout **⟵ HelpWanted No.1**
  (b) Adding a progress bar (or similar element) to the the bottom vscode bar **⟵ HelpWanted No.2**
3. Code it up

@James-Yu, @jlelong, @tamuratak -  I'd be interested in hearing your thoughts.

tecosaur:
Hmm. Well it looks like a good first step would be to see if I can get regex that works for ***all*** cases. To that end would you be able to point me to any good example documents / logs with those sorts of things so I can be more confident I'm no missing several out?

tecosaur:
Well, so far I have made modifications that work for a simple case (the single 12 page document I have happen on hand atm).

Having tried this... I like it. I ***really*** like it.

![image](https://user-images.githubusercontent.com/20903656/51872839-bd326e00-2352-11e9-938f-55336d418e71.png)

I've added a second `vscode.StatusBarItem` (to the right) that atm just shows what page it is at.
The regex I've used is quite simple, but for this document, it works. Further testing is absolutely needed.

The main things I could do with help on ATM are:
[...]

PR #1157, opened 2019-01-30
[Testers Wanted] Compilation progress

tecosaur:
A follow up to #1152 

image

## Areas where Help Wanted
**Ordered by priority**

1. Testing that all pages come up
2. Identifying regex to determine when preamble finished
3. Howto move webview source into seperate `.html` file
4. Function to calculate ETA

## Progress so far

 - [x] Simple Regex
 - [x] Robust Regex
 - [x] Intercept stdout and get page no.
 - [x] Adding status bar
 - [x] Get total pages
 - [x] Visual refinement (status bar)
 - [x] Track speed of page-generation
 - [x] Compilation info webview panel*
 - [x] Visual refinement 2 (graph)

*useful for identifiying slow pages etc.

## Not quite ready yet...

This is very much in need of testing / review, Specificly.
 - [x] Testing against more tex documents (re: page number regex)
   - 12 page simple doc ✓
   - 18 page doc with tikz ✓
   - 90 page multi-file doc ✓
 - Identifying any other stdout info lines that should be used (if they exist)
   - [x] latexmk: rule started
   - [x] more than just latexmk regex
 - Comments on current implementation


src/components/builder.ts
```diff
@@ -56,6 +58,12 @@ export class Builder {
         this.extension.logger.displayStatus('sync~spin', 'statusBar.foreground')
         this.preprocess(rootFile)
 
+        this.pageTotal = undefined
+        // @ts-ignore
+        pdfjsLib.getDocument(rootFile.replace(/\.tex$/, '.pdf')).promise.then(doc => {
```
zoehneto: This needs to be 

```
pdfjsLib.getDocument(this.extension.manager.tex2pdf(rootFile, true)).promise.then(doc => {
```

 in order to work with `outDir` specified.

src/components/logger.ts
```diff
@@ -35,6 +39,65 @@ export class Logger {
...
+          } else {
+            this.status2.text = 'Preamble'
+            this.status2.tooltip = ''
+            return
+          }
+        }
+
+        const generateProgressBar = (proportion: number, length: number) => {
+          const wholeCharacters = Math.trunc(length * proportion)
+          const extraEights = Math.round(
```
zoehneto: This should be `extraEighths` (an extra `h` before the `s`). Also applies to the other variables.

zoehneto:
With the `outDir` fix mentioned above this works quite nicely when using latexmk on a 90 page multi file project.

zoehneto:
Have you considered using [`vscode.window.withProgress`](https://code.visualstudio.com/updates/v1_12#_extension-authoring) for the progress bar? It might result in a somewhat cleaner result, since the current progress bar changes its width when updating progress (maybe because it doesn't use a monospaced font).

tecosaur:
### Re: proposed code commit review changes

Thanks for pointing this things out! I'll make those changes when I get back to my computer latter :smiley:

### Re: vscode.window.withProgress

I did consider it, however felt like this feature belonged better in the bar at the bottom, to be it seems like a better 'format' to have it in (even if the progress bar is rubbish). Particularly for large documents compilation doesn't happen quickly so I wanted something less in the way.

What I'm planning on going to make the progress bar less jumpy is to add fractional width spaces ask required to the 'eighths' as the blocks are all the same size. Hopefully this will make it much less noticeable.

zoehneto:
There is a `Window` option in the `ProgressOptions` for [`vscode.window.withProgress`](https://code.visualstudio.com/api/references/vscode-api#window.withProgress), does that not appear in the bar at the bottom (just asking because I couldn't find a demo gif)?

tecosaur:
I found that appears as a modal.

zoehneto:
Ok, that is obviously not useful.

src/components/builder.ts
tecosaur: Applied in 0a24c2c32033d2920b37254642988e5004bf213d

src/components/logger.ts
tecosaur: Applied in 0a24c2c32033d2920b37254642988e5004bf213d

tecosaur:
Note: Last PR added run number indicated by `⑴`, `⑵` etc. However **(Bug!)** the _first time_ that a pdf is built, rule number is not imidiately incremented to 1 (it should be).

zoehneto:
I really like the run indicators! One issue on my system though is, that ⑴ and ⑵ are much smaller than ⑶, which looks weird and also makes it hard to read. Maybe you could use [Circled Digit](http://xahlee.info/comp/unicode_circled_numbers.html) where at least the first ten numbers are consistent (on whatever font I'm using).

tecosaur:
@zoehneto I think I'll add that (and other variants) as a setting.

My current (major) 'next step' is I'd like to make calculate an ETA if at all possible to do semi-accurately. I'm thinking for the first run just extrapolate from the average page time (excluding preamble) and for subsequent runs compare the ratio of page times between runs and incorporate the times for the current run in making a prediction. @jlelong given you seem to have a decent understanding of most things LaTeX, and a very strong applied maths background :stuck_out_tongue:  — would you have any thoughts?

tecosaur:
### Update
What I just added can be summarised by this image:
image

tecosaur:
### Update \# 2

Latest hurdle that I would benefit from help in trying to overcome: How to create a webview using a local html file as a source, not just one long string (With the line numbers going well into the triple digits it's getting a bit ridiculous).

I have seen some stuff regarding getting a uri from `vscode.Uri(context.get...` however:
 1. I can't see the extension context being provided, which seems to be required in that method
 2. I found what was done in `viewer.ts` however I couldn't see anything that could be easily transfered

@jlelong @James-Yu would either of you be willing to advise?

tecosaur:
### Update \# 3

image

This will be a refined a bit at a indeterminate later date

### Bugs!

From looking at the info in this new webview I noticed that:
 - The preamble times seem wrong
 - The second run is missng a page! (the output ***is*** 18 pages)

image

Any help in diagnosing or fixing bugs appriciated

James-Yu:
Whooa! A significant progress on this amazing feature!

Sorry that I can't really contribute in these days since it is currently during the Chinese lunar new year. Will return to keyboard after it.

Happy Chinese new year!

tecosaur:
Thanks @James-Yu I look forward to when you're back behind a keyboard :smiley: 
In the meantime:

### Update \# 4
A prettier graph, with hover info

screenshot from 2019-02-04 01-45-22

package.json
```diff
@@ -1005,6 +1018,41 @@
...
+            "⒈ ⒉ ⒊ …"
+          ],
+          "markdownDescription": "The style of number to use to inidcate the run number"
+        },
+        "latex-workshop.progress.barLength": {
+          "type": "integer",
+          "default": 15,
+          "markdownDescription": "How many characters long the progress bar should be"
+        },
+        "latex-workshop.progress.barStyle": {
```
zoehneto: The `barStyle` is missing a default value. Without that there is no progress information for users who don't set the bar style. This is especially problematic since the vscode settings ui pretends the first value (`Block Width`) is set but when you open the json settings editor you see there is actually no value specified (should be improved / fixed by Microsoft/vscode#58724).

tecosaur:
I'm trying to find a line that indicates that the preamble has finished.

The best candidate so far is:
```
(./demoFile.aux)
\openout1 = `"demoFile.aux"'.
```

Does anybody have a better idea, or know if this is a good / bad one?

package.json
tecosaur: Good catch! Fixed.
Also, any thoughts on https://github.com/James-Yu/LaTeX-Workshop/pull/1157#issuecomment-460202074 ?

zoehneto: I'm not quite sure what you are suggesting in that comment, but I'm also not sure the values for the preamble are wrong. This is how my output looks when making a change in the middle of the document (notice that the time for the preamble is only short in the first run and that every run changes different pages):
image

I think it may well be that that is just how latex works / the nature of the output produced by pdflatex.

tecosaur:
@zoehneto the issue lies in a mistake I made in how I timed the preamble. I accidently made it so the preamble is marked as complete when pdflatex starts, not when the preamble ends! Hence my current search for a stdout line I can use to work out when the preamble is actually complete.

Also I notice there are a large number of missing pages, would you mind sharing your `.log` file so I can try to see if there are any deficiencies in my regex?

zoehneto:
Regarding the graph: it looks really cool but I only discovered that there is more information (run time, run name, page number) by looking at your screenshot. Even if one moves the mouse across the graph it isn't necessarily obvious, since the information only appears if you hit the knee of the graph not if you hit the edge. I would suggest adding permanent axis labels for page number and time + a legend for the run colors.

tecosaur:
Regarding the graph, on my local version I fixed the page no. axis label to be permanent. And the only reason I haven't added 'time' yet is because it takes more than one line (need to rotate canvas, then rotate back)

zoehneto:
This is my log for that run: [link](https://pastebin.com/PmXWibJY).

Not having looked deeper into your implementation one source of error would probably be that you only include latex runs. If you look at my log you can see that it contains two biber runs which are currently not accounted for by the progress bar (though I can't say how long they actually take).

tecosaur:
@zoehneto If you'd be willing to try compiling with my latest commit and showing me what you get I'd greatly appreciate it!

zoehneto:
So this is what it looks like now:
image
image

Considering how homogeneous the graph is between compilations I think the time calculation should be correct now. I'm still missing permanent axis tick labels and a legend though (though considering how much overlap there is between the lines maybe we don't need a legend). Other than that a really nice result, thanks for the great work.

zoehneto:
One case not handled yet (which might might happen for people who don't build on save but only periodically) is if many pages are added. I tested this by building the pdf with half my chapters not included and then building again with all chapters. Once you have more pages than in the last build the progress bar just keeps on growing (to double the normal size in my case).

tecosaur:
I did write this to account for that
```javascript
                    const width =
                    Math.max(
                        ...Object.values(this.pageTimes).map(
                        pt => Object.values(pt).length
                        ),
                        this.pageTotal ? this.pageTotal : 0
                    ) - 1;
```
but it sounds like it's not working as intended. I'll put it on my todo list.

tecosaur:
@zoehneto let me know how you find it now.

In particular, I'm interested in:
 - Graph resizing
 - Getting all the pages

tecosaur:
@James-Yu @jlelong If either of you have an spare time, I'd be interested on your thoughts on this section in the first post:

> ## Areas where Help Wanted
> 
> **Ordered by priority**
> 
> 1. Testing that all pages come up
> 2. Identifying regex to determine when preamble finished
> 3. How to move webview source into seperate `.html` file
> 4. Function to calculate ETA

zoehneto:
It's getting better and better. 

Regarding the graph: I'd still prefer to have labels under the x ticks. Also I think the graph should grow in width to use the space that is available. Now that we have a time for most pages it is quite hard to hit the right page on the small graph (at least for my > 100 page document).

I still don't get all pages in the following case (missing pages: 26, 27, 32, 33, 34), the path in the square brackets seems to break the regex:
```
[25] [26 <./content/pictures/related-work/something.png (PNG copy)>] [27 <./content/pictures/related-work/something-else.png (PNG copy)>] [28]) [29] (./content/chapters/5_data_sources.tex [30]
chapter 5.
[31] [32 <./content/pictures/basics/something.png (PNG copy)>] [33 <./content/pictures/applications/something.png (PNG copy)>] [34 <./content/pictures/applications/something.png>] [35] (/usr/share/texmf-dist/tex/latex/bera/t1fvm.fd) [36] [37]) [38] [39] (./content/chapters/6_anomaly_detection.tex [40]
```

zoehneto:
This solves the regex problem partially but it seems there is also a more complicated case involving newlines:
```
[48 <./content/pictures/text-parameter-analysis/Klassifizierung.pdf

pdfTeX warning: pdflatex (file ./content/pictures/text-parameter-analysis/Klassifizierung.pdf): PDF inclusion: invalid other resource which is no dict (key 'ProcSets', type <array>); ignored.
>]
```

tecosaur:
Hmmm. Here's some regex that will almost certainly catch everything
```javascript
/\[(\d+)[^\[\]]*\]/
```
The reason why I didn't have that to start is due to trying to be as specific as possible. False positives would give a non-int page number, which wouldcause an error (currently) missing pages.

I think I'll change the regex to that and add `isNaN`.

tecosaur:
@zoehneto  Done and Done.

The only thing I haven't from your comment is numbering the tick lines. This is a case of personal preference. I think think the graph is better when it is visually cleaner, and think that adding numbering would disrupt that. However, I can see that some people would prefer numbers. This may be another setting to add with the values:
 - no ticks, no numbers
 - ticks only
 - numbers only
 - ticks and numbers

zoehneto:
Great, you are probably right about the graph labels. The page parsing is almost perfect now, I'm only missing about two pages per run. In the following example I'm missing page 72 which makes me wonder if the problem is with the way the pages are handled and not the regex here:

```
[72] [73] [74]) [75] [76 <./content/pictures/prospect/capture-mysql-tpcc-partial.png>]
```

tecosaur:
That's very odd, and I'm not sure I can do anything about that. The regex is run every time new data is pushed to the log string, and as such the regex ends with `$`. If page 72 isn't being matched I would imagine it's because the line added to the log includes the space after or something that messes up the regex. I'm definitely not removing the `$` because I don't want every page number to be matched at once :stuck_out_tongue:.

If you would consider changing the regex on line 162 to `/\[(\d+)[^\[\]]*\] ?$/` and let me know if that changes anything that would be great.

zoehneto:
That regex does seem to solve the problem, I now see all pages.

Another oddity I noticed, which may or may not be a bug, is rule startup time after running something other than pdflatex. When I have a build in the following form (with latexmk): `pdflatex -> biber -> pdflatex -> pdflatex` the first and last pdflatex invocation have a startup time of 100 - 200 ms while the pdflatex after biber has a startup time of 1200 ms (10x bigger than the other cases). It seems to me that the execution time for tasks not outputting page numbers (like biber) are probably incorrect.

tecosaur:
Regarding what you've noticed with startup time being incorrect. The issue here is determining when the last running _thing_ has finished.

This comes under the _'    * more than just latexmk regex'_ point in the original post. Atm, I'm not sure what to do. I'll probably just poke around one or two log files and see if I can identify individual services but that definitely isn't robust.

I'll get back to you.

James-Yu:
@tecosaur pls notify me when you think the PR is fine for a release, in which I suppose a preliminary version is already sufficient. I'd like to bump the major version to 6 on emphasizing this feature and the synctex one just merged.

tecosaur:
Sure. I'll go over the code and test it once or twice more. I'm just going to remove the _ETA_ bit until someone comes up with a way of predicting the number of pdfLaTeX runs. 

I'd like to
 - find regex expressions to mark when the preamble has finished, and when other rules start/finish
 - externalise the `.html` and at this point, I think the class should be in its own file

However, the end user will only notice one of those and it would only be a minor tweak.

In other words, it's basically ready. One thing I would appreciate though would be if you'd be able to test it with a few of your documents to make sure that everything works as intended. I've tested it with mine but they are all quite similar in format etc.

I'll remove the ETA bit soon, and once it's been tested a bit more I'd say it's ready for release.

zoehneto:
@tecosaur have you tested this with recipes other than the latexmk one (for example the default `pdflatex ➞ bibtex ➞ pdflatex ×2`)?

tecosaur:
@zoehneto not yet I'm afraid :(

I'm ashamed to admit that I have yet to use bibtex :scream:

@James-Yu I'd say it's now ready for premliminary release, pending further testing results.

zoehneto:
It would be very helpful if someone could test a non latexmk workflow then. I would assume that behaves differently from latexmk, since latexmk is just one tool invocation with unified output and a regex to determine rule start, while multi step recipies would have one output per step and we should use the step invocation to mark the start (and to time it).

tecosaur:
For something completely different I can try `pdflatex ➞ ⬡ sage ➞ pdflatex` and see what happens.

tecosaur:
#### Update

Found issues when using not-latexmk. Fixing, ETA next few hours (got other stuff to do too).
This doesn't add support for bibtex, just makes it so that pdflatex comes up. I may add sage too.

tecosaur:
Looks like a few hours = tomorrow

tecosaur:
@zoehneto That should improve it. This is what you can expect now.

image

However other specific rules (e.g. bibtex) would be better added by someone running them. I'll try adding a little token inline bibliography and adding a basic label, however it is possible to log better info (through the revamped `BuildInfo.displayProgress(text)` which now accepts string as well as integers that represent page numbers).

tecosaur:
@James-Yu I think this should be releasable now.

zoehneto:
I think you introduced a bug with page numbering:
image

tecosaur:
Gah! Why'd I jinx it?

Thanks @zoehneto
Sorry for the false alarmn @James-Yu

tecosaur:
@zoehneto I think I know what caused that. Previously if the same page number was reported twice in the same rule it would overwrite the previous entry (obscuring anything funny that happens).

I've changed this behavious so that all reported pages are shown. My document didn't have anything strange happen so I thought it was all fine, but obviously not. What else do you have logged, and is there anything else notable that happening?

zoehneto:
It seems that pdflatex first logs warning about the first couple of pages (overfull hbox etc.) and then starts with the actual work. I've uploaded my log [here](https://pastebin.com/55YgTYuf).

tecosaur:
Thanks that's pretty helpful. Would you have any suggestions for dealing with it (I'll take a look soon-ish).

zoehneto:
I guess the only proper way of dealing with it is to add the duration of the second appearance of a page to the first appearance, that way the progress bar doesn't jump back and we get a total time per page which should be sufficient for most users.

tecosaur:
Just looked at your log, I can't see any clear regex way so that looks like the best bet. I'll implement it and get back to you.

tecosaur:
@zoehneto Give that a spin a let me know how you find it :smiley:

zoehneto:
It still has issues:
image

tecosaur:
Oh well. I'll take another look soon ish.

tecosaur:
@zoehneto It looks like I forgot to get the pages of the current rule for `pageMatchArray`, so I _was_ checking for duplicates by seeing if the page number of the current page = the current rule name... 
I'm not surprised that didn't work. Let me know if that fixes it.

zoehneto:
@tecosaur I'm still getting the same issue:
image

tecosaur:
@zoehneto I see. Unfortunately, I'm unable to replicate this issue with any of my documents.

zoehneto:
@tecosaur I'll take a look at the code and try to figure out what causes the issue.

src/components/buildinfo.ts
```diff
@@ -0,0 +1,868 @@
...
+
+        this.configuration = vscode.workspace.getConfiguration('latex-workshop')
+
+        if (current === 0) {
+            this.currentBuild.stepTimes[`${this.currentBuild.ruleNumber}-${this.currentBuild.ruleName}`][`T${+new Date()}-Wait Time`] =
+                +new Date() - this.currentBuild.lastStepTime
+        } else {
+            if (this.currentBuild.ruleProducesPages) {
+                // if page already exists, add times instead of making new entry
+                const pageAlreadyExistsRegex = new RegExp(`^T\d+-PAGE:${current}`)
```
zoehneto: The regex needs to be `^T\\d+-PAGE:${current}` (the backslash needs to be escaped), otherwise it never matches.

zoehneto:
With my regex fix I get the following result (note the ordering issue (probably due to ordering by time) and the rule timings in rule 3):
image

zoehneto:
Also there is an `undefined` entry at the bottom:
image

src/components/buildinfo.ts
```diff
@@ -0,0 +1,868 @@
...
+                const pageMatchArray = Object.keys(this.currentBuild.stepTimes[`${this.currentBuild.ruleNumber}-${this.currentBuild.ruleName}`]).map(pageLabel => Boolean(pageLabel.match(pageAlreadyExistsRegex)));
+                if (pageMatchArray.indexOf(true) !== -1) {
+                    this.currentBuild.stepTimes[`${this.currentBuild.ruleNumber}-${this.currentBuild.ruleName}`][Object.keys(this.currentBuild.stepTimes)[pageMatchArray.indexOf(true)]] += +new Date() - this.currentBuild.lastStepTime
```
zoehneto: `Object.keys(this.currentBuild.stepTimes)[pageMatchArray.indexOf(true)]` needs to be ``Object.keys(this.currentBuild.stepTimes[`${this.currentBuild.ruleNumber}-${this.currentBuild.ruleName}`])[pageMatchArray.indexOf(true)]``

This fixes both the undefined entry and the entry for previous rules.

zoehneto:
With my last two code comments the merging works as intended. The only issue remaining then is ordering of the page numbers, though that is not critical and it could be argued that we might want to keep the order in which the pages appeared in the compiler output.

tecosaur:
Thanks for taking the time to fix that. It should be easy to change it so that the order is by most recently updated. I'll push your changes + that modification and I think we should be good for an initial release! :smiley:

src/components/buildinfo.ts
tecosaur: Done

src/components/buildinfo.ts
tecosaur: Done

tecosaur:
@zoehneto That should do it

zoehneto:
I can confirm everything works as intended now :)

Before merging you should probably delete the `console.log()` statements and the commented out code and this PR should be ready to merge.

tecosaur:
Great. Thanks for reminding me about the comments and console logging.

tecosaur:
@zoehneto Done but I kept what I consider to be helpful comments, and have two TODOs.

@James-Yu Ready now.

PierreMarchand20:
It seems that it does not really work for me: 
image

here's my recipe: 
```
    {
      "name": "latexmk",
      "command": "latexmk",
      "args": [
        "-synctex=1",
        "-interaction=nonstopmode",
        "-file-line-error",
        "-lualatex",
        "%DOC%"
      ]
    }
```

I can submit an issue if you prefer

tecosaur:
@PierreMarchand20 I'm quite confident I know what the issue is.
While I developed this feature to work generally, the only workflows I could easily test were what I have used: latexmk + pdflatex.

I notice you are using luatex. As this is not one of the rules that I tested I suspect it is not working because it produces a different page output format, and/or is not recognised as producing page outputs.

Since I am aware of this limitation, I have made it so it should be quite easy to add other rules.

https://github.com/James-Yu/LaTeX-Workshop/blob/f47250cc66a9bb5bf067310c2fa3b3f813a5c299/src/components/buildinfo.ts#L84-L93

You may just need to add `"luatex"` or whatever it comes with from latexmk to line 88, a regex that matches lulatex's initialisation line as a new var, and then copy+paste+modify the following

https://github.com/James-Yu/LaTeX-Workshop/blob/f47250cc66a9bb5bf067310c2fa3b3f813a5c299/src/components/buildinfo.ts#L114-L120

(You probably just need to change the rule name on line 115 and the regex reference)

As long a luatex uses a similar page output format that should do it.

tecosaur:
@PierreMarchand20 if you have difficulties with that then if you could share your log file I'll see what I can do, but if you can do it your end you should be able to have it both working faster, and I'd be more confident that it works.

PierreMarchand20:
Ok, I will try as soon as I can !

PierreMarchand20:
I have a little bit of time to look at it this week, but I have never tried to modify a vscode package. Where are these files to modify ?

ChristianKurz:
It is shown on top of the code snippets tecosaur posted:

`LaTeX-Workshop/src/components/buildinfo.ts` and
`LaTeX-Workshop/src/components/buildinfo.ts`

Just clone this repository, make your changes to the files above and make a pull request.

I am also using a luatex & biber toolchain and would love to see this improvement. Reach out to me if you need any help.

PierreMarchand20:
Ok, but my main problem is how can I use it then in VScode ? And do I have to uninstall my current version of the package ? (How can I send you a private message ?)

ChristianKurz:
I am not sure either. I think you could follow the following tutorial:
[Building, Debugging and Sideloading the extension in Visual Studio Code](https://github.com/Microsoft/vscode-go/wiki/Building,-Debugging-and-Sideloading-the-extension-in-Visual-Studio-Code)

Which is meant for another extension. Obviously, exchange the extension name where appropriate.

PierreMarchand20:
Ok, I did the changes @tecosaur mentioned
```
        const hardcodedRulesPageProducing = ['pdflatex', 'pdftex', 'lualatex']
        const hardcodedRulesOther = ['sage']

        const rulePdfLatexStart = /This is pdfTeX, Version [\d\.\-]+[^\n]*$/
        const ruleSageStart = /Processing Sage code for [\w\.\- \"]+\.\.\.$/
        const ruleBibtexStart = /This is BibTeX[\w\.\- \"\,\(\)]+$/
        const ruleLuaTexStart = /This is LuaTeX[\w\.\- \"\,\(\)]+$/
```

and 
```
        } else if (this.currentBuild.stdout.match(ruleLuaTexStart)) {
            this.currentBuild.ruleName = 'LuaTex'
            this.currentBuild.ruleProducesPages = false
            this.currentBuild.stepTimes[`${++this.currentBuild.ruleNumber}-${this.currentBuild.ruleName}`] = {}
            this.displayProgress(0)
            this.currentBuild.lastStepTime = +new Date()
        }
```

and it seems to recognise LuaTex: 
image
but no plot of compilation time and page number

ChristianKurz:
I think you got the regex wrong. Also by looking at the Luatex output, it does not seem to output page numbers. If thats the case, use
```typescript
this.currentBuild.ruleProducesPages = false
``` 

by the way, you can get code-snippets with the correct highlighting by appending the language name to the code fence like so:
```
```typescript
        } else if (this.currentBuild.stdout.match(ruleLuaTexStart)) {
            this.currentBuild.ruleName = 'LuaTex'
            this.currentBuild.ruleProducesPages = false
            this.currentBuild.stepTimes[`${++this.currentBuild.ruleNumber}-${this.currentBuild.ruleName}`] = {}
            this.displayProgress(0)
            this.currentBuild.lastStepTime = +new Date()
        }
```
```

**Comment by PierreMarchand20** [comment]
Actually the regex is not optimal, we should take something similar to 'pdflatex', so something like
```typescript
const ruleLuaTexStart = /This is LuaTeX, Version [\d\.]+[^\n]*$/
```
but it does not change the output since luatex is recognised anyway as you said

PierreMarchand20:
Well you edited your first message, but you were right :-) it works with 
```typescript
this.currentBuild.ruleProducesPages = true
```
I will do the pull request then.

PierreMarchand20:
Actually, using latexmk, there is an issue with the plot:
Capture d’écran 2019-04-03 à 15 35 54
See the strange blue points ?

PierreMarchand20:
For some reason, the pages are not built one after the other
image
may be it is something specific to lualatex, I do not know

Issue #1159, opened 2019-01-31
Improve SyncTex accuracy


zoehneto:
**Is your feature request related to a problem? Please describe.**
SyncTex only provides line level accuracy (because column information is not stored by Tex). This makes revising documents by looking at the pdf and the synctexing to the source file very cumbersome, as you always have to search where in the highlighted line you wanted to edit.

**Describe the solution you'd like**
Other latex editors work around this issue by combining the cursor position (which can be used to get the correct line using SyncTex) with the surrounding text to match an exact location. See [this](https://tex.stackexchange.com/a/59822) Tex Stackexchange answer for a more detailed description.

tamuratak:
Keeping each line short, less than 100 characters, in a tex file solves this issue. So, I am not interested. 

It is possible to get the text contents of a pdf file with pdf.js through `<div class="textLayer">` although the contents are wholly split with lots of div elements, which makes implementation difficult.

zoehneto:
@tamuratak keeping lines shorter than 100 characters reduces the issue, but then you are wasting a lot of screen real estate (I'm currently using a limit of 150 characters).

I might look into implementing this myself when I have time, just wanted to post this since I came across the issue.

PR #1161, opened 2019-02-01
Improve accuracy for reverse synctex by checking text selection

zoehneto:
SyncTex only supports line level syncing (not column level). By also taking the text around the selection point into account we can greatly improve accuracy (see #1159).

This PR only implements the strategy for reverse synctex, because synctex provides a clear mapping to a line here. 

Forward synctex is not modified, since with forward synctex we would first have to try to find out which text corresponds with the line in the editor. Considering both repetition in the text and limits on the before and after text (if it gets too long we risk having latex constructs on both sides and being unable to find a match), there is a high risk of reducing the accuracy of forward synctex without implementing a complex heuristics based approach with fuzzy matching.

### DIFF
--- a/src/components/locator.ts
+++ b/src/components/locator.ts
@@ -295,11 +295,11 @@ export class Locator {
                 }
             }
 
-            const row = record.line - 1
+            let row = record.line - 1
             let col = record.column < 0 ? 0 : record.column
             // columns are typically not supplied by SyncTex, this could change in the future for some engines though
             if (col === 0) {
-                col = this.getColumnBySurroundingText(doc.lineAt(row).text, data.textBeforeSelection, data.textAfterSelection)
+                [row, col] = this.getRowAndColumn(doc, row, data.textBeforeSelection, data.textAfterSelection)
             }
             const pos = new vscode.Position(row, col)
 
@@ -309,6 +309,29 @@ export class Locator {
                 this.animateToNotify(editor, pos)
             })
         })
+    }
+
+    private getRowAndColumn(doc: vscode.TextDocument, row: number, textBeforeSelectionFull: string, textAfterSelectionFull: string) {
+        let tempCol = this.getColumnBySurroundingText(doc.lineAt(row).text, textBeforeSelectionFull, textAfterSelectionFull)
+        if (tempCol !== null) {
+            return [row, tempCol]
+        }
+
+        if (row - 1 >= 0) {
+            tempCol = this.getColumnBySurroundingText(doc.lineAt(row - 1).text, textBeforeSelectionFull, textAfterSelectionFull)
+            if (tempCol !== null) {
+                return [row - 1, tempCol]
+            }
+        }
+
+        if (row + 1 < doc.lineCount) {
+            tempCol = this.getColumnBySurroundingText(doc.lineAt(row + 1).text, textBeforeSelectionFull, textAfterSelectionFull)
+            if (tempCol !== null) {
+                return [row + 1, tempCol]
+            }
+        }
+
+        return [row, 0]
     }
 
     private getColumnBySurroundingText(line: string, textBeforeSelectionFull: string, textAfterSelectionFull: string) {
@@ -349,11 +372,11 @@ export class Locator {
                     return previousColumnMatches[a] > previousColumnMatches[b] ? a : b
                 }))
             } else {
-                return 0
+                return null
             }
         }
         // Should never be reached
-        return 0
+        return null
     }
 
     private indexes(source: string, find: string) {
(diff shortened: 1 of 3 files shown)


zoehneto:
@James-Yu @jlelong any feedback on this PR?

jlelong:
@zoehneto Unfortunately I do not have much to work on the extension so I mainly focus on issues.
Anyway, I have tested your PR and it does really improve `pdf -> tex` accuracy.

In some cases, it may fail. For instance, consider a very long line containing 
```
continuous time process, see 
```
and later
```
continuous time procedures, see
```
Now, in the `pdf` click, on `pro` in `procedures`, then you are redirected to `continuous time process` instead of `continuous time procedures`.

src/components/locator.ts
```diff
@@ -296,6 +294,20 @@ export class Locator {
...
+            const row = record.line - 1
+            let col = record.column < 0 ? 0 : record.column
+            // columns are typically not supplied by SyncTex, this could change in the future for some engines though
+            if (col === 0) {
+                const line = doc.lineAt(row)
+                if (data.textBeforeSelection !== '' && line.text.indexOf(data.textBeforeSelection) > -1) {
+                    col = line.text.indexOf(data.textBeforeSelection) + data.textBeforeSelection.length
+                } else if (data.textAfterSelection !== '' && line.text.indexOf(data.textAfterSelection) > -1) {
+                    col = line.text.indexOf(data.textAfterSelection)
+                }
```
jlelong: The failure example I mention in my post is mostly related to the use of `else if`.  If the `textBeforeSelection` is matched, then we do not use the `textAfterSelection`. Maybe, we could try to use both to further improve accuracy. I can make more tests if needed.

zoehneto:
@jlelong thanks for testing this. I have changed the algorithm so it now finds all columns with text before or after them which matches the text around the selection in the browser. It then returns the column with the most matches (2 if text before and after match or 1 if only text before or after match).

jlelong:
It works better like and this is a very handy feature!
Unfortunately, I have across two strange behaviours.
- Clicking on the first word of line always gets you to the next line in the `.tex` file.
- Clicking on different places of the last word of a line brings to different `.tex` locations when this word occurs several time.

I will keep doing tests.

zoehneto:
The first issue is caused by synctex incorrectly identifying the line in the source document (since this PR only searches for the correct cursor position on the line provided by synctex). My assumption in this PR is, that the line provided by synctex is correct, otherwise we would require a much more complex solution.

For the second case the solution is probably to take all the text around the cursor in the browser and then check for possible cursor locations in the input line with increasingly longer parts of text until we find only one location. I will look into that.

Another issue I find quite annoying and I'm not quite sure is fixable with vscode apis is that the vscode window is not focused when I invoke synctex from a browser instead of a vscode tab. If anyone knows how to achieve that I'd be very grateful.

zoehneto:
@jlelong the last commit should fix issue two.

zoehneto:
After using this PR for a bit I found some instances where synctex was only wrong by one line. The last commit improves these cases by checking the line before and after the line given by synctex. This partially addresses  @jlelong s issue one but doesn't help if synctex is off by many lines (commonly around the first and last lines of the page and around headlines).

James-Yu:
Just tested the PR and it works quite well in most cases. I'd merge it and try to release with the compilation process PR in the following days. They will make a new major version 6.

PierreMarchand20:
It works really great ! But it kind of breaks in some environnements or with mathematical expressions. For example 

```

\documentclass{article}
\usepackage{amsfonts,amssymb,amsthm,amsmath}
\newtheorem{theorem}{Theorem}

\begin{document}
 
I like \(\widetilde{H}^{1/2}(\Gamma)\) but I cannot find it. 

\begin{theorem}
    Let \(\Gamma \subsetneq \partial \Omega\) with \(\Omega \subset \mathbb{R}^d\) and suppose that there exists a continuous extension operator \(E:H^{1/2}(\Gamma_p)\rightarrow H^{1/2}(\Gamma)\).
\end{theorem}  
\end{document}
```

You can click on "Let", "with" for example or the first line outside the theorem environnement.

zoehneto:
@PierreMarchand20 what you are describing is caused by two different issues:
- Synctex doesn't work very well in environments (the line provided by synctex is often off by more than 5 lines).
- Matching text on either side of the cursor in the presence of commands, environments, special characters etc. In your example we would have to map from the omega symbol to `\Omega` which we realistically can not do. The only solution I see is to try and retrieve only plain text using a regex and then doing some sort of fuzzy matching to determine the best result.

The current implementation works well enough for me so I probably won't be doing any additional work on this in the near future but I'd love to see a PR for it.

PR #1180, opened 2019-02-19
Use mutex instead of ChildProcess.kill

### LINKED ISSUE #1166: Sometimes, the build action lasting very, very long time and it can not end. When I restart the VScode, it can work well.
It seems that after I modified the tex file and build the tex file several times, the problem will occur. There is no error to report. And the state of build represent it is building now, but it can not end successfully. After I restart the VScode, it can build well again. 
OS: win10
VScod:1.31.0
latex-workshop:5.21.0

Now, What can I do for fixing the problem?
Besides, What can I do for stop the building task? Thanks very much in advance.

tamuratak:
This is a PR to fix #1166.

I think the cause of the issue is that `ChildProcess.kill` is used to terminate a current latex processing. Immediately after calling it, we restart a new latex processing. Because we don't wait for the killing process to finish, `Builder.currentProcess` is not appropriately set, this issue occurs. One possible scenario is the following.

1. `Builder.currentProcess` is not appropriately set.
2. LaTeX Workshop tries to kill a wrong child process as a previous LaTeX building process.
3. Multiple LaTeX build processes run parallel and hang.
4. We cannot kill one of them.

Except for the above race condition, using kill might be problematic.

1. Signals are not supported on Windows. Node.js and libuv emulate them on Windows. So we should avoid using signals on Windows as possible as.
2. Killing latex processing leaves auxiliary files inconsistent.

In this PR, we use mutex to avoid such a race condition.

To avoid a build error related to an unused local variable, I include the npm module await-semaphore in this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1160,6 +1160,7 @@
     "workspaceSymbolProvider": "true"
   },
   "dependencies": {
+    "await-semaphore": "^0.1.3",
     "chokidar": "^2.0.3",
     "fs-extra": "^5.0.0",
     "glob": "^7.1.1",
--- a/package.json
+++ b/package.json
@@ -1160,7 +1160,6 @@
     "workspaceSymbolProvider": "true"
   },
   "dependencies": {
-    "await-semaphore": "^0.1.3",
     "chokidar": "^2.0.3",
     "fs-extra": "^5.0.0",
     "glob": "^7.1.1",
--- a/src/commander.ts
+++ b/src/commander.ts
@@ -38,7 +38,7 @@ export class Commander {
         const externalBuildCommand = configuration.get('latex.external.build.command') as ExternalCommand
         if (externalBuildCommand.command) {
             const pwd  = path.dirname(vscode.window.activeTextEditor.document.fileName)
-            this.extension.builder.buildWithExternalCommand(externalBuildCommand, pwd)
+            await this.extension.builder.buildWithExternalCommand(externalBuildCommand, pwd)
             return
         }
         const rootFile = await this.extension.manager.findRoot()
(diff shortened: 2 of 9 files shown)


James-Yu:
Very interesting. One quick question, why don't include the semaphore package as a dependency? I'm not familiar with the package though.

tamuratak:
The `await-semaphore` package does not include a compiled js file, only ts files. So when we include the package as a dependency, tsc tries to compile them, which occurs a build error related to an unused local variable. 

Changing `tsconfig.json` might solve the build issue. But I cannot find an appropriate setting to make the build work fine.

tamuratak:
Another possible scenario is that

1. With a recipe having multiple steps, there is a time in which `Builder.currentProcess` points to the process of a previous step on each step.
2. If `kill` is called during the time, this issue occurs.

tamuratak:
Yet another (the most) possible scenario is that

1. Two short sequence of `ctrl+s` may fire concurrent bulid events before `Builder.disableBuildAfterSave` is set `true`.
2. This issue occurs.

James-Yu:
The previous implementation of killing process does not take aaynchronous execution in account, which caused a lot of troubles. May I know which scenarios this PR can handle if not all? I will look into a way to use package dependency without triggering compiler error.

tamuratak:
I think this PR can handle all the scenarios above I have mentioned.

James-Yu:
I agree with you that there is no obvious solution to handle the package error. Therefore I would leave it as is. When you think the PR is good to be merged, please let me know.

tamuratak:
It is ready now. 👍

DRjy:
When we can update the latex-workshop to fix the problem? Thank you very much.

Issue #1185, opened 2019-02-20
Incorrect highlighting for mathematics delimiters


simon-nicholls:
The math delimiters `$` and `$$` are being matched across semantically mismatched sections of text, leading to scoping and highlighting problems.

For example an opening `$` can be matched within inline code, and the closing `$` in a nested code block:

```latex
A Haskell type definition for \(\mathbb{N}\), and testing \inlinecode{<$>}

\begin{code}
  data Nat = Zero | Succ Nat
  test = (++"...gnom!") <$> ["pizza", "ljubljanska"]
\end{code}
```

which will incorrectly set a scope of `support.class.math.block.tex` for

```latex
$>}

\begin{code}
  data Nat = Zero | Succ Nat
  test = (++"...gnom!") <$
```

I initially raised the issue on JustusAdam/language-haskell#82

I would instead expect `$` matching to only be done within same-semantic chunks of text, and beyond that I'd like to disable matching inside code environments if possible.

MacOS 10.14.3, VS Code 1.31.1, Workshop 6.0.0

simon-nicholls:
Wow, that was quick, thanks!

As per details in https://wiki.haskell.org/Literate_programming, can we also add `spec` as a verbatim code block? Where `code` will be compiled by the Haskell compiler, `spec` is a block that will be treated as Haskell code, but that shouldn't be compiled as part of the module (it's better for example code snippets, or Bird-Meertens style equational reasoning).

When it comes to typesetting, `code` and `spec` are implemented in terms of `minted` or `listings` for example, but `code` and `spec` are the repeated content tags expected for LaTeX unliteral processing in Haskell, so are great to have.

I used `\inlinecode` in my example, but there doesn't appear to be a standard for that. In fact, going by the `lstinline` you already had in your pattern match, and noting that minted uses `mintinline`, I think `codeinline` would be a more fitting name.

Thanks again. I'll look into how I can test repo versions of extensions, so I can have a play myself.

jlelong:
@simon-nicholls can you be more specific.

I can declare the `code` and `spec` as verbatim. Easy. Are these related to Haskell? I could also declare as them embedded Haskell code and it would get highlighted as so.

`\inlinecode` is already considered as verbatim

[...]

PR #1192, opened 2019-02-22
Literate haskell

simon-nicholls:
This pull request alphabetically inserts an `hscode` declaration, as discussed at https://github.com/James-Yu/LaTeX-Workshop/issues/1185, which I based on the `luacode` declaration below it.

It also renames the `inlinecode` declaration to `hsinline`, which had been added for that same issue. This makes the naming consistent with the way `minted` and `listings` work, by suffixing instead of prefixing, and is specific to Haskell. e.g. from the minted `\newmintinline` docs:

```latex
\newmintinline{perl}{showspaces}

X\perlinline/my $foo = $bar;/X
```

and it's `code` equivalent which will work with the new `hscode`:

```latex
\newminted{cpp}{gobble=2,linenos}

\begin{cppcode}
```

`code` or `inline` get added after the language name. i.e. `hscode` and `hsinline`.

### DIFF
--- a/syntax/LaTeX.plist
+++ b/syntax/LaTeX.plist
@@ -521,6 +521,34 @@
 		</dict>
 		<dict>
 			<key>begin</key>
+			<string>((?:\s*)\\begin\{(hscode(?:\*)?)\}(?:\[.*\])?)</string>
+			<key>captures</key>
+			<dict>
+				<key>1</key>
+				<dict>
+					<key>patterns</key>
+					<array>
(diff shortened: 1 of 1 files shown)


jlelong:
Considering the `minted` documentation, I have decided to declare `\\[a-z]+inline` as verbatim. See 7a12c9f.
We will see if anybody complains...

simon-nicholls:
Nice; I think it's definitely worth a go.

Issue #1153, opened 2019-01-29
Enforce parsing of all files


dbogatov:
**Is your feature request related to a problem? Please describe.**

I dynamically substitute files, so I need parser (and chktex specifically) to work with all `*.tex` files.

**Describe the solution you'd like**

Have an option to lint all files, not just the root and what's included from the root.

**Additional context**

Here is the project structure:

```
main.tex
preamble.tex
cli.tex
assignments/
    hw-1/
        content.tex
        meta.tex
    hw-2/
        content.tex
        meta.tex
    ...
```

and these lines of code in main

```latex
\input{assignments/\assignment/meta} % chktex 27
\begin{document}
    \input{assignments/\assignment/content} % chktex 27
\end{document}
```

as you may guess, value of `\assignment` is externally changed.
This way I can compile my 20 assignments with script, reusing same preamble.

Because the content is not included explicitly (eg. ` \input{assignments/hw-1/content}`), chktex does not work for it.

jlelong:
I am wondering whether this is the good way to proceed.
Maybe, it would be better to support macro replacement for expressions of the type
```
\\(def|(re)?newcommand){\\commandname}{value}
```
when computing the tree structure of the LaTeX project.

jlelong:
Parsing `\newcommand` definition may not be satisfying as we will never be able to correctly evaluate the `\if` section and recursivity may kill us. 
I am actually thinking of adding a new configuration setting to define some macros (without arguments) in the same spirit as the `defines` entry in `c_cpp_properties.json`, see [here](https://code.visualstudio.com/docs/languages/cpp#_configuring-intellisense). 

Any thoughts on that?

dbogatov:
Allowing to define macros in settings would be nice.

My concern, however, was not about parsing my own macros.
I want a setting that instructs the extension to parse all `*.tex` files, not only those which a reachable in a tree starting at main.

jlelong:
> I want a setting that instructs the extension to parse all `*.tex` files, not only those which a reachable in a tree starting at main.

Unfortunately, this is unlikely to be implemented as it does not fit in the current logic of the extension.

dbogatov:
Replying to https://github.com/James-Yu/LaTeX-Workshop/issues/731#issuecomment-468950368

For me a big win would be at least having chktex parsing all the files.

jlelong:
Closed by PR #1199.

Issue #1194, opened 2019-02-23
Intellisense and bibDirs not working in 6.0.0


nbeckers:
After updating to 6.0.0 the citation browser/Intellisense is not working anymore. My `.bib` file is in a different location as the source code, but setting `latex-workshop.latex.bibDirs` to that folder (absolute path) does not solve the problem.

The citation browser was working in 5.21.0 just fine (here I also set `latex-workshop.latex.additionalBib` to the `.bib` file (absolute path)).

Thanks!

jlelong:
I confirm that it does work. 
Post the relevant part of your `.tex` file along with the value of `latex-workshop.latex.bibDirs` and the logs of the extension.

nbeckers:
I got it working now too. 
The problem seemed to be that I didn't call `\bibliography` directly in my main `.tex` file, but through a custom command which then called `\bibliography`.

My apologies if this issue was unnecessary, but I thought it was odd that the Intellisense/citation browser worked fine in 5.21.0 with my own original `.tex`files, but not in 6.0.0 with `latex-workshop.latex.bibDirs`. My guess is that this worked because I specifically specified the `bib` file I needed in `latex-workshop.latex.additionalBib` and the new `bibDirs` functionality only selects the `.bib` file that it found in the `\bibliography` command?

Thanks!

jlelong:
> My guess is that this worked because I specifically specified the `bib` file I needed in `latex-workshop.latex.additionalBib` and the new `bibDirs` functionality only selects the `.bib` file that it found in the `\bibliography` command?

You are right. I am working on a new mechanism using the last compilation to detect files that are through non standard macros.

Issue #731, opened 2018-08-03
a setting to disable automatic parsing of .tex files in the workspace.


davydden:
### Description

it appears that LeTeX-Workshop automatically searches for all `*.tex` files in the workspace and tries to parse them. I would like to be able to disable this feature. I usually have a mix-bag workspaces (C++/Python/LaTeX) and would prefer not to spend any resources to parse random `*.tex` files I never plan to compile but they happen to be in C++/Python projects' folders.

### Log messages

```
[09:16:25] Parsing <path/to/something/.../some.tex>
```

### How to reproduce

1. add another project to workspace which happen to have `*.tex` file.
2. call `Build LaTeX` from some other `*.tex` file in a different folder.

### Expected behavior

LaTeX workshop should not parse each and every `*.tex` file it finds in the workspace.

### Actual behavior

there appears to be a background process that looks for all `*.tex` files in the workspace folders and parse them.

### Additional Information

nothing special.

James-Yu:
Actually it only parses tex files in the latex project, which is imported/input in the root file.

It does not cost much to parse files at least.

davydden:
> Actually it only parses tex files in the latex project, which is imported/input in the root file.

now, that's not the case. I open and edit a several `*.tex` files and `LaTeX-Workshop` picks up some completely unrelated `*.tex` files from elsewhere, like installation prefixes of libraries.

James-Yu:
Please provide a minimal working example.

davydden:
> Please provide a minimal working example.

see `how to reproduce`.

davydden:
> only parses tex files in the latex project

a bit off-topic, but AFAIK there is nothing like `latex project` in VSCode, you just have a workspace with folders added. At least I don't mark those folders as being C++, Python  or LaTeX projects anyhow. That's also because sometimes they contain all of those inside a single project.

solidxsnake:
Why was this closed? I would also like to disable automatic parsing. This makes VSCode unusable for the particular project I'm working on (which can easily have tens to over a hundred .tex files, which are all auto-generated from templates). Any time I modify any of the templates, VSCode starts to parse all of the .tex files in my workspace. I'd prefer it not parse them at all, or perhaps having an option to exclude/include specific directories for parsing.

gusbemacbe:
I also would like to disable automatic complying, because it is too boring while I am writing and saving many times

jlelong:
[...]

PR #1199, opened 2019-02-25
Use the last compilation to find and watch all the dependencies

jlelong:
This PR is related to #1153 and [#1194](https://github.com/James-Yu/LaTeX-Workshop/issues/1194#issuecomment-466745848).

Running the LaTeX compiler with the -recorder option (automatically added by `latexmk`) generates a `.fls` file, which contains the list of all input files (`.tex`, `.sty`,...) and output files (`.aux` in particular). Then, we can extract the tree of `.tex` files and from the `.aux` files, we get all the `.bib` files.

The results provided are of course always a bit outdated as they do not take into account the changes since the last compilation but it will undoubtedly help with files included using personal macros.

This new mechanism helps 
- intellisense for commands, references and citations
- finding the correct root file when the current file is included via personal macros or the file name itself is a macro.

I am making a call for testers before merging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -636,11 +636,10 @@
           "type": "string",
           "enum": [
             "never",
-            "onSave",
             "onFileChange"
           ],
-          "default": "onSave",
-          "markdownDescription": "When the extension shall auto build LaTeX project using the default (first) recipe.\n`onSave` builds the project upon saving a `tex` file in vscode. `onFileChange` builds the project upon detecting a `tex` file change, even modified by other applications."
+          "default": "onFileChange",
+          "markdownDescription": "When the extension shall auto build LaTeX project using the default (first) recipe.\n `onFileChange` builds the project upon detecting a file change in any of the dependencies, even modified by other applications."
         },
         "latex-workshop.latex.autoBuild.cleanAndRetry.enabled": {
           "type": "boolean",
--- a/package.json
+++ b/package.json
@@ -632,6 +632,23 @@
           "default": [],
           "markdownDescription": "Patterns of files to exclude from the root detection mechanism.\nSee also `latex-workshop.latex.search.rootFiles.include`. For more details the [wiki](https://github.com/James-Yu/LaTeX-Workshop/wiki/Multi-File-Projects)."
         },
+        "latex-workshop.latex.watch.files.ignore": {
+          "type": "array",
+          "default": [
+            "**/*.bbx",
+            "**/*.cbx",
+            "**/*.cfg",
+            "**/*.clo",
+            "**/*.cnf",
+            "**/*.def",
+            "**/*.fmt",
+            "**/*.lbx",
+            "**/*.map",
+            "**/*.pfb",
+            "**/*.tfm"
+          ],
+          "markdownDescription": "Files to ignore from the watching mechanism used for triggering autobuild.\nThis property must be an array of globs pattern. The patterns are matched against the absolute file path. To ignore everything inside the `texmf` tree, `**/texmf/**` can be used."
+        },
         "latex-workshop.latex.autoBuild.run": {
           "type": "string",
           "enum": [
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -284,6 +284,7 @@ export class Builder {
         this.extension.logger.addLogMessage(`Successfully built ${rootFile}.`)
         this.extension.logger.displayStatus('check', 'statusBar.foreground', 'Recipe succeeded.')
         this.extension.viewer.refreshExistingViewer(rootFile)
+        this.extension.manager.findAdditionalDependentFilesFromFls(rootFile)
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
         if (configuration.get('synctex.afterBuild.enabled') as boolean) {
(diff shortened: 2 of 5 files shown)


xu-cheng:
FYI, the `.fls` may contain auxiliary files created by the latex, which should not be considered as dependency. For example, `\includegraphics{./foo.eps}` will generate the following content in the `.fls`
```
INPUT ./foo.eps
INPUT ./foo-eps-converted-to.pdf
```
Here, the `-eps-converted-to.pdf` is generated by latex from `.eps` file. I am not sure whether would this cause any problem if both are watched to trigger the build.

As side note, instead of trying to reinvent the wheel (i.e. parsing both tex source and log), why not just using `latexmk -pvc` (which is used in https://github.com/lervag/vimtex, an excellent vim plugin for latex) or `latexmk -deps`

jlelong:
> As side note, instead of trying to reinvent the wheel (i.e. parsing both tex source and log), why not just using `latexmk -pvc` (which is used in https://github.com/lervag/vimtex, an excellent vim plugin for latex) or `latexmk -deps`

It assumes that everyone uses `latexmk`, which is not the case. Some people use far more elaborate compiling toolchain. The `.fls` file is generated by the `-recorder` option of `latex`, `pdflatex`, `xelatex`,... Moreover parsing the `.fls` file is really easy and gives almost all the information we need, the `.bib` files are listed in aux files.

xu-cheng:
Thanks for explanation. 

>  Moreover parsing the `.fls` file is really easy and gives almost all the information we need, the `.bib` files are listed in aux files.

From my testing, it seems this only applies to the bibtex not biblatex. Therefore, you may want to read `.bcf` in this case.

jlelong:
Thanks for the tips concerning `biblatex`. Anyway, we still scan `.tex` files for `\addbibresource`, which enables us to find any `.bib` file not included using a personal macro.

jlelong:
### List of changes implemented so far

- In addition to the standard structure parser, we read all the dependent files from the `.fls` when it exists: classes, packages, styles, graphics, included `tex` files, ... All these files are watched to trigger build when `latex.autoBuild.run` is set to `onFileChange`. Unlike the structure parser, the use of `.fls` makes the detection of dependencies robust to the use of any macros.
- We detect `.bib` from the `.aux` (this is only valid when using `bibtex`). When using `biblatex`, the standard parser was already scanning the content of `\addbibresource`. Any detected `.bib` file is watched to trigger build when `latex.autoBuild.run` is set to `onFileChange`.
- We do not watch all `**/*.tex` anymore.

These changes fix #1153. 

### Thoughts for future update

- I am considering merging the `onSave` and `onFileChange` for `latex.autoBuild.run` as I find the difference a bit confusing. Any suggestion on that?
- We need a setting to disable the entire parsing/watching mechanism (#731).

**Testers are welcome**: I need your feedback!

James-Yu:
Sorry for off the track for a long time. It looks like we don't track the tex changes by other applications any more? I would suppose changing the config back to a boolean should be adequate.

jlelong:
> It looks like we don't track the tex changes by other applications any more?
Yes, we do. Instead of watching all `**/*.tex`, we watch all the files listed as `INPUT` in the `rootFile.fls`, which basically corresponds to what `latexmk` outputs with the `-deps` option. We track the changes in any dependency: packages, classes, input `.tex` files, graphs, `.bib` files... All these dependent files are also parsed for intellisense.

The real advantage is that the dependencies are computed by the `latex` compiler itself and are therefore robust to any intricate constructions. The downside is that the dependencies are only computed at compile time and not updated in real time. To try to keep them up to date in real time, I continue using the standard manager parser. 

### Things yet to be done

- [ ] I think we should merge the `onSave` and `onFileChange` choices for `latex.autoBuild.run` and use the `onFileChange` approach, ie watch any dependent file be it modified inside vscode or not.
- [ ] Add a setting for the extensions of the files to watch.
- [ ] Deal with files that are produced by the compilation but listed as dependencies - eg, `rootFile.run.xml`, -eps-converted-to.pdf` - because they tend to launch several compilations in a row when saving a file.
- [ ] Write the documentation for all this on the wiki.

xu-cheng:
> Things yet to be done

Maybe add parsing `.bcf` file for biblatex?

jlelong:
`biblatex` is already support using the standard manager parser, which scans `.tex` files for `\addbibresource`. So I think we get away with it without parsing `.bcf` files.
See my [comment above](https://github.com/James-Yu/LaTeX-Workshop/pull/1199#issuecomment-468618024).

xu-cheng:
These are my thoughts.
* Standard parser cannot handle many other ways to include bib resources for biblatex. For example, it does not support `refsection` or any custom macro/command/environment. 
* I think we can stop to use standard parser to track build dependencies. Since it becomes unnecessary after this PR and it may contribute to better performance.
* Parsing `.bcf` file should not be too hard as it is just a xml file.

jlelong:
I cannot se the problem with the `refsection` environment. This PR is already quite complicated so I will keep the `biblatex` issue aside for now and when the PR is merged I will investigate the parsing of `.bcf`. Meanwhile, you are welcome to submit a PR yourself.

Issue #1211, opened 2019-03-04
File path in diagnostics is broken when the path contains umlauts


lrabe:
**Expected behavior:** When clicking a diagnostic entry in the "Problems" tab, the editor should jump to the indicated location in the file. 

**Actual behavior:** The editor fails to open the file if the path contains a German umlaut (e.g. ö, ü, ä). The behavior also differs between LaTeX errors and linter warnings: When clicking a LaTeX error, everything after the umlaut is removed from the path. When clicking a linter warning, the umlaut is replaced by a question mark symbol. 

**Example:**

![vscode_edit](https://user-images.githubusercontent.com/9059119/53758845-2d3a9380-3ebf-11e9-87ba-7a421a5dce10.jpg)

I suspect that this might be related to the following lines: 
https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/components/parser.ts#L281
https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/components/parser.ts#L228

- OS: Windows 10
- VS Code version: 1.31.1
- Extension version: 6.0.0

jlelong:
The truncation issue is fixed for LaTeX errors.

> When clicking a linter warning, the umlaut is replaced by a question mark symbol.

This is an encoding issue. vscode expects uft8 encoded paths whereas it is probably not your default system encoding.

lrabe:
@jlelong Thanks. Could you explain what you mean by the "default system encoding"?
I've never had an issue like this in any other program, so I'm not sure if this really is the case. 

The only related setting I could find is the following (which is unchecked by default).

![](https://i.imgur.com/WvUmZIB.png)

jlelong:
I don't use Windows so I am afraid I cannot help much but it seems that Windows uses `UTF-16` for filename encoding and not `UTF-8`. Maybe other users can help...

tamuratak:
Please show us log messages.  Tell us which TeX distribution you are using. Which LaTeX engine and linter. Exact versions of all these programs.

![2018-11-07 20 34 24](https://user-images.githubusercontent.com/10665499/48129417-c4565c80-e2cc-11e8-89de-05b36ef1fa76.png)

Thank you for your cooperation.

lrabe:
Sure, here you go: 

<details>
<summary>LaTeX Workshop Output</summary>
[14:36:58] Initializing LaTeX Workshop.
[14:36:58] Creating LaTeX Workshop http and websocket server.
[...]

PR #1230, opened 2019-03-14
convert filename encoding

tamuratak:
This is a PR to solve #1211. The encoding of filenames in the problems panel will be converted if necessary.


James-Yu:
Sorry that I did not follow the discussion and this PR. Is it good for merging?

tamuratak:
I think it is ready.

PR #1241, opened 2019-03-20
Fix external reverse synctex on Windows

### LINKED ISSUE #1240: Reverse synctex broken on Windows when not using synctexjs
**Describe the bug**
Reverse synctex doesn't work on Windows when using external synctex.

**To Reproduce**
0. Use a Windows machine
1. Disable internal synctex (`"latex-workshop.synctex.synctexjs.enabled": false`)
2. Compile Latex file
3. Open pdf viever
4. Ctrl+click somwhere
5. Nothing happens

**Expected behavior**
Corresponding line in Latex source should be highlighted in VSCode.

**Logs**
Please paste the log messages here. It is very important to identify problems. If you think the logs are unrelated, please say so.

<details>
<summary>LaTeX Workshop Output</summary>

[15:53:43] Executing synctex with args edit,-o,96:206.9180754226268:453.76853055916774:c:\Users\dasch8\latex\main.pdf
</details>

<details>
<summary>Developer Tools Console</summary>

console.ts:134 [Extension Host] rejected promise not handled within 1 second: Error: ENOENT: no such file or directory, lstat 'c:\Users\dasch8\latex\main.tex
'
t.log @ console.ts:134
console.ts:134 [Extension Host] stack trace: Error: ENOENT: no such file or directory, lstat 'c:\Users\dasch8\latex\main.tex
'
	at Object.realpathSync (fs.js:1744:7)
	at Object.exports.wrapFsWithAsar.fs.realpathSync (ELECTRON_ASAR.js:335:40)
	at module.exports.t.Locator.<anonymous> (C:\Users\dasch8\.vscode\extensions\james-yu.latex-workshop-6.2.2\out\src\main.js:371:335399)
	at Generator.next (<anonymous>)
	at s (C:\Users\dasch8\.vscode\extensions\james-yu.latex-workshop-6.2.2\out\src\main.js:371:330388)
t.log @ console.ts:134
log.ts:173   ERR ENOENT: no such file or directory, lstat 'c:\Users\dasch8\latex\main.tex
': Error: ENOENT: no such file or directory, lstat 'c:\Users\dasch8\latex\main.tex
'
	at Object.realpathSync (fs.js:1744:7)
	at Object.exports.wrapFsWithAsar.fs.realpathSync (ELECTRON_ASAR.js:335:40)
	at module.exports.t.Locator.<anonymous> (C:\Users\dasch8\.vscode\extensions\james-yu.latex-workshop-6.2.2\out\src\main.js:371:335399)
	at Generator.next (<anonymous>)
	at s (C:\Users\dasch8\.vscode\extensions\james-yu.latex-workshop-6.2.2\out\src\main.js:371:330388)
</details>

**Desktop (please complete the following information):**
 - OS: Windows 10
 - VS Code version: 1.32.3
 - Extension version: 6.2.2

**Additional context**
I tracked the bug down to here:
https://github.com/James-Yu/LaTeX-Workshop/blob/589a0eb26c088e48ac21990930f803e2b01eb962/src/components/locator.ts#L273-L281

`fs.realpathSync(record.input)` fails because there is a newline character at the end of the file path. If Docker is used for synctex it also fails because the path to the file in the container is not replaced with the path to the file in the host. I'm creating a pull request with my fix.

dasch8:
Fixes #1240, however I couldn't verify that it doesn't break anything on other platforms.

### DIFF
--- a/src/components/locator.ts
+++ b/src/components/locator.ts
@@ -269,7 +269,12 @@ export class Locator {
             }
         } else {
             record = await this.invokeSyncTeXCommandBackward(data.page, data.pos[0], data.pos[1], pdfPath)
+            record.input = record.input.replace(/(\r\n|\n|\r)/gm, '')
+            if (docker && process.platform === 'win32') {
+                record.input = path.join(path.dirname(pdfPath), record.input.replace('/data/', ''))
+            }
         }
+
         // kpathsea/SyncTeX follow symlinks.
         // see http://tex.stackexchange.com/questions/25578/why-is-synctex-in-tl-2011-so-fussy-about-filenames.
         // We compare the return of symlink with the files list in the texFileTree and try to pickup the correct one.
@@ -280,11 +285,7 @@ export class Locator {
             }
         }
 
-        let filePath = path.resolve( record.input.replace(/(\r\n|\n|\r)/gm, '') )
-        if (docker && process.platform === 'win32') {
-            filePath = path.resolve(path.dirname(pdfPath), record.input as string)
-        }
-
+        let filePath = path.resolve(record.input)
         this.extension.logger.addLogMessage(`SyncTeX to file ${filePath}`)
         vscode.workspace.openTextDocument(filePath).then((doc) => {
             let viewColumn: vscode.ViewColumn | undefined = undefined


src/components/locator.ts
```diff
@@ -280,11 +285,7 @@ export class Locator {
             }
         }
 
-        let filePath = path.resolve( record.input.replace(/(\r\n|\n|\r)/gm, '') )
-        if (docker && process.platform === 'win32') {
-            filePath = path.resolve(path.dirname(pdfPath), record.input as string)
-        }
-
```
jlelong: After moving this block inside the  `else` part above, nothing deals with docker when using `synctexjs`.  This is a bit surprising. 

It looks like two things may be mixed up here and it works when using external `synctex`.  Have you tested that docker works both with external `synctex` and `synctexjs`.

jlelong:
I have the feeling that it may be sufficient to move the line

```js
 let filePath = path.resolve( record.input.replace(/(\r\n|\n|\r)/gm, '') )
```

right before the `for` loop to fix #1240. Can you test and confirm that?

src/components/locator.ts
jlelong: I do not have Windows so just to be sure have you tested that using docker and `synctexjs` works on Windows?

dasch8: I just tested this and `synctexjs` doesn't work with Docker on Windows anyway because it includes it's own check whether the file exists before we can patch the path: The block could be moved out of the else branch though.

dasch8:
> I have the feeling that it may be sufficient to move the line
>  let filePath = path.resolve( record.input.replace(/(\r\n|\n|\r)/gm, '') )
> right before the for loop to fix #1240. Can you test and confirm that?

To fix the problem on Windows without Docker putting `record.input = record.input.replace(/(\r\n|\n|\r)/gm, '')` above the for loop would work. For it to also work with Docker we need to patch the path before we pass it to `fs.realpathSync`.

src/components/locator.ts
jlelong: For the docker and `synctexjs` issue, it seems we just need to add something like
```js
if (docker && process.platform === 'win32') {
    record.input = record.input.replace('/data/', '')
}
```
juste before https://github.com/James-Yu/LaTeX-Workshop/blob/8dd06def151053ceecfec4bd90547863885a090b/src/components/synctex.ts#L206

Unfortunately I cannot test it. Could you make a PR or do you prefer me to do it and rely on you for the testing?

jlelong:
Merged in e75d5fd.

src/components/locator.ts
dasch8: I can make a PR. I also noticed that the cleaning command doesn't work if I build with Docker. I'll see whether I find further issues and will make a PR as soon as possible.

PR #1251, opened 2019-03-25
pdf Viewer: Allow multi-page view

ChristianKurz:
I like having pdf-pages shown side by side. Using "inline-block" allows to select a zoom-level to allow for that.

Perhaps an option to display as inline-block would be better than forcing this setting.

### DIFF
--- a/viewer/viewer.css
+++ b/viewer/viewer.css
@@ -271,6 +271,7 @@
 }
 
 .pdfViewer .page {
+  display: inline-block;
   direction: ltr;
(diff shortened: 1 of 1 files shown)


zoehneto:
Isn't that behavior already provided by the horizontal scrolling option (or maybe more usefully be the spread modes) of the viewer?

ChristianKurz:
The wrapped scrolling mode indeed does provide this feature. I will close this pr.

PR #1254, opened 2019-03-26
find output files (*.axu...) in outDir instead of rootDir

Lencerf:
`result` contains *.aux, *.xdv files and the following codes try to find these files in `rootDir`:
```
const outputFilePath = path.resolve(rootDir, result[4])
```
If my understanding is correct, these files are actually located in `outDir` instead of `rootDir`, since if `-outdir=` is specified for `latexmk`, `latexmk` will output all these files to `outdir`. Therefore `rootDir` should be changed to `outDir`.

### DIFF
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -390,7 +390,7 @@ export class Manager {
                     inputFiles.add(inputFilePath)
                 }
             } else if (result[3]) {
-                const outputFilePath = path.resolve(rootDir, result[4])
+                const outputFilePath = path.resolve(outDir, result[4])
                 if (outputFilePath) {
                     outputFiles.add(outputFilePath)
                 }


jlelong:
I guess, you are right and I will your PR. Do you experience any issue related to this?

Lencerf:
Thanks! The issue I have is the extension does not automatically compile the tex file when I save the tex file. The problem comes from the personal compile script I use. To avoid generating files like *.aux in the project directory and speed up compiling, I wrote a script to symlink all the project files in a RamDisk and then run latexmk in the RamDisk path. Since latexmk is working in RamDisk, all the *.aux files are generated in the RamDisk. However, the following codes in `manager.ts` still try to find *.aux files in rootDir, which is the directory where the original tex file is located.
https://github.com/James-Yu/LaTeX-Workshop/blob/444331e558b6a0cfeb849bff817c59738febbd1f/src/components/manager.ts#L393-L395
This later causes 
https://github.com/James-Yu/LaTeX-Workshop/blob/444331e558b6a0cfeb849bff817c59738febbd1f/src/components/manager.ts#L431
to crash because `auxFilePath ` actually does not exist.

I also find a workaround to solve this issue, that is to explicitly specify `-outdir=` for latexmk, then all the OUTPUT paths in *.fls are absolute paths and `path.resolve(rootDir, result[4])` can still give the correct path of *.aux. However, I still think we should change `rootDir` to `outDir, because logically *.aux, *.log files are in the output directory, not in the root directory, although in most cases they are the same.

PR #1263, opened 2019-04-01
add accents commands

tamuratak:
This is a PR to add accents commands.


James-Yu:
Good catch! Will merge when you think ready.

tamuratak:
Yes, it is ready.

PR #1266, opened 2019-04-01
Support for ts-tex language

b4ckup:
I wrote a [language service](https://gitlab.com/b4ckup/ts-tex) that allows the user to call typescript code from latex (or any other) documents. I also just published a [vs extension](https://gitlab.com/b4ckup/vscode-ts-tex) that delivers that language service to the user. It's basically a way to write type safe macros with full ts intelli sense out of document files. The thing is I had to define my own language (tstex/ts-tex) because I could not find a way to deterministically overwrite the .tex syntax highlighting. Still, I would love to have the normal LaTeX language services available together with what vscode-ts-tex provides for the ts function calls. So in the first commit, I've added ts-tex to the languages. It would be awesome if you could support me (and my "language") on this!


James-Yu:
I'm personally conservative to add new tex-related languages to the extension. There are a lot of such ones in the wild, and I would like to include those recognized by at least a small group of people. My 2c only.

b4ckup:
I understand. I looked into solving the problem differently and it turns out I can inject my own grammar into the *text.tex* scope. Thanks anyways for the great extension!

Issue #1265, opened 2019-04-01
Preview support for code-server (an online version of vscode)


imxieyi:
**Is your feature request related to a problem? Please describe.**
This plugin can already be installed on [code-server](https://github.com/codercom/code-server). Most functions including compiling works just fine out of the box. However, the preview function is broken. It is because the preview extension is listening on localhost as shown in [viewer.ts#L126](https://github.com/James-Yu/LaTeX-Workshop/blob/0ef15bc11f6fe487a20334bf4acae17d2be465c7/src/components/viewer.ts#L126), which is impossible to be connected from a remote host. What makes things worse is that the port number is random.

**Describe the solution you'd like**
If the viewer can be implemented without extra http server, it would be the best solution possible.

**Describe alternatives you've considered**
I wonder if it is possible to reverse proxy the viewer server so that the port for pdf viewer and vscode can be the same. Alternatively the preview url prefix should be able to be configured so that I can reverse proxy the viewer on a subpath using nginx. At the same time the port number should be fixed.

jlelong:
Can you test PR #1271.

tamuratak:
Since VS Code Remote Development is available now, I think we do not need to support code-server. 

To my opinion, code-server itself should support port forwarding for WebView as VS Code Remote Development does. This is an upstream issue.

PR #1271, opened 2019-04-03
Make url and port viewer configurable

jlelong:
Use the following configuration variables

    latex-workshop.viewer.pdf.internal.url
    latex-workshop.viewer.pdf.internal.port

This is related to #1265.

### DIFF
--- a/package.json
+++ b/package.json
@@ -777,6 +777,11 @@
           ],
           "markdownDescription": "PDF viewer used for [View on PDF] link on \\ref."
         },
+        "latex-workshop.viewer.pdf.internal.port": {
+          "type": "number",
+          "default": "0",
+          "markdownDescription": "Define the port to listen on for communicating with the internal viewer. The default value \"0\" means the port is chosen randomly by the application."
+        },
         "latex-workshop.view.pdf.external.command": {
           "type": "object",
           "default": {
--- a/src/components/server.ts
+++ b/src/components/server.ts
@@ -2,6 +2,7 @@ import * as http from 'http'
 import * as ws from 'ws'
 import * as fs from 'fs'
 import * as path from 'path'
+import * as vscode from 'vscode'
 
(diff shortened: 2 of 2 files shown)


tamuratak:
From the viewpoint of security, I am against making host configurable, which would unintentionally allow remote access to local pdf and other files. That would require us to implement access control of the server. I want to avoid that.

We should force users to use a reverse proxy with access control, e.g., ssh port forwarding.

I think making port configurable is not a problem.

imxieyi:
> From the viewpoint of security, I am against making host configurable, which would unintentionally allow remote access to local pdf and other files. 
I agree. Actually I am using HTTP basic authentication for the whole code-server. A warning could be added to warn user about potential risks.

I think only advanced users who are aware of this risk as well as how to prevent it will bother modifying this configuration. So this is not a severe issue.

jlelong:
Would it be sufficient to make to have port configurable?

imxieyi:
No. Even if I use nginx to reverse proxy it, it is still not possible to launch viewer from ide since the url is still localhost.

tamuratak:
I am strongly against making host configurable. Users often copy and paste their settings from blog entries and stackoverflow without thinking about security risks. Assuming "only advanced users" is too optimistic.

Using nginx is not an appropriate solution. Please use ssh port forwarding. Then, it would be sufficient to have port configurable.

imxieyi:
Is it possible to add cookies check or some authentication to the preview server? I think this can be enabled with a custom url.

jlelong:
The discussion seems to have gone beyond the original scope of the PR. For now, it will only merge the port configuration feature and will leave any further improvements on supporting `vscode-server` to the community. PRs are welcome.

fockez:
I want to provide code-server in jupyter docker with jupyterhub, so the PDF viewer still needed.

PR #1274, opened 2019-04-03
lualatex rule added

PierreMarchand20:
As discussed in #1157, it adds the compilation progress feature for lualatex. But, there is still a bug concerning the ordering of the pages, I hope @tecosaur will take a look.


tecosaur:
Hi @PierreMarchand20, it's great to see you looking into luatex support! I read what you said about the issue you're having in #1157. While I'd love to take a look, unfortunately, I'm quite busy at the moment, and can't see it letting off for a good few weeks. What I can do though is point you in a few (hopefully useful) directions.

1. I found getting the log from "LaTeX Compiler" in the output tab of vs code, copying it, and pasting it into an online regex debugger like https://regex101.com/ to be quite helpful for working out what's going on (make sure to set it to javascript mode).
2. You seem to have it working now but since there seemed to be some confusion — to be clear these are the steps to get a working development version of an extension
    2.1. Clone repo
    2.2. Open repo (folder) in vs code
    2.3. run `npm install` within repo
    2.4. Press `F5` to run+debug
    2.5. If extension separately installed, vs code will ignore it and use your version
3. Re: the funny stuff happening with page numbers, from memory (other than regex issues) there was some precursory pass that seems to be run that was created extra matches. At some point I gave the page info a unique datstamp so it wouldn't be overriden. This lead to the duplicate pages. To fix this I simply undid that change. This change and un-change applies to all rules, so I'm not really sure whats happpening in your case…

Other than that, while it's not terrible I'd rate my code about 7/10 for quality, so there may be some hidden gremlins you need to unearth. Hopefully not, and I can't think of anything offhand other than not getting around to reformating the rule structure to make it more easily extendible, but it's worth noting.

PierreMarchand20:
I ended up doing exactly what you said in points 1) and 2) (I was quite relieved when I discovered regex101 xD).

Since I am starting to write my phd thesis and I do not know any javascript or typescript, I am not sure I will be able to do much more, I will try to take a quick look tomorrow, but can you point out to the code that does what you say in point 3) 

But if the pages are not built in a sorted order, may be it would be better to do a histogram than a linear plot, so that we can update the timing in an unsorted order. (I am not sure I am clear ahah)

tecosaur:
Erm. Everything I have seen so far has pages done in sorted order…

re: point 3, the code I mentioned was (as mentioned) later removed.

If it helps though, this is the bit that ads a new page entry

https://github.com/James-Yu/LaTeX-Workshop/blob/70e85fbd963593b1fa4eccde4481c30da3473181/src/components/buildinfo.ts#L686-L688

This is within the `displayProgress` function.

You could probably add a `console.log` whwnever the regex matches which should let you see with certainty the order in which you are given the pages.

PierreMarchand20:
Actually, even with pdflatex I get unsorted pages number resulting in weird plot (and the colouring is not really nice for dark theme also)
image
image

tecosaur:
Hmm.
 - Haven't seen any page order issues recently (screenshot is of a 41-page doc)
 - Have seen colour issues. I used vscode css colour variables so something might have changed upstream

image

 - Also noticed font issue
image

I'll be able to take a look at the color / font issues this weekend at the earliest, within a fortnight at the latest.

Sorry I'm not much help with the page numbers :(

As a short-term solution have you tried implementing a sorting function? Something like (incomplete code)
```javascript
arr.sort((obj1, obj2) => {
    return obj1.pageNo > obj2.pageNo;
})
```
(I forget the precise syntax, order, etc. but that _should_ be about right)

PierreMarchand20:
Well using a histogram, let us say you get the page 3 at first, you can put a column at the abscisse 3, then you get the first page, you can add a column at the abscisse 1. And you update the histogram like that, just adding column at the right abscisses.

You cannot do that with a linear plot, but I agree it would be more difficult to put several histogram in the same plot.

tecosaur:
Fixed colours and font in https://github.com/tecosaur/LaTeX-Workshop/commit/453699a94cdb1159fae642dd288ba311e1714c73

PR #1280, opened 2019-04-05
option to configure rootFile explicitly in (workspace)settings

b4ckup:
Sorry, the other pull-request was on the wrong branch...
I made a setting *latex-workshop.rootFile* available and implemented a convenient command *latex-workshop.configureRootfile* that let's you choose a rootFile from **\*\*/\*.{tex,latex}** files in the current working directory and saves the setting to the workspace configuration. I think this is really useful, as I would always choose defining the rootFile explicitly over some kind of magic to find it. Also it is very convenient to switch rootFiles via *vscode.showQuickPick()*, although during the most projects I think one workspace equals to one rootFile.
I placed the order first in the *manager.findRoot()* order, so the eplicit config will always be first. This might be debatable though

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,6 +170,11 @@
     ],
     "commands": [
       {
+        "command": "latex-workshop.configureRootfile",
+        "title": "Choose your LaTeX rootFile from the current directory",
+        "category": "LaTeX Workshop"
+      },
+      {
         "command": "latex-workshop.navigate-envpair",
         "title": "Navigate to matching begin/end",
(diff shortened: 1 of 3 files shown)


jlelong:
I am not sure to understand what is the advantage of this new approach over using the magic comment `% !TEX root = relative/or/absolute/path/to/root/file.tex`, see the [wiki](https://github.com/James-Yu/LaTeX-Workshop/wiki/Multi-File-Projects#the-root-file).

PR #1282, opened 2019-04-06
Fix #1281. Add option to use previous recipe on build

### LINKED ISSUE #1281: Use the last used latex recipe on file save
I regularly use several latex recipes, not only the first one in latex-workshop.latex.recipes.
It would be great and natural that auto-build process use by default the last used recipe in command palette (which is not necessarily the first one in latex-workshop.latex.recipes).
Is there a way to achieve this?

(There is no  magic comments in the root file)

latex-workshop

trevorgunn:


James-Yu:
It's a good feature. Please add a config item to make it configurable, and set the original behavior as default.

jlelong:
@James-Yu what do you think of also using this config item for the command `latex-workshop.build` (`ctrl+alt+b`) ? In this case, we should probably rename `latex-workshop.latex.autoBuild.recipe` to `latex-workshop.latex.build.defaultRecipe` or any better suggestion.

James-Yu:
I like the a lot. What about `latex-workshop.latex.recipes.default`? Yours are also fine for me.

trevorgunn:
@jlelong Sorry, I should have been clearer about what my code does.

It already applies to both `latex-workshop.build` as well as autoBuild (and any other feature that would call a build without specifying a recipe).

Possibly this is bad software engineering on my side but I didn't see any reason why autobuild would want to have the toggle but build would always use the first recipe. So I just stuck the new code in wherever `let recipe = recipes[0]` appeared and now it applies to both.

jlelong:
@trevorgunn what you did is fine, I should have been more careful. I will fix the wiki accordingly.

jlelong:
@James-Yu  the name `latex-workshop.latex.recipes.default` cannot be used because we already have `latex-workshop.latex.recipes` and an item cannot be the prefix of an other one. To keep things simple I will use `latex-workshop.latex.recipe.default`.

James-Yu:
Absolutely fine. Sorry for ignoring that 🤣

Issue #1278, opened 2019-04-05
$$ Equation preview show an extra "$" at left


Enn83:
The equation preview in latex file (when the mouse is over an equation) show an extra "$" when use two "$$" like, for example : $$ \Sigma^* = \bigcup_{n\in\doubleN} \Sigma^n $$, even if they are space between the "$$" and the equation.

To reproduce, simply type an equation bewteen $$ ......... $$.

 - OS: Windows 10 1809
 - VS Code version: 1.32.3
 - Extension version: 6.3.0

Here an ilustration :
![Double $](https://user-images.githubusercontent.com/46783975/55628041-f3a3c380-57af-11e9-8879-98996a9d50fe.jpg)

Good luck with LaTeX-workshop !

tamuratak:
`$$` is not supported. See https://github.com/James-Yu/LaTeX-Workshop/wiki/Hover#previewing-equations.

trevorgunn:
You [shouldn't be using $$ in LaTeX](https://texfaq.org/FAQ-dolldoll) to begin with.

However, this isn't the only issue with the current regex: https://regex101.com/r/bijiXw/1

I've created a more robust regex that does what we want: https://regex101.com/r/ksPutY/1

`(?<!\$|\\)\$(?!\$)(?:\\.|[^\\])+?\$|\\\(.+?\\\)|\$\$(?:\\.|[^\\])+?\$\$`

Someone just has to figure out how to pass the `$$...$$` matches to MathJax so it doesn't render an extra pair of dollar signs.

Enn83:
Thanks for all your answer, in fact I've not seen $$ ... $$ wasn't supported. 

From what trevorgunn sent me, I'll start using \[ and \] in my new documents.

PR #1284, opened 2019-04-07
Support preview for $$...$$

jlelong:
@tamuratak Can you have a look at this to make sure I have understood everything right?

### DIFF
--- a/src/providers/hover.ts
+++ b/src/providers/hover.ts
@@ -284,13 +284,13 @@ export class HoverProvider implements vscode.HoverProvider {
     }
 
     private stripTeX(tex: string) : string {
-        if (tex.startsWith('$') && tex.endsWith('$')) {
-            tex = tex.slice(1, tex.length - 1)
-        }
-        if (tex.startsWith('\\(') && tex.endsWith('\\)')) {
+        if (tex.startsWith('$$') && tex.endsWith('$$')) {
             tex = tex.slice(2, tex.length - 2)
-        }
-        if (tex.startsWith('\\[') && tex.endsWith('\\]')) {
+        } else if (tex.startsWith('$') && tex.endsWith('$')) {
+            tex = tex.slice(1, tex.length - 1)
+        } else if (tex.startsWith('\\(') && tex.endsWith('\\)')) {
+            tex = tex.slice(2, tex.length - 2)
+        } else if (tex.startsWith('\\[') && tex.endsWith('\\]')) {
             tex = tex.slice(2, tex.length - 2)
         }
         return tex
@@ -402,7 +402,7 @@ export class HoverProvider implements vscode.HoverProvider {
             const envname = this.getFirstRmemberedSubstring(document.getText(r), envBeginPat)
             return this.findHoverOnEnv(document, envname, r.start)
         }
-        const parenBeginPat = /(\\\[|\\\()/
+        const parenBeginPat = /(\\\[|\\\(|\$\$)/
         r = document.getWordRangeAtPosition(position, parenBeginPat)
         if (r) {
             const paren = this.getFirstRmemberedSubstring(document.getText(r), parenBeginPat)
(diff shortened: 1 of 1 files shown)


trevorgunn:
This doesn't work because the current regex doesn't capture `$$...$$`, it captures `$$...$` plus other undesirable behaviour which I have outlined in my previous comment https://github.com/James-Yu/LaTeX-Workshop/issues/1278#issuecomment-480553004

The regex I posted: `(?<!\$|\\)\$(?!\$)(?:\\.|[^\\])+?\$|\\\(.+?\\\)|\$\$(?:\\.|[^\\])+?\$\$` (which goes in lines 530 and 545 in your version of the file) fixes this and I would have tried to send a pull request with similar changes to yours except I ran into some unusual behaviour:

example

I wasn't sure what could be causing this. It turns out this was there before I made any changes. Somehow the cursor is passed to MathJax as well as the tex.

tamuratak:
It is difficult to know which `$$` is the start of math environment. We must parse entire TeX document to know that.

```
$$
a
$$
b
$$
c
$$
```


So, we had better leave `$$` unsupported.

I want to use @trevorgunn's regex to prevent hovers on `$$`.

trevorgunn:
That's another good point. If we keep the new code in stripTex and use my regex, we will be able to handle $$...$$ as long as both pairs of $$'s are in the same line and otherwise it will ignore it. This might be an acceptable compromise.

tamuratak:
@jlelong, if you want to modify complex code base like hover preview, please consider writing unit tests seriously. Without tests, it is difficult to make sure these PRs works well. 

#969 is out of date now. I will renew it if you want.

jlelong:
@tamuratak I fully agree with regarding unit tests. Unfortunately, this is worth it only if we can write exhaustive unit tests and this is far beyond the time we devote the development of this extension. All the more so as, we would need both unit and functional tests to validate most of LW's features.

@trevorgunn I have the feeling we kind of over think on the  `$$` preview issue. Preview **is triggered by the user**, so this is up to the user to hover on opening `$$` and not on closing ones. In response to [comment](1284#issuecomment-480635936), there is no need to parse the whole document to compute matching pairs. With this is mind, I cannot find a case in which the user hovers on a valid position and the preview is badly rendered. If so, post a minimal working example.

> This doesn't work because the current regex doesn't capture `$$...$$`, it captures `$$...$` plus other undesirable behaviour which I have outlined in my previous comment https://github.com/James-Yu/LaTeX-Workshop/issues/1278#issuecomment-480553004

`$$....$$` are not detected by the same regex as `$..$`. First, we test `$$...$$` and if there is no match we use the regex. See the function [findOverOnTex](https://github.com/James-Yu/LaTeX-Workshop/blob/b47ccca38821a596ee960c4c5c11807ddad33c4c/src/providers/hover.ts#L398).

@trevorgunn for the cursor issue, see this variable `latex-workshop.hover.preview.cursor.enabled`.

tamuratak:
@jlelong  if writing tests harms your motivation, I do not push that strongly. Keeping motivation has the highest priority for OSS development.

trevorgunn:
@jlelong This is the test file I am using

    \documentclass{article}

    \begin{document}
      Double dollar signs match from the first to third $$abcd$$

      Matching between double dollar signs $$abc$$ def $$ghi$$

      Escaped dollar signs are still caught: \$2 is $1 + 1$ dollars

      $$
      eq^1
      $$
      This should not be hover-able.
      $$
      eq_2
      $$
    \end{document}

Each group is its own minimal test.

Current behaviour (master branch):
 - First line hovers on `$$abcd$` and displays `$abcd` (which we know already)
 - Second line hovers on `$$abc$` and `$def$` and `$ghi$`
 - Third line hovers between `$2 is$`
 - Fourth group does nothing

With your changes (preview branch):
 - First three lines have the same behaviour
 - Fourth group will hover on the first second and third pair of `$$`. First and third do what we want, second gives the "This should not be hover-able." text

With your changes + adding negative look-arounds to the regex (i.e. the first `\$` becomes `(?<!\$|\\)\$(?!\$)`:
  - first line behaves well although causes issues hovering at the last two dollar signs
 - second line will capture `$$ def $$` if you hover at the second pair of dollar signs but is otherwise okay
 - third line is okay
 - fourth group same as above

Treating `$$...$$` as an inline hover avoids this issue of hovering at the second pair of `$$` but won't handle multiple lines. I think `$$...$$` has enough quirks that it will be hard to get it exactly right. So far treating `$$...$$` inline only has had the best behaviour in my opinion.

jlelong:
@tamuratak To be clear I am in favour a TDD. For LW, it seems to be a really tough work and I do not know how to use tests with CI for vscode's extensions.

@trevorgunn thanks for your examples and detailed explanation, I will finish this in the coming days.

PR #1300, opened 2019-04-12
Merge latex-mathsymbols_cmd.json into commands.json

tamuratak:
This is a PR to merge `data/packages/latex-mathsymbols_cmd.json` into `data/commands.json`. Commands in `latex-mathsymbols_cmd.json` are defined by default in LaTeX. So, we should merge them into `commands.json`.

`dddot` and `ddddot` are deleted because they are defined in `amsmath_cmd.json`.

### DIFF
--- a/data/commands.json
+++ b/data/commands.json
@@ -253,7 +253,9 @@
   },
   "underline":{
     "command":"underline",
-    "snippet":"underline{${1}}"
+    "snippet":"underline{${1}}",
+    "detail":" ̲",
+    "documentation":"COMBINING LOW LINE"
   },
   "rule":{
     "command":"rule{width}{height}",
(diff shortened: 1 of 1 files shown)


jlelong:
> Commands in `latex-mathsymbols_cmd.json` are defined by default in LaTeX. So, we should merge them into `commands.json`.

An alternative approach would be automatically load some `_cmd.json` based on the type of document, see [here](https://github.com/texstudio-org/texstudio/blob/master/completion/cwlAliases.dat)  for a list of packages to be loaded whatever `usepackage` directives we have.

tamuratak:
As you like. When you choose the alternative approach, please merge #1307.

tamuratak:
See #1313.

PR #1301, opened 2019-04-13
add some details into commands in amsmath_cmd.json

tamuratak:
add some details into commands in `data/packages/amsmath_cmd.json`.

### DIFF
--- a/data/packages/amsmath_cmd.json
+++ b/data/packages/amsmath_cmd.json
@@ -117,27 +117,37 @@
   "iiiint": {
     "command": "iiiint",
     "package": "amsmath",
-    "snippet": "iiiint"
+    "snippet": "iiiint",
+    "detail": "⨌",
+    "documentation": "Quadruple integral operator"
   },
   "iiint": {
     "command": "iiint",
     "package": "amsmath",
-    "snippet": "iiint"
+    "snippet": "iiint",
+    "detail": "∭",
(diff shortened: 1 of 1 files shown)


jlelong:
All `_cmd.json` files are automatically generated from [`cwl` files](https://github.com/LaTeXing/LaTeX-cwl/). So any change in these files will lost when we update them. With this in mind, your PR may get lost in the future. The same applies to #1303.

James-Yu:
What about we introduce a new file for the package, e.g., `amsmath_msg.json` or other names apply, to store the custom information? The json objects stored here can be merged with those in `_cmd.json`, and changes persist through potential future script re-runs.

tamuratak:
These PRs are generated by a script. So, I will try to rewrite scripts under `dev` into ones generating these details and documents.

jlelong:
@tamuratak do you say that you have a way to retrieve the details automatically from somewhere else?

@James-Yu an other and perhaps a little simpler approach would be to modify the scripts in `dev` to first load the `_cmd.json` if it exists to retrieve the `details` and `documentation` fields and then insert these fields into the new generated file from the `cwl`.

tamuratak:
Yes, from this file, http://milde.users.sourceforge.net/LUCR/Math/data/unimathsymbols.txt. We have generated `data/unimathsymbols.json` from this file.

tamuratak:
See #1305.

PR #1302, opened 2019-04-13
Update commands.json

innerlee:
Save a little space.

### DIFF
--- a/data/commands.json
+++ b/data/commands.json
@@ -22,13 +22,13 @@
   },
   "equation":{
     "command":"equation",
-    "snippet":"begin{equation}\n\t\\label{${1:<label>}}\n\t$0\n\\end{equation}",
+    "snippet":"begin{equation}\\label{${1:<label>}}\n\t$0\n\\end{equation}",
     "detail":"Insert an equation environment",
     "label": "\\begin{equation} ... \\end{equation}"
   },
   "align":{
     "command":"align",
-    "snippet":"begin{align}\n\t\\label{${1:<label>}}\n\t$0\n\\end{align}",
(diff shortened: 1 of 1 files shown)


trevorgunn:
Align probably shouldn't have a label in it's autocomplete now that I think about it. Since it tags each line, you'd expect to be putting a label on each line not just the first one.

innerlee:
How about deleting labels for both equation and align. It is hard for me to think about a label before writing the equation.

trevorgunn:
My workflow is usually:

- put equation in `\[...\]`
- later realize I would like to reference it
- add an equation environment and copy

So for me, it is beneficial to autocomplete a label for equation.

innerlee:
According to the [Fisher's rule](http://www.pamitc.org/documents/mermin.pdf) of ``number all displayed equations``, it is a good habit to use `\equation` rather than `\[ \]`

trevorgunn:
Numbering every equation is far from the norm in mathematics from what I've seen. I think physics might use that style as the norm in which case it might make sense not to have a label in as default since most equations aren't later referred to. Maybe it is a good idea to add an option to keep or suppress the labels since it seems common enough for people to want different behaviour.

innerlee:
Sure there are different styles. But considering other latex commands such as `\section`, which do not insert labels for the authors, I prefer leave the labels for the authors. Adding them is not hard since `\label` has intellisense suggestions, but deleting them is really a pain.

James-Yu:
I generally agree that we may remove `\label` from the snippets. It's really an undetected pain for me when writing. Any other comments?

innerlee:
Agree removing the label. Feel free to close this pr and make the change :)

PR #1317, opened 2019-04-16
Add contributing.md

trevorgunn:
It contains only build instructions right now but I hope it will help some people (it would have helped me to have this).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,20 @@
+# Contributing to LaTeX-Workshop
+
+## Prerequisites for building the extension
+
+Make sure you have installed:
+
+- [`node.js`](https://nodejs.org/)
+- `npm`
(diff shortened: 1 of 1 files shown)


jlelong:
Good idea to write those things!
Actually, I do not have `yarn` installed. I only run `npm install` and it installs everything. I have never really figured out how the two work together.

trevorgunn:
One of the reason's I made this was because all of the instructions I found for building extensions were with npm and this repository has a `yarn.lock` rather than a `package-lock.json`. So I wanted to figure out how to do it with `yarn`. It's exactly the same: just replace `npm install` with `yarn install`. The difficult part is figuring out how to undo `npm install` so those packages don't conflict with `yarn`.

It may be worth mentioning something about using `npm install` instead and adding `package-lock.json` to `.git/info/exclude`.

PR #1321, opened 2019-04-20
Make go back and forward buttons on pdf viewer work with SyncTeX

tamuratak:
1. add a forward button on pdf viewer.
2. make the back and forward buttons work with SyncTeX. They also work with clicking links as before.

Apr-20-2019 12-57-25


James-Yu:
Simply. Too. Awesome.

Tell me when ready for merging.

tamuratak:
Thank you.  It is ready now.

PR #1323, opened 2019-04-20
Less escapes in wordpattern (also fixes #1322)

### LINKED ISSUE #1322: Usage of snippets without prefix in equation
Hi,

First, many thanks for this wonderful plugin that I use every day.

I have a tiny suggestion : while typing, I frequently use some custom snippets. For example, if I type "frac" + TAB, then it insert "\frac{}{}" (with placeholders).

But if I do this at the beginning of an equation delimited by "$" (so, if I type "$frac"), that snippet doesn't work.

Looks like it depends of the "wordPattern" setting in "syntax/syntax.json", and adding "\\$" in the regex seems to do the job : 
```
"wordPattern": "(__)|(\\*\\*)|(\\@.)|(\\.\\.\\.)|([^\\$\\`\\~\\!\\#\\%\\^\\&\\*\\(\\)\\-\\=\\+\\[\\{\\]\\}\\\\\\|\\;\\:\\'\\\"\\,\\.\\<\\>\\/\\?\\s\\@_]{2,})"
```

Could you please modify that regex ?

Thanks !

trevorgunn:
I didn't see a reason why `@` needed to be escaped or why most of the characters in `[^ ... ]` needed escapes. I did some tests to make sure it still works the same.


jlelong:
The variable `wordPattern` is mainly used to properly trigger intellisense. Your simplification does not seem to have side effects on snippets. Let us merge.

PR #1324, opened 2019-04-21
Show number given to ref in hover on the ref

tamuratak:
This is a PR to show a number given to ref in hover on the ref. The numbers are read from `aux` files just after compilation finished. Ideally, the numbers should be displayed in the equation by MathJax using `\tag` command. However, the support of `\tag` by MathJax is not mature. 

I recommend that we leave this feature disabled by default and wait for new releases of MathJax.

スクリーンショット 2019-04-21 8 52 41

### DIFF
--- a/package.json
+++ b/package.json
@@ -1103,6 +1103,11 @@
           "default": true,
           "markdownDescription": "Enable Hover on References."
         },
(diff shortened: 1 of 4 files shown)


tamuratak:
I close this PR to look into using `\tag`.

PR #1326, opened 2019-04-21
Show label or given number to ref in hover on the ref

tamuratak:
This is a PR to show a label or a number given to the ref on the hover. The numbers are read from the aux file just after compilation finished.

When enabled and after compilation.

スクリーンショット 2019-04-21 16 58 12

When disabled or not yet compiled.

スクリーンショット 2019-04-21 17 02 07

### DIFF
--- a/package.json
+++ b/package.json
@@ -1103,6 +1103,11 @@
           "default": true,
           "markdownDescription": "Enable Hover on References."
         },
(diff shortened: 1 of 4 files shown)


tamuratak:
@James-Yu this is ready to merge.

tamuratak:
Thanks.

PR #1349, opened 2019-04-30
add ip address setting when view PDF file in web browser

xfdywy:
Let the users choose the IP address when using web browser to view the PDF file.  This feature is  convenient for the user who wants to view the PDF file on other machine or share with other people.

When the IP address is set to "0.0.0.0" the web browser need to open "127.0.0.1".


jlelong:
See #1271 for a discussion on this.

Issue #1314, opened 2019-04-15
Synctex reverse


atomdistortion:
Issue Type: <b>Bug</b>

It is often, when reverse Synctex does not work, especially when I switch it to another LATEX project or other generated file in the same project. And this function often starts work  after removal of synctex files and generating new pdf. But it happens not always. Sometimes this functionality appears suddenly, in other cases I have to restart VSCode once or two times, in hard cases it is necessary to restart system.

Extension version: 6.4.0
VS Code version: Code 1.33.1 (51b0b28134d51361cf996d2f0a1c698247aeabd8, 2019-04-11T08:27:14.102Z)
OS version: Windows_NT x64 10.0.17763

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz (8 x 3600)|
|GPU Status|2d_canvas: enabled<br>checker_imaging: disabled_off<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>rasterization: enabled<br>surface_synchronization: enabled_on<br>video_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Memory (System)|15.87GB (9.55GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|

</details>

innerlee:
Heisenbug, don't know how to trigger it, but it happens often

atomdistortion:
Probably this happens with non-english languages, for ex. I'm writing in russian.

tamuratak:
Please notice that LaTeX Workshop does not support multiple projects in one window. See https://github.com/James-Yu/LaTeX-Workshop/issues/12#issuecomment-279634020.

lyyc199586:
same issue

atomdistortion:
That helps (build LaTeX project, I choose right-mouse menu). I don't have many projects, just several separate files in the same workspace, they use common data. Nevertheless, that is not convenient. I save a project by Ctrl-S (then it build the project), and choose the root file by the left-side browser, so could be it possible to check chosen file and build it or customize that ability.

tamuratak:
@atomdistortion, I understand that you are proposing a feature request. Is that right? If so, please close this issue and open a new issue.

[...]

PR #1350, opened 2019-04-30
check whether selection.anchorNode is null.

tamuratak:
Related to #1314.  `selection.anchorNode` is always set a value, not `null`. So, normally, we do not have to check whether it is `null`. However, in this [log](https://github.com/James-Yu/LaTeX-Workshop/issues/1314#issuecomment-487858272), it is said to be `null`. That might be a bug of VSCode.


tamuratak:
`selection.anchorNode` can be null. [spec](https://www.w3.org/TR/selection-api/#selection-interface).

PR #1360, opened 2019-05-06
Add command to switch between equation environments.

trevorgunn:
Example of feature

I wanted to add this feature because I often find myself needing to change from `\[ ... \]` to `\begin{equation} ... \end{equation}` if I want to add a label later or to change to `\begin{align*} ... \end{align*}` if I need multiple lines.

The code is pretty basic, it just searches backwards and forwards until it finds the beginning and end of some equation environment and then does a replace. As such, it can produce garbage if the user feeds it a bad starting point. Nonetheless, I'm happy with how it works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -195,6 +195,11 @@
         "category": "LaTeX Workshop"
       },
       {
+        "command": "latex-workshop.switch-equation-environment",
+        "title": "Switch equation environment",
+        "category": "LaTeX Workshop"
+      },
+      {
         "command": "latex-workshop.saveWithoutBuilding",
         "title": "Save without Building",
         "category": "LaTeX Workshop"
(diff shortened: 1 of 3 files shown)


jlelong:
Some functions already exist to select the current environment name, see the [wiki](https://github.com/James-Yu/LaTeX-Workshop/wiki/Environments#navigating). I guess we would just need to improve them to also take into account `\[ .... \]`.

trevorgunn:
Yes, I saw some of the functions in `envpair.ts` but they were designed around `begin{...}` and `end{...}` and I didn't take the time to figure out how deeply ingrained that is and how to modify it.

So the current plan is: I will close this pull request and modify my command to use `envpair.locateMatchingPair()`. I don't think there's a good way to do what I want to do with the current commands since there's no good way to modify those to switch between `\[ ... \]`. But hopefully we can reuse the helper functions which anyways do a better job finding matching pairs than my code.

PR #1345, opened 2019-04-28
Fix #1342. Support single and double line font warnings.

trevorgunn:
Fixes #1342 

Example warnings:
```
LaTeX Font Warning: Font shape `TU/SimSun(0)/bx/n' undefined
(Font)              using `TU/SimSun(0)/m/n' instead on input line 22.

LaTeX Font Warning: Some font shapes were not available, defaults substituted.

LaTeX Font Warning: Font shape `OT1/cmss/m/it' in size <10.95> not available
(Font)              Font shape `OT1/cmss/m/sl' tried instead on input line 15.

LaTeX Font Warning: Font shape `OT1/cmr/m/n' in size <15> not available
(Font)              size <14.4> substituted on input line 3.

LaTeX Font Warning: Size substitutions with differences
(Font)              up to 0.6pt have occurred.
```

~~There is already some support in the parser for multi-line errors but that support doesn't parse the `on input line xx` part. This handles the second line separately.~~

Update: refactored to reuse the old code block for multi-line errors. This should be a bit more efficient and more importantly: easier to maintain.

PR #1361, opened 2019-05-06
Extend font warning second line to other warnings.

trevorgunn:
While working on a paper, I realized that the font warning second line that I added to the parser in #1345 applies to other packages. Notably to `hyperref` if you try to stick math into your section names:

```
Package hyperref Warning: Token not allowed in a PDF string (PDFDocEncoding):
(hyperref)                removing `math shift' on input line 149.
```
Another example:
```
Package remreset Warning: The remreset package is obsolete:
(remreset)                \@removefomresset is defined.
```
Unfortunately `hyperref` places the error where the next line of text is. For example the error on line `149` was actually on line `147`. But this is an external issue.


trevorgunn:
It seems like this can be more than just two lines:
```
Package hyperref Warning: Composite letter `\textasciicircum+n'
(hyperref)                not defined in PD1 encoding,
(hyperref)                removing `\textasciicircum' on input line 149.
```
From `\section{\texorpdfstring{$\mathbf{P^n}$}{P\^n}}`

I'm modifying this now to handle these warnings better.

jlelong:
The message parser is very tricky so we really need heavy testing before merging to make sure it does not break something else.

trevorgunn:
I've so far tested all the basic errors:
 - badboxes
 - text command inside math
 - math command inside text
 - \begin not matched with an \end
 - bibliography errors:
   - missing references
   - bad commands in bib entry
   - rerun to get references
   - missing bibliography file

I haven't seen any regression from the current parser. Additionally, the new second line regex catches everything I intended it to (and nothing more).

PR #1362, opened 2019-05-06
Add command to switch between equation environments.

trevorgunn:
I've updated #1360 to use `envPair.locateMatchingPair()`. Functionality should remain the same.

Again, I'm creating a new command because I don't believe the current commands (e.g. `latex-workshop.select-envname`) can be made to work with `\[ ... \]` well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -195,6 +195,11 @@
         "category": "LaTeX Workshop"
       },
       {
+        "command": "latex-workshop.switch-equation-environment",
+        "title": "Switch equation environment",
+        "category": "LaTeX Workshop"
+      },
+      {
         "command": "latex-workshop.saveWithoutBuilding",
         "title": "Save without Building",
         "category": "LaTeX Workshop"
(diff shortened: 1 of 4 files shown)


trevorgunn:
There are some issues with inline switching that need to be resolved.

trevorgunn:
I may do some more testing if I think of any potential issues, but the latest commit resolves all the issues I had.

jlelong:
This is a very interesting feature but I would like to make it work with any environment not just the ones listed in `equationEnvironments`. Maybe we could the `suggestions` computed in the environment completer, see [environment.ts](https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/providers/completer/environment.ts). I still need to think about it.

trevorgunn:
Between the current functionality of `latex-workshop.select-envname` and this command, I'm not sure which use cases what you have in mind would address.

jlelong:
I would like to merge this new functionality with `latex-workshop.select-envname` but as you pointed out it seems quite difficult. One possible solution would be 
- to use your new function for switching between `\[....\]` and `\begin{} .... \end{}` while leaving a double cursor inside the `{}`
- use `latex-workshop.select-envname` to change environment name.
Ideally, I would like to have the two features wrapped into the same command.

PR #1366, opened 2019-05-08
Fix buildinfo

### LINKED ISSUE #1358: Performance drop for long stdout
**Describe the bug**
A project I am building prints huge amounts (8.5M) of text to stdout. Building it via LaTeX Workshop (I have configured it to to use `xelatex`) takes hours while executing the same `xelatex` command in the shell takes less than 10 minutes.

**How to reproduce**
Steps to reproduce the behaviour:
1. Have a project which generates large amounts of stdout.
2. Build the project with the xelatex recipe.

**Expected behaviour**
It takes the same time as running the same `xelatex` command in the shell.

**Root cause**
Building via LaTeX Workshop is slower because it starts to consume all CPU so `xelatex` gets much less resources than it gets when it is run in the shell.

[The performance analysis](https://github.com/James-Yu/LaTeX-Workshop/files/3145672/James-Yu.latex-workshop-unresponsive.cpuprofile.txt) has shown that 99% of CPU is spent in `match()` in `checkStdoutForInfo()`. Is it necessary to match the whole stdout? Wouldn't matching only the last line be enough? Is it necessary to match the same regular expression on the same string twice?

Removing `this.checkStdoutForInfo()` on line 79 in `buildinfo.ts` resolves this issue.

**Standard output**
The project has a lot of plots. The build produces 70k lines like this (beside others):
`NOTE: coordinate (-2.99E-02,-2.69E-05) has been dropped because it is unbounded`

**Desktop:**
 - OS: Linux x64 5.0.10-200.fc29.x86_64
 - VS Code version: 1.33.1
 - Extension version: 6.5.1

tecosaur:
Closes #1358. I did a quick test with one document, and I'm pretty confident (p < 0.01) that this speeds things up.
Also fixes some styling issues that have cropped up (vs code update broke some minor visual things at one point).

### DIFF
--- a/src/components/buildinfo.ts
+++ b/src/components/buildinfo.ts
@@ -32,7 +32,7 @@ export class BuildInfo {
             pageTotal: undefined,
             lastStepTime: +new Date(),
             stepTimes: {},
-            stdout: '',
+            stdout: '\n'.repeat(50),
             ruleNumber: 0,
             ruleName: '',
             ruleProducesPages: undefined
@@ -75,7 +75,7 @@ export class BuildInfo {
         }
 
         for (const line of lines.split('\n')) {
-            this.currentBuild.stdout += '\n' + line
+            this.currentBuild.stdout = this.currentBuild.stdout.substring(this.currentBuild.stdout.indexOf('\n') + 1) + '\n' + line
             this.checkStdoutForInfo()
         }
     }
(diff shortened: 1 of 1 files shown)


tecosaur:
NB: Implements 50 line stdout buffer

PR #1367, opened 2019-05-08
Change select-envname to handle \[...\]. Add toggle-equation-envname command.

trevorgunn:
Based on discussion in #1360 and #1362 .

Does a couple things:
 - Moves `equation` and `align` environments back into `environments.json` since they are no longer doing anything special (this allows for auto-completion inside `\begin{...}`)
 - `selectEnvName` is now called `envAction`
   - old functionality is preserved for the most part except it will now treat `\[...\]` as an environment and automatically convert it to `\begin{equation*}...\end{equation*}`
  - can also convert between `\begin{}...\end{}` and `\[...\]` without changing the selection via the new `toggle-equation-envname` command


jlelong:
This works pretty well but may fail when `\[ ... \]` contains an environment.

Consider
````latex
\[
  \begin{split}
     x&=y\\
     &=z
  \end{split}
\]
````
If the cursor is inside the split environment, there is no way to convert `\[ ... \]` to `\begin{equation*} ... \end{equation*}`. We could try to make a list of non top-level maths environments and ignore them in `toggle-equation-envname` 

A second issue is when `\\[<dimension>]` (eg. `\\[2ex]`) is used to add extra vertical space in `amsmath` environments. 
````latex
\[
  \begin{aligned}		
    1   & 2, \\[1ex]
    3 & 4
  \end{aligned}	
\]
````
Neither `toggle-equation-envname` nor `multicursor-envname` work.

File: package.json [hunk 1]
```diff
@@ -194,6 +194,10 @@
         "title": "Close current environment",
         "category": "LaTeX Workshop"
       },
+      {
+        "command": "latex-workshop.toggle-equation-envname",
+        "title": "Toggle between \\[...\\] and \\begin{}...\\end{}"
+      },
       {
```
jlelong: Add `"category": "LaTeX Workshop"`. 
Every function provided by the extension should be put in this category.

trevorgunn:
Thanks for reviewing. The first one was the behaviour I intended although it should be easy enough to improve by using a different regex. I think it is easier to look for top-level math environments rather than avoiding non-top-level environments.

The second one, I'll admit I didn't even know was a thing, but a simple negative look-behind should resolve it.

I'll go work on this now.

jlelong:
Everything works fine.
Could you write a small documentation on this by making a PR on [Environments.md](https://github.com/jlelong/LaTeX-Workshop-wiki/blob/master/Environments.md)? Thanks.

PR #1375, opened 2019-05-11
add more intuitive snippets

morpheusthewhite:
I think many snippets are not intuitive so I added better ones

### DIFF
--- a/snippets/latex.json
+++ b/snippets/latex.json
@@ -29,6 +29,11 @@
 		"body": "\\infty",
 		"description": "Insert infinity symbol"
 	},
+	"infinity": {
+		"prefix": "inf",
+		"body": "\\infty",
+		"description": "Insert infinity symbol"
+	},
 	"partial": {
 		"prefix": "@6",
 		"body": "\\partial",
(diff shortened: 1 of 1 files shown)


trevorgunn:
I'd prefer if all the snippets provided by the extension had a special prefix like `\` or `@` so that they don't show up whenever I type a word beginning with "in".

You can [add your own user snippets](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_creating-your-own-snippets) if you want to do something different.

jlelong:
I agree with @trevorgunn.

morpheusthewhite:
Well then you should apply this policy to all of your snippets, from `item` to `SPG` and so on

jlelong:
I agree that `item` looks odd. All the others are explained in the [wiki](https://github.com/James-Yu/LaTeX-Workshop/wiki/Snippets) and follow a common pattern. Note that the prefixes are all uppercased to avoid mixup with standard typing. The `B` and `S` patterns come from [vim-latex](http://vim-latex.sourceforge.net)

PR #1377, opened 2019-05-13
Add sum snippet

morpheusthewhite:
I added sum snippet following conventions

### DIFF
--- a/snippets/latex.json
+++ b/snippets/latex.json
@@ -105,6 +105,11 @@
 		"body": "\\int_{$1}^{$2}$0",
 		"description": "Insert an integral"
 	},
+	"sum": {
+		"prefix": "@A",
+		"body": "\\sum_{$1}^{$2}$0",
+		"description": "Insert an sum"
+	},
 	"Big|": {
 		"prefix": "@|",
 		"body": "\\Big|",


tecosaur:
Erm. Why `@A`? To me that may not be a good choice because there are symbols such as `\forall` which seem more likely to be `@A` than `\sum`.

James-Yu:
Em did not notice the short cut, will revert shortly.

jlelong:
Snippets quickly become very personal, which makes me believe that we should refrain from adding a few more here and there if they do not comply with the already implemented mechanism (see discussion in #1375). There are basically six kinds of snippets

1. snippets starting with `\`
1. snippets starting with `@` to enter Greek letters or common mathematical symbols (very much inspired by `vim-latex`)
1. snippets of the form `BXY` (or `BSXY`) to type an environment where `XY` are the first two letters of the environment name. The `BSXY` snippets enter the starred versions.
1. snippets starting with `F` to enter font commands 
1. snippets starting with `M` to enter mathematical  font commands 
1. snippets starting with `S` to enter sectioning commands (inspired by `vim-latex` too).

Of course, all this could be redesigned if someone comes up with a better idea but meanwhile not following these guidelines may be more destabilising than useful. Anyway, users can always define their own snippets outside of the extension.

morpheusthewhite:
Thank you @jlelong for the clarifications, I'll be more careful in the next pull requests

jackey79:
If I want to add a new snippet `!=`, it doesn't work.

jlelong:
This is because `!=` is not recognised as a word, see `wordPattern` in [`syntax.json`](https://github.com/James-Yu/LaTeX-Workshop/blob/master/syntax/syntax.json). Autocompletion is only triggered on words, see [vscode#64818](https://github.com/microsoft/vscode/issues/64818).

jlelong:
@arcstur When the new release is out (not scheduled yet), you can define your new `@`-suggestions using `latex-workshop.intellisense.atSuggestionJSON.replace`. See https://github.com/James-Yu/LaTeX-Workshop/wiki/Intellisense#-suggestions

PR #1452, opened 2019-06-20
Paste format

### LINKED ISSUE #1451: Formated Paste
Why have one half-baked new feature when you can have two!

As such I would like to present another feature I'm working on **formatted paste**.
Thanks to ~~stealing~~ building off the code in https://github.com/mushanshitiancai/vscode-paste-image/ I have managed to make it so that you can paste images into latex documents.

By default this puts the image into the first `graphicspath`, or if there are in none, in the cwd.

For a preview see [tecosaur/LaTeX-Workshop/paste-format](https://github.com/tecosaur/LaTeX-Workshop/tree/paste-format)

### Status

 - [x] Images
 - [x] Tables
 - ~~Rich text~~
 - [x] performing common substitutions (e.g. `"a"` to ``` ``a''```)

tecosaur:
This is a rather tentative PR, it would be nice if someone could try it out a let me know if there are any issues, or feedback. It does seem to more or less work though.

### Formatted Paste

 - Adds `ctrl`+`shift`+`v` keybinding for formatted paste
 - Images can now be pasted
 - Tables are converted to tabular environments
 - Some convenient text replacements are carried out
 - Pasting paths to images or `csv`s result in a tabular/includegraphics
 - Also tweaked `getGraphicsPath`, and added the `moment` library

Would close #1451


tamuratak:
I am not a fan of the feature on pasting images. That offers a minor feature while the code base is large and increases the cost of maintenance.

The features on pasting tables and quoted strings seem promising.

tecosaur:
To me the main draw for this was precisely pasting images.
I have documents where I have included hundreds of images,
and each one I’ve had do do the following process,

 1. Right click, save image to file
 2. Navigate to graphics folder
 3. Give file name
 4. Copy file name
 5. Go to latex file
 6. Trigger sniper for inserting figure / image
 7. Paste file name

This isn’t too bad for just one or two images, however when doing for large patches it quickly becomes tedious and irritating. Thus for me this enhancement offers significant benefit. The new list of steps is:

 1. Right click, copy image
 2. Paste into vscode

With regards to maintenance cost, a large portion of the code is determine what the file name should be, where the file should be places, and which script to use. This should all be rather stable. The other component is the shell scrips. These are all rather small, I can’t imagine this will pose much issue. Furthermore since the image pasting codebase is from another vscode extension, and fixes should be able to be easily applied from the source repository.

tamuratak:
I recommend you to create your own VS Code extension for that purpose.  I think that this PR does not have to be integrated into LaTeX Workshop to provide the feature.

Thinking smart ways of treating paths of image files seems more promising.

James-Yu:
I'm rather neutral on this feature. On the one hand it is quite fancy, but on the other I agree with @tamuratak that the table one is more promising. Look at the many tools people created to convert excel to latex.

tecosaur:
Hmm. Having a smart way of treating pasted file locations sounds like a nice idea 😃.
I haven’t considered writing a seperate extension, one issue I may find with that it would have to add yet another shortcut. Then there would be a different keybinding for:
 - pasting unformatted text
 - pasting reformatted text / tables
 - pasting images

While this isn’t a huge issue, it’s definitely a progression in an annoying direction.
The other reason why I’d consider this to be worth putting into the extension is just my own view of this one, in that this extension provides a _complete_ set of tools to make creating latex documentent an easy and enjoyable experience.

I’ll take a look at identifying file paths and possible actions on pasting, but I’ll finish with saying that I can see what you’re saying about the pasting of images, but I still feel that this extension would be better with than without it.

tecosaur:
Also re: tables there are two features I would like to add (in order of significance)

 1. Recognising multicolumn cells, and dealing with them appropriately
 2. Recognising tables is blocks of text. One promising avenue may be in locating bunches of `\t` since I wouldn’t imagine they appear much outside of tables.

tamuratak:
Treating only the paths of images will not require the code base of vscode-paste-image, which reduce the size of this PR. And it would be much easier to maintain that.

jlelong:
I am not really fan of pasting images either and I am quite reluctant to introducing features, which require a lot of dedicated code because of obvious maintainability issues. Moreover, I think a drag and drop approach to automatically insert the appropriate `\includegraphics` instruction without moving the image file would already be a great improvement and would be far less demanding in terms of code base (I guess this is what @tamuratak says).

Being able to paste tables (or even drag and drop `.csv` file) would be really awesome.

tecosaur:
I like the idea of drag and drop image files — I'm currently working on that in fact (and also `.csv` files). :smiley: 

I know I'm harking on about copy/paste. Most of the time that's no less effort than drag+drop of files.
The reason why I like to have copy + paste is that for online images, or images embedded in other files, it removed the intermediate step(s) of saving, naming, and copying again.

The large block of code for that is potentially anoying though. Hmmm, I'll let you know if I come up with a solution to this.

tecosaur:
You can now paste paths to `.csv` files and images :smiley: 

image

tecosaur:
Looks like drag and drop won't be possible as that just opens the file in vscode (and I can't see and api bits for that), however pasting the file path works :)

tecosaur:
Ok. So it seems like the one thing holding up the merging of this is the contention regarding pasting images.

From what I can tell opinion is split like so:
 - James — Neutral
 - Jlelong — concerned about maintainability
 - Tamuratak — concerned about maintainability
 - Me — thinks it's a good idea

I'll have a closer look at the codebase and report back with a longer comment, this is what I'm going to try to do:

1. Separate image pasting stuff into own sub-class
2. Go through code and see how much is relatively safe standard javascript (i.e. us basic libraries like `path`, no vscode api stuff etc.), and how much seems less-so

@jlelong @tamuratak do you think you would be open to the idea of provisionally including this if it can be done in such a way that _if_  there are any issues with it in the future it can easily be cut out? I figure that it's nicer to have it than not, and if it's removed because something with it breaks and there's an uproar — then to me that indicates that it's used enough to justify a minor maintenance cost.

tamuratak:
> do you think you would be open to the idea of provisionally including this 

No.

tecosaur:
Heh. At least that's straightforward :stuck_out_tongue_closed_eyes:

tecosaur:
Ok. So in light of what's been said I'm going to do is remove the image clipboard to file functionality. I think I can still make things a bit easier though. Instead, I'll add an option to make it so that linked files outsider of the current/graphics dir can be automatically relocated/renamed. This should make things easier for say getting a file from the downloads folder, or one of a bunch of images extracted from another file.

tecosaur:
Right. So now that's done. Removing pasting from clipboard support removed 150 lines, so that a roughly 20% reduction in line count. I'm not sure that the maintenance reduction is worth the feature reduction, but there you have it.

I think it works, but it's not terribly well tested. I'd appreciate feedback on the current state of things.

James-Yu:
I am a bit of headache when looking through the new config items related to `image`. Other than that part, I myself vote a yes for the PR. How do you think? @jlelong @tamuratak 

It would be much better to expose less options and details to the users if possible, in my opinion. The current dependency/hierarchy/priority of image-related config is a bit confusing. This might because I did not follow the development tightly, yet the users also did not. If the others would agree on the overall concept, we may proceed to a little bit of extra clean up.

tamuratak:
@tecosaur I am against this PR because it still includes the feature of moving a image file. @jlelong and I are against the feature.

tamuratak:
Apart from the feature, I am not sure whether the code quality of this PR is enough high to be merged. If enabled by default, that is required.

tecosaur:
I think I should be able to go though this PR and simplify simplify some things — particularly the image code from that other extension.

tecosaur:
Ok. On the "bit of headache when looking through the new config items related to image" front; I can consolidate the image `namePrefix` and `nameSuffix` into `defaultName`. I also think it's unlikely that anyone would like to change the path settings (`graphicsPath` falling back to `currentFileDir`) so that may not need to be a setting. `showFilePathConfirmInputBox` and `filePathConfirmInputBoxMode` should also be able to be merged into a single enum. That would take the number of image options from 10 to 5.

@James-Yu sorry the code / options are a bit dodgy. I wanted to get something that works first — which this now seems to, before refining it. Unsually this would be adding options, but since they were copied over in this case it's cutting them down.

tecosaur:
There we go. I feel that makes config a fair bit nicer. Also reduces the PR size a bit. :smiley:
Next I'll have a look at some of the pasted code, and see if I can tidy it up a bit.

@tamuratak I thought your issue was with the clipboard to file bit? Sorry if I misunderstood. If you have an issue with the file interaction, are you concerned about the stability of `fs.rename` or `fs.copyFile` — or do you not like the determining of the target image file path?

tamuratak:
We do not need the feature of moving the image file itself. No one except you is positive on it.

tamuratak:
The proposed way of treating image files is far away from the standard one.

tecosaur:
@tamuratak I see.
Re: the existance of the feature — can you think of any other methods that would help resolve https://github.com/James-Yu/LaTeX-Workshop/pull/1452#issuecomment-503948806, I'm not trying to be difficult, but that genuinely is a significant issue for me, and the driving factor behind the creation of this PR.If you can think of a better way of solving it I'm all ears.
Re: implementation —  would you be able to propose a more 'standard' approach?

tamuratak:
I recommended you to create your own VS Code extension for that purpose in the past post.

tecosaur:
Hmm. I imagine I could get around the issue mentioned in https://github.com/James-Yu/LaTeX-Workshop/pull/1452#issuecomment-503969142 by overwriting the paste shortcut defined in this PR, are create something like this — https://atom.io/packages/latex-image-paste. However, to me that very much defeats what seems to be the purpose of this extension; in that it serves as an _all in one_ latex environment.

tecosaur:
Re: pasting / drag&drop images, while clearly, it is not something you see much value in, taking a look at some other popular latex applications / extensions it seems that they either support it, or have had it requested.

 - Sublime Text's **LaTeX Tools** had image copy/paste requested and added here https://github.com/SublimeText/LaTeXTools/issues/1154
 - **LyX** supports copy/paste of images https://tex.stackexchange.com/a/58681/167605
 - **TeXstudio** had it requested here https://sourceforge.net/p/texstudio/feature-requests/1092/
 - **AUCTeX** had it added in this thread https://lists.gnu.org/archive/html/auctex/2010-12/msg00010.html
 - _(stopped looking)_

To me this suggests that there is relatively widespread interest in this functionality, and that the users of this extension would likely appreciate this capability.

tamuratak:
By "no one except you is positive", I mean "no authors and contributors of this extension except you is positive".

James-Yu:
Let's calm down and make this discussion constructive. It seems that currently the major dispute is on image-pasting. What about @tecosaur move the related code to a new branch for next-step-discussion, and make the paste table thing work well? The coding style may need an update as per @tamuratak suggeted, in particular

tecosaur:
@James-Yu that sounds like a good idea. I'll remove all of the contentious image related stuff, and then make a new branch for that. I'll also have a look at refactoring the `reformatText` bit.

tecosaur:
Right, now that's just table + text formatting. In a few hours I'll spin off a new branch for the image stuff and start a PR for discussing that seperately.

src/providers/completer/input.ts
```diff
@@ -40,12 +45,12 @@ export class Input {
...
         const content = utils.stripComments(fs.readFileSync(filePath, 'utf-8'), '%')
-        const regex = /\\graphicspath{(.*)}/g
-        let result: string[]|null
+        const regex = /\\graphicspath{(([^\{\}]|{.*})*)}/g
+        let result: string[] | null
         do {
             result = regex.exec(content)
             if (result) {
-                for (const dir of result[1].split(/\{|\}/).filter(s => s)) {
+                for (const dir of result[1].split(/\{|\}/).filter(s => s.replace(/^\s*$/, ''))) {
```
tamuratak: What is the purpose of this change?

tecosaur: For a multiline graphicspath,
```tex
\graphicspath{
    {somepath},
    {anotherpath}
}
```
matches which entirely consist of whitespace will be produced. This makes sure they are ignored.

package.json
```diff
@@ -1327,6 +1353,7 @@
     "mathjax": "2.7.5",
     "mathjax-node": "2.1.1",
     "micromatch": "^3.0.0",
+    "moment": "^2.24.0",
```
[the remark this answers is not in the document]
tecosaur: Nope. Removed in 3616244ba24a302b1980d0b5a059bbfb0e726616

package-lock.json
```diff
@@ -3159,6 +3159,11 @@
         }
       }
     },
+    "moment": {
```
tamuratak: Please remove this package also.

src/components/paster.ts
```diff
@@ -0,0 +1,230 @@
...
+            throw new Error('Table is not consistent')
+        }
+
+        const columnType: string = vscode.workspace.getConfiguration('latex-workshop.formattedPaste')['tableColumnType']
+        const booktabs: boolean = vscode.workspace.getConfiguration('latex-workshop.formattedPaste')['tableBooktabsStyle']
+        const headerRows: number = vscode.workspace.getConfiguration('latex-workshop.formattedPaste')['tableHeaderRows']
+        const tabularRows = cells.map(row => '\t' + row.join(' & '))
+
+        if (headerRows && tabularRows.length > headerRows) {
+            const headSep = booktabs ? '\t\\midrule\n' : '\t\\hline\n'
```
tamuratak: We should consider the case that the end of line character is `\r\n`.

tecosaur:
@James-Yu would you let me know if you like the change in aa0a77c28bed39b516657b6407b45e1f44041560

package-lock.json
tecosaur: Done in 7e2091b9ee648b876c20c7d49035ba51e455e7f5

src/components/paster.ts
tecosaur: Can't say I considered that — two questions.
 1. When would this be an issue?
 2. How would you know when to use which?

tamuratak: See `eol` property.
- https://code.visualstudio.com/api/references/vscode-api#TextDocument

tecosaur: Right. See f397f21c2c08df420f455073ccddc36890a45d3b

src/components/paster.ts
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    public pasteFile(editor: vscode.TextEditor, baseFile: string, file: string) {
+        const TABLE_FORMATS = ['.csv']
+        const extension = path.extname(file)
+
+        if (TABLE_FORMATS.indexOf(extension) !== -1) {
+            const contents = fs.readFileSync(path.resolve(baseFile, file), 'utf8')
+            if (extension === '.csv') {
+                // from: https://stackoverflow.com/a/41563966/3026698
```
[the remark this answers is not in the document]
tecosaur: Hmm. Looking around I think https://www.npmjs.com/package/detect-tabular looks like it could be good (seems like a thin wrapper around three other libraries) however I haven't yet been able to work out how to use it…

src/components/paster.ts
```diff
@@ -0,0 +1,220 @@
...
+            throw new Error('Table is not consistent')
+        }
+
+        const columnType: string = vscode.workspace.getConfiguration('latex-workshop.formattedPaste')['tableColumnType']
+        const booktabs: boolean = vscode.workspace.getConfiguration('latex-workshop.formattedPaste')['tableBooktabsStyle']
+        const headerRows: number = vscode.workspace.getConfiguration('latex-workshop.formattedPaste')['tableHeaderRows']
+        const tabularRows = cells.map(row => '\t' + row.join(' & '))
+
+        if (headerRows && tabularRows.length > headerRows) {
+            const eol = editor.document.eol.toString()
```
tamuratak: This is wrong. `eol` returns `number`, `1` or `2`. Please read the document carefully.

src/components/paster.ts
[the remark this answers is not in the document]
tamuratak: Why did you choose the package? Search the word, csv, not tabular.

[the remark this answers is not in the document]
tecosaur: Ahh. I think that looks nice because it supports csv, json, xlsx, and ods.
Perhaps just use https://www.npmjs.com/package/csv-parser now and potentially that later.

src/components/paster.ts
tecosaur: Ok. Sorry about that. Should be fixed in 1d8a6e36efa94817fcabc0c3908271f2bc66678d

src/components/paster.ts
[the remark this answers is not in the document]
tecosaur: Now using csv-parser in 6b2d5471fe67dc920c24c9a91dfb3155ecef2c21

tecosaur:
Unless anyone has any other comments I think this should be good now.

src/components/paster.ts
```diff
@@ -0,0 +1,201 @@
...
+        if (!editor) {
+            return
+        }
+
+        const fileUri = editor.document.uri
+        if (!fileUri) {
+            return
+        }
+        if (fileUri.scheme === 'untitled') {
+            vscode.window.showInformationMessage('Before paste image, you need to save current edit file first.')
```
jlelong: Suggested message instead : "You need to the save the current editor before pasting image."

src/components/paster.ts
```diff
@@ -0,0 +1,201 @@
...
+                        this.pasteTable(editor, body)
+                    })
+            }
+        }
+    }
+
+    public pasteTable(editor: vscode.TextEditor, content: string) {
+        this.extension.logger.addLogMessage('Pasting: Table')
+        // trim surrounding whitespace
+        content = content.replace(/^\s*/, '').replace(/\s*$/, '')
```
jlelong: Why not using `trim()`?

src/components/paster.ts
```diff
@@ -0,0 +1,201 @@
...
+        if (!isConsistent) {
+            throw new Error('Table is not consistent')
+        } else if (cells.length === 1 && cells[0].length === 1) {
+            this.pasteNormal(editor, content)
+            return
+        }
+
+        const columnType: string = vscode.workspace.getConfiguration('latex-workshop.formattedPaste')['tableColumnType']
+        const booktabs: boolean = vscode.workspace.getConfiguration('latex-workshop.formattedPaste')['tableBooktabsStyle']
+        const headerRows: number = vscode.workspace.getConfiguration('latex-workshop.formattedPaste')['tableHeaderRows']
```
jlelong: You should call `vscode.workspace.getConfiguration('latex-workshop.formattedPaste')` once.

src/components/paster.ts
jlelong: The variable `content` seems to be a multiline string. Should the trimming be applied to each line?

src/components/paster.ts
```diff
@@ -0,0 +1,201 @@
...
+                edit.replace(current, tabular)
+            }
+        })
+    }
+
+    public reformatText(text: string, removeBonusWhitespace = true) {
+        function doRemoveBonusWhitespace(str: string) {
+            str = str.replace(/\n\n/g, '\uE000')
+            str = str.replace(/\s+/g, ' ')
+            str = str.replace(/\uE000/g, '\n\n')
```
jlelong: What is the point of `str = str.replace(/\n\n/g, '\uE000')` and then `str = str.replace(/\uE000/g, '\n\n')`?

src/providers/completer/input.ts
```diff
@@ -19,18 +19,23 @@ export class Input {
...
         /* Check .gitignore if needed */
         if (vscode.workspace.getConfiguration('search', null).get('useIgnoreFiles')) {
             try {
-                gitIgnoredFiles = (cp.execSync('git check-ignore ' + files.join(' '), {cwd: baseDir})).toString().split('\n')
-            } catch (ex) { }
+                gitIgnoredFiles = cp
+                    .execSync('git check-ignore ' + files.join(' '), { cwd: baseDir })
+                    .toString()
+                    .split('\n')
+            } catch (ex) {}
```
jlelong: I do agree that the reformatting changes of this file improve readability and that they should be integrated at some point but I think you should refrain from introducing such modifications inside a new feature in order to keep a cleaner file history.

tecosaur:
Thanks for the reviews @jlelong I'll get right on it.

src/components/paster.ts
tecosaur: We don't actually need this restriction any more because most formatted pastes won't need the file to be saved to a location (images being an exception) so I've tweaked that in f29b86e391b2398b60d95276ecf85257fed2e3e1

src/components/paster.ts
tecosaur: Hmm. Thinking about this a bit more we want to trim all white space other than tabs (because blank cells at the start/end of a line are possible). I think 0fdfc7cd1f512d90da202b49742943fe9f90b644 should do the trick.

src/components/paster.ts
tecosaur: Fixed in 4bbcc93175dd610259dfe12a4d131b1a5bfabfb6.

src/components/paster.ts
tecosaur: Something I've noticed a fair bit when pasting from various sources is a lot of bogus newlines and spaces appear. White this doesn't affect the output it would be nicer to not have this happen. Sometimes pastes have special stuff such as weird spaces which can be a bit annoying. Hence I want to replace all white space with plain ol' breaking, singlewidth spaces.

The one exception to this is paragraphs, which are separated by a double newline. Hence I replace double new lines with a private use character (I think it's reasonable to expect this to not appear in any strings), and then turn those back into double newlines.

Thinking a bit more about this a few minor improvements could be made though. See a548e342bc4e6b6770f7aa42a5c39c84a6519338.

src/providers/completer/input.ts
tecosaur: Oops that was the linter + the `.tslint` file. de30d08bf929461b960d0569c4d99eef955608d7 should make that better.

tecosaur:
Ok that _should_ do it. @jlelong @tamuratak sorry for the sloppiness that you've had to point out. I appriciate you both taking the time to go through this PR and identify the dodgy bits. Hopefully it's good now :smile:.

tecosaur:
@James-Yu is this in a state which you'd be happy to merge yet?

werunom:
Thanks for this PR. I tried going through this in order to grasp the nature of this feature --- what it is, how to use, etc. But it is difficult to get these in this discussion. 

Can I please request a small note on this feature in the wiki? Thanks!

tecosaur:
@werunom Would you consider something like this to be a more useful entry in the changlog?

## [7.1.0] - 2019-07-07
### Added
 - Added formatted paste (#1452), `ctrl`+`shift`+`v`
   - Reformats some Unicode text for LaTeX
   - Table cells are turned into a `tabular`
   - Pasting the location of a `.csv` file pastes a table with the contents

_I could write a bit in the wiki, however I would like to wait until  #1478 is resolved (hopefully merged).
If it is, that would result in an entry like this to the changelog:_

### Added
 - Enhanced 'formatted paste' feature (see v7.1.0)
   - Can now paste an image from the clipboard (via `ctrl`+`v` _or_ `ctrl`+`shift`+`v`)
   - Can do a 'formatted paste' with an image file location to paste a filled-out template (like snippets), and can move/copy image to local folder / graphicspath

werunom:
@tecosaur - thanks! Yes...this is clearly and shows how I can use the new feature.

tecosaur:
@James-Yu do you think the changelog would be worth upsetting?

James-Yu:
Will do when the other PR is resolved.

Issue #1455, opened 2019-06-22
strange recipe failure 


musm:
```
    "latex-workshop.latex.recipes": [
      {
        "name": "latexmk ➞ copyPDF",
        "tools": ["latexmk", "copyPDF"]
      }
    ],
    "latex-workshop.latex.tools": [
    {
        "name": "latexmk",
        "command": "latexmk",
        "args": [
            "-synctex=1",
            "-interaction=nonstopmode",
            "-file-line-error",
            "-pdf",
            "-outdir=%OUTDIR%",
            "%DOC%"
        ],
        "env": {}
        },
        {
            "name": "copyPDF",
            "command": "cmd.exe",
            "args": [
                "copy /Y \"%OUTDIR%/%DOCFILE%.pdf\" \"%DIR%/\"",
            ],
            "env": {}
        },
    ]
```

```
[11:51:37] Recipe step 2: cmd.exe, copy /Y "c:/Users/Mus/test/out/rev.pdf" "c:/Users/Mus/test/
[11:51:37] LaTeX build process spawned. PID: 16680.
[11:51:37] Recipe returns with error: 1/null. PID: 16680.
```

Why does this fail?  Running the commands from cmd works fine just not  in VS Code recepie?

jlelong:
You have to split `copy /Y \"%OUTDIR%/%DOCFILE%.pdf\" \"%DIR%/\"` into an array of arguments otherwise it is considered as a single argument and gets extra quotes.

musm:
I'm pretty sure I've tried every permutation to no avail.
```
            "args": [
                "copy", "/Y",  "%OUTDIR%/%DOCFILE%.pdf",  "%DIR%",

            ],
```

```
[15:11:34] A step in recipe finished. PID: 2620.
[15:11:34] Recipe step 2: cmd.exe, copy,/Y,c:/Users/Mus/test/rev.pdf,c:/Users/Mus/test
[15:11:34] LaTeX build process spawned. PID: 16532.
[15:11:34] Recipe returns with error: 1/null. PID: 16532.

```

musm:
The update still doesn't output the std message
```
  "latex-workshop.latex.tools": [
    {
      "name": "latexmk",
      "command": "latexmk",
      "args": [
        "-synctex=1",
        "-interaction=nonstopmode",
        "-file-line-error",
        "-pdf",
        "-outdir=%OUTDIR%",
        "%DOC%"
      ],
      "env": {}
    },
[...]

PR #1458, opened 2019-06-22
Show stderr on recipe error

tamuratak:
This would be useful. For example, #1455.


musm:
could we get a release with this soon?

PR #1478, opened 2019-07-04
Move formatted paste to LaTeX Utilities, add api to get graphics path

tecosaur:
### Summary

 - Can now paste images from clipboard to file!!! (yay)
 - Adds shortcut `cmd`+`v`, just pastes normally unless blank (in which case it tries an image)
 - Clipboard to file accomplished via calling short os-specific scripts
 - This works but could maybe do with a bit of refinement

### Background

To me this is a _"Yes! Finally, this is fantastic!"_ type feature, however as seen by how heated #1452 became obviously not everyone shares my enthusiasm.

#### Simplification of the image inclusion process

Old Process

 1. Right click, save image to file
 2. Navigate to graphics folder
 3. Give file name
 4. Copy file name
 5. Go to latex file
 6. Trigger sniper for inserting figure / image
 7. Paste file name

New Process

 1. Right click, copy image
 2. Paste into vscode

#### Demand / Support in other editors

 - Sublime Text's **LaTeX Tools** had image copy/paste requested and added here https://github.com/SublimeText/LaTeXTools/issues/1154
 - **LyX** supports copy/paste of images https://tex.stackexchange.com/a/58681/167605
 - **TeXstudio** had it requested here https://sourceforge.net/p/texstudio/feature-requests/1092/
 - **AUCTeX** had it added in this thread https://lists.gnu.org/archive/html/auctex/2010-12/msg00010.html
 - _(stopped looking at this point)_

#### Structure of new feature

<a href="https://code2flow.com/9D5GTD"></a>

When pasting an image there are seven main methods called:
 - pasteImage
 - loadImageConfig
 - replacePathVariables
 - getImagePath
 - saveAndPaste
 - saveClipboardImageToFileAndGetPath
 - renderImagePaste

#### pasteImage
Just get's selection text and calls some of the other main functions.
As long as the other functions work this will to.
 - calling other class functions

 _Needs maintaining if: other class functions change_

#### loadImageConfig
Bog standard reading config variables from vscode settings and then applying `replacePathVariables`
 - vscode config reading

_Needs maintaining if: vscode api introduced drastic breaking changes_

#### replacePathVariables
Just performs a bunch of text replacements.
 - String manipulation
 - Requires `this.extension.completer.input.graphicsPath`

_Needs maintaining if: `this.extension.completer.input.graphicsPath` changes_

#### getImagePath
Uses selected text or config + user to determine image path.
- String / path manipulation.

_Needs maintaining if: `path` introduces drastic breaking changes_

#### saveAndPaste
Calls `ensureImgDirExists` which just uses `fse`, then either runs `fsCopy`/`fsRename` or `saveClipboardImageToFileAndGetPath`.
 - `fs`/`fse`
 - calling `saveClipboardImageToFileAndGetPath`

 _Needs maintaining if: `fs`/`fse` introduces drastic breaking changes_

#### saveClipboardImageToFileAndGetPath
Uses `spawn` from childprocess, calls short os-specific scripts.
Scripts are short and simple, should be quite stable.

_Needs maintaining if: childprocess depreciates `spawn` or clipboard interaction method with os drastically changes (unlikely)_

#### renderImagePaste
Uses config from `loadImageConfig` then does text replacement.

_Needs maintaining if: text replacement config scheme changes_

### Maintainability conclusion

Shouldn't require much.


tamuratak:
I am against this PR as I explained in the [previous PR](https://github.com/James-Yu/LaTeX-Workshop/pull/1452). I do not want to continue the discussion as I have explained enough. I think this is up to the authors, @James-Yu @jlelong, now.

tecosaur:
@James-Yu @jlelong it would be good if you could give some feedback on this PR.

James-Yu:
I may need a bit of time to go through it. Please remind me if I did not show up in a few days.

tecosaur:
@James-Yu *cough

package.json
```diff
@@ -293,6 +293,11 @@
         "title": "Toggle follow cursor",
         "category": "LaTeX Workshop"
       },
+      {
+        "command": "latex-workshop.paste",
```
James-Yu: Can we merge the two pastes together as one command? I don't think overwriting `ctrl+v` shortcut is a good idea. We came across similar things previously (the enter feature), and a huge number of side-effects need to be resolved.

package.json
```diff
@@ -531,6 +536,12 @@
         "key": "alt+enter",
         "when": "editorTextFocus && !editorReadonly && editorLangId == 'latex' && !suggestWidgetVisible"
       },
+      {
+        "key": "ctrl+v",
```
James-Yu: See above.

package.json
```diff
@@ -1281,6 +1292,49 @@
           "type": "integer",
           "default": 1,
           "description": "Number of header rows to assume. Set to 0 to disable.`"
+        },
+        "latex-workshop.formattedPaste.image.method": {
```
James-Yu: I am strongly concerned about moving the image files. I suppose we can simplify the workflow. If it is an image file, reference its absolute path. If it's a clipboard image, save in default folders as below. My 2c.

package.json
```diff
@@ -1281,6 +1292,49 @@
...
+          ],
+          "enumDescriptions": [
+            "reference the image where it is",
+            "copy image to target location",
+            "move image to target location"
+          ],
+          "default": "leave",
+          "description": "When pasting the path to an image, determine how it is treated. With copy and move 'target location' is graphicsdir or cwd"
+        },
+        "latex-workshop.formattedPaste.image.template": {
```
James-Yu: Are there other possible templates than this one?

package.json
```diff
@@ -1281,6 +1292,49 @@
...
+          "type": [
+            "string",
+            "array"
+          ],
+          "default": "\\includegraphics{${imageFilePath}}",
+          "description": "The template to use for pasting the image (use an array for multiline strings), use ${imageFilePath} to represent the image file. Can also use ${imageFileNameWithoutExt}"
+        },
+        "latex-workshop.formattedPaste.image.defaultName": {
```
James-Yu: Just make it the default and remove the config. Also, I'm a bit worried about the size of moment, especially when we can play with `new Date()` for date and `new Date().toLocaleTimeString()` for time.

package.json
```diff
@@ -1281,6 +1292,49 @@
...
+          "type": "string",
+          "default": "Y-MM-DD-HH-mm-ss",
+          "description": "The default copied image file name. The value of this config will be pass to the 'format' function of moment library (a js time manipulation library), you can read document https://momentjs.com/docs/#/displaying/format/ for advanced usage. You can include unformatted text by enclosing with square brackets [like so]."
+        },
+        "latex-workshop.formattedPaste.image.filePathConfirmInputBoxMode": {
```
James-Yu: Also remove this one and show relative path related to the root of project on clipboard case.

package.json
```diff
@@ -1359,6 +1413,7 @@
     "mathjax": "2.7.5",
     "mathjax-node": "2.1.1",
     "micromatch": "^3.0.0",
+    "moment": "^2.24.0",
```
James-Yu: See above.

src/main.ts
```diff
@@ -247,7 +247,8 @@ export async function activate(context: vscode.ExtensionContext) {
 
     vscode.commands.registerCommand('latex-workshop.showCompilationPanel', () => extension.buildInfo.showPanel())
 
-    context.subscriptions.push(vscode.commands.registerCommand('latex-workshop.formattedPaste', () => extension.paster.paste()))
+    context.subscriptions.push(vscode.commands.registerCommand('latex-workshop.paste', () => extension.paster.paste()))
```
James-Yu: merge

src/components/paster.ts
```diff
@@ -207,4 +252,390 @@ export class Paster {
 
         return text
     }
+
```
James-Yu: In general, too many codes on path manipulation and file copying/moving. The whole workflow should be simplified, and the extension should have as little footprint on files as possible. In my opinion, file on disk -> use the path. Not on disk -> save to default with some tweaks for users.

package.json
tecosaur: Yes. For instance what I will probably set that to is:
```json
        [
            "\\begin{figure}[!htb]",
            "\t\\centering",
            "\t\\includegraphics[]{${imageFilePath}}",
            "\t\\caption{${imageFilePathWithoutExt}}",
            "\t\\label{fig:${imageFilePathWithoutExt}}",
            "\\end{figure}"
        ]
```

tecosaur:
Thanks @James-Yu I'll go through all of those either tonight or tomorrow :)

package.json
tecosaur: Didn't think about that enter issue. Done in 4af193574c0483b73c57ede6387cd92e092bbecf

package.json
tecosaur: See 4af193574c0483b73c57ede6387cd92e092bbecf

src/main.ts
tecosaur: Did in 4af193574c0483b73c57ede6387cd92e092bbecf

package.json
tecosaur: Seems reasonable. I'll just hardcode that to `"leave"`. The one extra bit I'll leave is if text is selected when pasting — then the image will just be copied over. The only added complexity due to this is the following three lines.

```ts
import { promisify } from 'util'
const fsCopy = promisify(fs.copyFile)
...
fsCopy(oldPath, imagePath)
```
Done in e59e6ed02fa660a75e6f300790018c7c5bb65b66

package.json
tecosaur: I'll mark this resolved for now but let me know if you want me to do anything with this.

package.json
tecosaur: Seems reasonable. I'll set the name to `image<number>`. Done in 51c026ab907634c1401d7fe18a1de974f5c1bc90

package.json
tecosaur: See 51c026ab907634c1401d7fe18a1de974f5c1bc90

package.json
tecosaur: See 51c026ab907634c1401d7fe18a1de974f5c1bc90

package.json
tecosaur: Ooops, wrong comment. This is now done in d775384bf69c0cb0ac049c3f7e12a94ffbd6d2ae

src/components/paster.ts
tecosaur: Not sure what to do to further reduce path manipulation, I think with the changes I've just made the other half of your comment is resolved.

tecosaur:
@James-Yu let me know if that's better now, and if you'd like any more changes.

tecosaur:
@James-Yu *cough

jlelong:
As @tamuratak, I am still not convinced by the _Paste from clipboard_ function. This is true that this feature has been requested in other _LaTeX_ extensions but it is not clear at that it has been implemented. 

 > Sublime Text's LaTeX Tools had image copy/paste requested and added here SublimeText/LaTeXTools#1154

Not clear that pasting from clipboard has indeed been implemented but rather only pasting a file path.

 > TeXstudio had it requested here https://sourceforge.net/p/texstudio/feature-requests/1092/

But not implemented

 > AUCTeX had it added in this thread https://lists.gnu.org/archive/html/auctex/2010-12/msg00010.html

It does not seem to have been integrated into AUCTeX.

Anyway, if you believe pasting from clipboard is important, it goes beyond LaTeX and there are indeed several vscode extensions to achieve this. In particular, it looks like [https://github.com/mushanshitiancai/vscode-paste-image](mushanshitiancai/vscode-paste-image) could do it almost out of the box.

tecosaur:
@jlelong Yea, I have mixed feelings on this too; _particularly_ given that https://github.com/mushanshitiancai/vscode-paste-image exists. Ultimately what made me think it's worth going in this direction is that it fits with my view of this extension as a "one stop shop" for the functionality you'd want with LaTeX documents.

> It does not seem to have been integrated into AUCTeX.

Re: AUCTeX for it being integrated you need to skip to https://lists.gnu.org/archive/html/auctex/2010-12/msg00015.html where I saw "I integrated it into auctex/tex.el and it is working nicely" and then the discussion continues in further threads.

tamuratak:
@tecosaur stop such an argument. With your argument, we must include every function requested by users. Nonsense. We must decide which one to implement.

tecosaur:
I usually see adding functionality requested by users to be a good thing :stuck_out_tongue: but yes — definitely can be a slippery slope in some cases. Anyway, I don't think there's much point in me discussing this further so I'll just wait untill James decides to weigh in.

jlelong:
Aiming at building the All-In-One extension for LaTeX may sound a promising idea but it may turn into a trap because it will lead the extension to grow far too much. We must always weigh the benefit for most users of introducing a feature versus the maintainability issues and potential bugs. 

Concerning this PR, I do believe that it would better fit directly into [https://github.com/mushanshitiancai/vscode-paste-image](https://github.com/mushanshitiancai/vscode-paste-image) as making the LaTeX support would only be a small add-on. Modularity is always good practice in software design. Before going any further, I would suggest @tecosaur to get in touch with the maintainers of [https://github.com/mushanshitiancai/vscode-paste-image](https://github.com/mushanshitiancai/vscode-paste-image).

tecosaur:
@jlelong I think that sounds like a good way of handling this. Hence, this is my revised plan
 - head over to https://github.com/mushanshitiancai/vscode-paste-image and get in touch
 - make PR there (hopefully)
 - add small integration in this extension so that image pasting can happen with formatted paste _something like:_
```ts
// almost-code
if (clipboard.text === '') {
    try {
        vscodeExecuteCommand('vscode-paste-image.paste')
    } catch (e) {
        vscodeShowMessage('Install vscode-paste-image to be able to paste images')
    }
}
```

tecosaur:
Addendum. The one bit I wouldn't be sure how to do would be passing the reference for `graphicsPath` over. Would anyone have any suggestions for that?

James-Yu:
While I am interested in seeing this feature, I totally agree with @jlelong on modularity and maintainability. Let's see if the other extension would incorporate the main logic so that we only need to have a minimal adapter.

tecosaur:
I have created https://github.com/mushanshitiancai/vscode-paste-image/issues/50 and I'll report back when something's happening :smiley:

tecosaur:
Ok. It's now been five days and I haven't seen any sign of a response. Looking at other open issues, the last few haven't got any response either (Going back to early May). Hmmm, I'm not sure how to proceed from here. @James-Yu any thoughts?

tamuratak:
@tecosaur make your own extension.

tecosaur:
Done. This now just contains an adaptor + some formatting changes that slipped in.
@James-Yu how would you feel about merging this now?

tamuratak:
Why do we need this change on LaTeX Workshop's side? You misunderstand James' idea https://github.com/James-Yu/LaTeX-Workshop/pull/1544#issuecomment-515844860.  LaTeX Utilities uses some API exposed by LaTeX Workshop. Not the opposite. If changes on LaTeX Workshop's side are necessary on every new feature by LaTeX Utilities, the maintainability of LaTeX Workshop does not get better.

tecosaur:
@tamuratak Other than the API revealing on this extension's side, changes are _not required_ for this feature. The feature works with just LaTeX Utilities using the "Paste an Image File" command. I do not see this as indicative of how other feature will be either, take the "Live Reformatting" — that is also implemented in LaTeX Utilities and requires no change this side.

**However**, it makes logical sense that when you paste an image path the same thing happens as with CSVs etc. and if you have an image in the clipboard "formattedPaste" would logically be the command to insert it into the document, and to have the user remember yet another shortcut seems wholly unnecessary. I also fail to see any real maintinence cost induced by effectively adding four lines twice.
```ts
if (this.extension.utilities) {
    this.extension.utilities.exports.pasteImage()
} else {
    vscode.window.showInformationMessage('Install LaTeX Utilities to paste images')
}
```
I must say, to me it seems a bit ridiculous to be concerned about the 'maintinence cost' of that.

In James's comment he talkes about modularising features. This has been done. I see now reason why we can't add a very small number of simple lines to make things work better though.

tamuratak:
When another person with a fancy idea "Paste X" appears, they will want another change on LaTeX Workshop's side. You have to discuss with each other which "Paste X/Image" to prioritize in LaTeX Workshop's side. That is a kind of increase in the maintenance cost of LaTeX Workshop. We can avoid such a thing with an appropriate API.

Please think about what kind of API exposed by LaTeX Workshop will solve your problem.

tecosaur:
Thank you for explaining that @tamuratak. I was thinking of it as just this single feature.
Perhaps if we want to fully follow the idea mentioned earlier (in the live reformatter PR)

> - LaTeX Workshop; gives you all the basics to get started with LaTeX in vscode without too much hassle
> - LaTeX Untilities; all collection fancy ease-of-life features that the majority of users may not use, but a significant number of people will be very interested in

Then the best place for the "formattedPaste" feature itself woud be in `LaTeX Utilities`? It would definately make it much cleaner to add future pasting capability in the Utilities extension if the formatted paste feature is there.

What do you think?

James-Yu:
I would vote yes for moving all formatted paste features to utilities. That makes it more clear.

tecosaur:
Ok. I think that should resolve all of this rather nicely. I'll move it tonight.

James-Yu:
Besides this, we may also think of what existing features may be better re-positioned in LaTeX Utilities. To name one, the word counter feature is way too isolated and may be good as a module in LU.

tecosaur:
To me (regarding which features should be where) I think the first question that should be asked is the approach. To me, there are three main ways we can go about labelling issues as Workshop or Utilities.

1. Identify all the 'core' / essential features. All other features are labelled extra and should be moved to Utilities
2. Identify all 'extra' features. These features are moved to Utilities. All other features are labelled 'core'.
3. List all features. Go through and label as 'core' or 'extra', move extra to Utilities.

James-Yu:
I am implicitly taking the third approach. Will have more PRs to LU in a few days.

tamuratak:
This PR includes too much unrelated changes. Please create a new PR of moving formattedPaste to LU for an accurate history.

tecosaur:
I think I a new issue should be used for the discussions, I can't imagine that this is the end. Let's move to #1558

Issue #1486, opened 2019-07-08
Should stop using webpack


tamuratak:
We now are using [webpack](https://github.com/James-Yu/LaTeX-Workshop/blob/master/package.json#L1346) to minimize the size of javascript files. It compiles all the files into `main.js`. That makes log messages useless and debugging extremely difficult. For example, see the console message in #1472.

webpack introduced in #864 to reduce the startup time of this extension. However, we can achieve that without webpack by using lazy loading.

James-Yu:
It seems that webpack is still [recommended by MS](https://code.visualstudio.com/api/working-with-extensions/bundling-extension).

tamuratak:
We have `**/*.map` in `.vscodeignore`. If we remove the line and include the source-map files in the release, will the console display the original files? To my testing, the answer is no. The VS Code team does not support the source map files  from webpack. Possible related to microsoft/vscode/issues/76659. Although they recommend using webpack, they seem not to support the source map of webpack. That is contradictory.

Anyway, we have no test.  Testing is also [recommended](https://code.visualstudio.com/api/working-with-extensions/testing-extension).  Nevertheless, no one single test. So, error messages are the only clue to debug. Removing them from logs is not a smart decision.

jlelong:
From [https://code.visualstudio.com/docs/nodejs/nodejs-debugging](https://code.visualstudio.com/docs/nodejs/nodejs-debugging)

> Are you using Webpack? By default, it outputs paths with a `webpack:///` prefix, which the debug adapter can't resolve. You can change this in your Webpack configuration with the `devtoolModuleFilenameTemplate` option, or try using the 'inspector' protocol, which provides some extra options for resolving these paths.

Maybe that helps.

Testing would indeed be a tremendous improve for maintaining the extension but we definitely do not have the manpower to implement it. We can barely fix issues. I agree with @tamuratak that error messages are of prime importance to analyse bugs.

PR #1492, opened 2019-07-09
stop using webpack.

tamuratak:
Related to #1486.


James-Yu:
What do you think about this change? @jlelong

Issue #1445, opened 2019-06-17
Incomplete Support for Subfiles Package in Combination with Docker Integration


sbaeurle:
**Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]**

You still see this issue?: Yes

**Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.**

**Describe the bug [Required]**
Using the subfiles package in combination with the docker integration is not working as expected.
When compiling a subfile in any subdirectory the compilation fails due to incorrect volume binding.
Latex Workshop is mounting the subdirectory into the data folder of the docker container, thus preventing access to the needed root tex.

**To Reproduce**
1. Create a Latex Document with subfiles package
```
\usepackage{subfiles}
\documentclass{article} 
\begin{document}
\subfile{content/introduction}
\end{document}
```
2. Create subfile in subdirectory (e.g. content/introdution.tex)
```
\documentclass[../document]{subfiles}
\begin{document}
\chapter{Introduction}
Hello!
\end{document}
```
3. Compilation will fail

**Expected behavior**
Latex Workshop starts Docker with the correct root directory mounted into the /data folder inside the container.

**Logs [Required]**
```
_irrelevant_\content>docker run -i --rm -w /data -v "_irrelevant_\content:/data" tianon/latex latexmk -synctex=1 -interaction=nonstopmode -file-line-error -pdf -outdir=./ introduction 
Latexmk: This is Latexmk, John Collins, 1 January 2015, version: 4.41.
Rule 'pdflatex': File changes, etc:
   Non-existent destination files:
      'introduction.pdf'
------------
Run number 1 of rule 'pdflatex'
------------
------------
Running 'pdflatex  -synctex=1 -interaction=nonstopmode -file-line-error -recorder  "introduction.tex"'
------------
Latexmk: applying rule 'pdflatex'...
This is pdfTeX, Version 3.14159265-2.6-1.40.17 (TeX Live 2016/Debian) (preloaded format=pdflatex)
 restricted \write18 enabled.
entering extended mode
(./introduction.tex
LaTeX2e <2017/01/01> patch level 3
Babel <3.9r> and hyphenation patterns for 83 language(s) loaded.
(/usr/share/texlive/texmf-dist/tex/latex/subfiles/subfiles.cls
Document Class: subfiles 2012/05/23 Federico Garcia
[...]

Issue #1443, opened 2019-06-17
intellisense is not coherent when using subfiles


werunom:
**Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]**

You still see this issue?: Yes

**Describe the bug [Required]**

I have the following setup:
`config-file` --- has all the tex configurations (like packages, citation style, etc.)
`tex-file` --- I import the `config-file` in each tex file. 

In this configuration, I observed that there is a difference in intellisense suggestions seen in `config-file` and `tex-file`. 

**To Reproduce**
Type latex commands and see what autosuggestions are given

**Logs [Required]**
Logs might not be required

**Screenshots**
Autocomplete suggestions in `Config-file`
![main-file-cropped](https://user-images.githubusercontent.com/24664855/59584910-e3d42400-90fc-11e9-9ff7-c2e0192813a9.png)

Autocomplete suggestions in `Tex-file`
![sub-file-cropped](https://user-images.githubusercontent.com/24664855/59584923-e9316e80-90fc-11e9-8ac2-429ed8d0ffe0.png)

As you can see, calling `textcite` in `tex-file` does not give all the various autofill options available in the `config-file`. And some other commands (like `autocite`) are not even suggested in the `sub-file`.

**Desktop [Required]:**
 - OS: Linux manjaro
 - VS Code version: 1.35.1
 - Extension version: 7.0.2

**Additional context**

This bug might be related to [this issue](https://github.com/James-Yu/LaTeX-Workshop/issues/1438).

jlelong:
This is the same root cause as #1438. Intellissense depends on the packages used and subfiles do not include any package. Hence, the problem.
I suggest to continue the discussion in #1438.

Issue #1438, opened 2019-06-15
Inline citation doesn't popup in subfiles


jrwrigh:
Foreword: I originally wrote the issue up as a bug report because I was trying to track down a bug. I have now found a "solution" to the bug, but said solution is quite inconvenient. 

**Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]**

You still see this issue?: Yes

**Describe the bug [Required]**
Intellisense does not pop-up with citation key suggestions in one of my LaTeX projects. Specifically, a project that uses subfiles. See the "Additional Context" section below for more details.
![CitationExampleWrongThesis](https://user-images.githubusercontent.com/20801821/59554326-5ca57580-8f6f-11e9-8145-7ad40c24f0f8.gif)

**To Reproduce**
1. Go to existing project
2. Start typing in "\cite{"
3. Hit 'Enter' to autocomplete and bring up the intellisense "suggestions" (idk what to call it)
4. Start typing in citation key, but it does not come up

**Expected behavior**
Intellisense to show the citation keys and insert them. This is done on a single files LaTeX project using the same `.bib` file.
![CitationExampleCorrectAIAA](https://user-images.githubusercontent.com/20801821/59554337-7b0b7100-8f6f-11e9-9920-ffc85b7052fe.gif)

**Logs [Required]**
This log is after reloading the VS Code window with nothing but LaTeX Workshop extension loaded, and immediately performing the reproduction steps.
<details>
<summary>LaTeX Workshop Output [Required]</summary>
[13:15:32] Initializing LaTeX Workshop.
[13:15:32] pdflatex is provided by MiKTeX
[13:15:32] Creating LaTeX Workshop http and websocket server.
[13:15:32] LaTeX Workshop initialized.
[13:15:32] Linter for root file started.
[13:15:32] Found root file from active editor: c:\Users\jrwrigh\gitRepos\ms_thesis\background.tex
[13:15:36] Root file changed from: undefined. Find all dependencies.
[13:15:36] Instantiating a new file watcher for c:\Users\jrwrigh\gitRepos\ms_thesis\background.tex
[13:15:36] Parsing c:\Users\jrwrigh\gitRepos\ms_thesis\background.tex
[13:15:36] Cannot find file c:\Users\jrwrigh\gitRepos\ms_thesis\background.fls
[13:15:36] Server created on 127.0.0.1:62975
[13:15:36] Snippet data loaded.
[...]

PR #1475, opened 2019-07-04
Format all the things

tecosaur:
The formatting in this project seems a tad inconsistent.

So, for your perusal here's a PR which is the result of running ~300 files in this project through tslint + prettier. Quite a few of the modified json files just have an extra newline at the end.

Formatting config:
 - Max line length: 120
 - spaces per indent: 4 (ts), and 2 (json)

All other settings are defaults / using this project's tslint.

**I don't expect this to be merged, but do let me know what you think.**

They're also some tslint rules that can be set to help set a more consistent coding style. I've added these rules.

 - [no-null-keyword](https://palantir.github.io/tslint/rules/no-null-keyword/)
 - [no-return-await](https://palantir.github.io/tslint/rules/no-return-await/)
 - [no-string-literal](https://palantir.github.io/tslint/rules/no-string-literal/)
 - [no-string-throw](https://palantir.github.io/tslint/rules/no-string-throw/)
 - [prefer-object-spread](https://palantir.github.io/tslint/rules/prefer-object-spread/)
 - [static-this](https://palantir.github.io/tslint/rules/static-this/)
 - [arrow-return-shorthand](https://palantir.github.io/tslint/rules/arrow-return-shorthand/)
 - [interface-over-type-literal](https://palantir.github.io/tslint/rules/interface-over-type-literal/)
 - [no-boolean-literal-compare](https://palantir.github.io/tslint/rules/no-boolean-literal-compare/)
 - [no-unnecessary-qualifier](https://palantir.github.io/tslint/rules/no-unnecessary-qualifier/)
 - [object-literal-key-quotes](https://palantir.github.io/tslint/rules/object-literal-key-quotes/)
 - [ordered-imports](https://palantir.github.io/tslint/rules/ordered-imports/)
 - [prefer-method-signature](https://palantir.github.io/tslint/rules/prefer-method-signature/)
 - [prefer-while](https://palantir.github.io/tslint/rules/prefer-while/)
 - [space-before-function-paren](https://palantir.github.io/tslint/rules/space-before-function-paren/)
 - [space-within-parens](https://palantir.github.io/tslint/rules/space-within-parens/)
 - [arrow-parens](https://palantir.github.io/tslint/rules/arrow-parens/)
 - [linebreak-style](https://palantir.github.io/tslint/rules/linebreak-style/)
 - [newline-before-return](https://palantir.github.io/tslint/rules/newline-before-return/)
 - [number-literal-format](https://palantir.github.io/tslint/rules/number-literal-format/)
[...]

PR #1496, opened 2019-07-10
New subfiles package support

jlelong:
This PR aims at solving 
- #1438, #1443: completion issues
- #1445: wrong root directory.

The `subfiles` package enables either to compile the main file or any subfile containing `\documentclass[main.tex]{subfiles}`, which breaks the automatic detection of the `rootFile`. So far, the `rootFile` detection was not consistent
- If the main file is not opened, then the `rootFile` is the subfile. This creates issues #1438, #1443, #1445
- If the main file is opened, then the `rootFile` is the main file and switching the subfile editor does not change anything.

The new approach of this PR.
- The internal `rootFile` is **always** defined as the main file, which solves all the issues listed above.
- All the interactive commands `build, `clean` and `view` use a quick pick box to ask the user which file is to be considered as the root File.
- The non-interactive functions `autobuild`, `autoclean` and forward `synctex` rely on the value of the new configuration variable `latex-workshop.latex.rootFile.useSubFile` to choose between the main file and the subfile. When set to `true` (its default value), the subfile is considered.
- The `synctex` after build automatically picks the right root file from the `build` function.

This is a major refactoring and change of the `build`, `clean`, `synctex` and `view` tools. On top of my tests, I would like other people to test this before merging, especially because I do not use `subfiles` that much.

### DIFF
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -238,7 +238,7 @@ export class Builder {
         })
 
         this.currentProcess.on('exit', (exitCode, signal) => {
-            this.extension.parser.parse(stdout)
+            this.extension.parser.parse(stdout, rootFile)
             if (exitCode !== 0) {
                 this.extension.logger.addLogMessage(`Recipe returns with error: ${exitCode}/${signal}. PID: ${pid}. message: ${stderr}.`)
                 this.extension.buildInfo.buildEnded()
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -250,7 +250,7 @@ export class Builder {
                         this.extension.logger.displayStatus('x', 'errorForeground', `Recipe terminated with error. Retry building the project.`, 'warning')
                         this.extension.logger.addLogMessage(`Cleaning auxillary files and retrying build after toolchain error.`)
 
-                        this.extension.commander.clean().then(() => {
+                        this.extension.cleaner.clean(rootFile).then(() => {
                             this.buildStep(rootFile, steps, 0, recipeName, releaseBuildMutex)
                         })
                     } else {
@@ -261,7 +261,7 @@ export class Builder {
                 } else {
                     this.extension.logger.displayStatus('x', 'errorForeground')
                     if (['onFailed', 'onBuilt'].indexOf(configuration.get('latex.autoClean.run') as string) > -1) {
-                        this.extension.commander.clean()
+                        this.extension.cleaner.clean(rootFile)
                     }
                     const res = this.extension.logger.showErrorMessage(`Recipe terminated with error.`, 'Open compiler log')
                     if (res) {
@@ -312,7 +312,7 @@ export class Builder {
         // }
         if (configuration.get('latex.autoClean.run') as string === 'onBuilt') {
             this.extension.logger.addLogMessage('Auto Clean invoked.')
-            this.extension.cleaner.clean()
+            this.extension.cleaner.clean(rootFile)
         }
     }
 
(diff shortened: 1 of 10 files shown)


src/components/server.ts
```diff
@@ -13,6 +13,7 @@ export class Server {
     wsServer: ws.Server
     address: string
     port: number
+    pdfFile: string | undefined = undefined
```
tamuratak: What is the purpose of this property? Where do you use it?

jlelong: Unused stuff I forgot to remove after a change of idea...

src/components/server.ts
```diff
@@ -54,6 +54,7 @@ export class Server {
         if (request.url.indexOf('pdf:') >= 0 && request.url.indexOf('viewer.html') < 0) {
             // The second backslash was encoded as %2F, and the first one is prepended by request
             const fileName = decodeURIComponent(request.url.replace('/pdf:', ''))
+            this.pdfFile = fileName
```
trevorgunn: You need to remove this line as well as line 77 or else the compiler complains.

src/components/builder.ts
```diff
@@ -306,13 +306,13 @@ export class Builder {
         this.extension.completer.reference.setNumbersFromAuxFile(rootFile)
         this.extension.manager.findAdditionalDependentFilesFromFls(rootFile)
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
-        if (configuration.get('synctex.afterBuild.enabled') as boolean) {
-            this.extension.logger.addLogMessage('SyncTex after build invoked.')
-            this.extension.locator.syncTeX()
-        }
+        // if (configuration.get('synctex.afterBuild.enabled') as boolean) {
+        //     this.extension.logger.addLogMessage('SyncTex after build invoked.')
```
trevorgunn: Should we copy this line to the corresponding place in `viewer.ts`?

src/components/server.ts
jlelong: My bad!

src/components/builder.ts
jlelong: Yes but it is not really synctex after build. It gets called every time the viewer is refreshed so even when we first open it. Anyway this might be worth putting a log message.

src/components/manager.ts
```diff
@@ -11,6 +11,7 @@ import {Extension} from '../main'
 export class Manager {
     extension: Extension
     rootFiles: object
+    localRootFiles: object
```
tamuratak: Should be typed as `{ [key: string]: string | undefined }`.

src/components/manager.ts
```diff
@@ -11,6 +11,7 @@ import {Extension} from '../main'
 export class Manager {
     extension: Extension
     rootFiles: object
```
tamuratak: Should be typed as `{ [key: string]: string }`

src/components/locator.ts
```diff
@@ -93,7 +93,7 @@ export class Locator {
         }
     }
 
-    syncTeX(args?: {line: number, filePath: string}, forcedViewer: string = 'auto') {
+    syncTeX(args?: {line: number, filePath: string}, forcedViewer: string = 'auto', pdfFile?: string) {
```
tamuratak: This function is too long. It needs refactoring.

jlelong: I agree that the `syncTeX` function is quite lengthy but its logic is quite clear and many others are also lengthy.
I have not changed much in this function actually. So I think I am going to leave it like it.

Some day, we should take some time to do a bit a refactoring.

tecosaur: (related to jlelong's comment: #1475)

Issue #1472, opened 2019-07-04
vscode 1.36 breaks some features when used with remote-ssh


Blaok:
**Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]**

You still see this issue?: Yes -- well I have to keep the remote-ssh extensions

**Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.**

**Describe the bug [Required]**
With vscode 1.36 (June 2019) and remote-ssh, some features of this extension are no longer working, including but not limited to:
+ built-in tab synctex
+ default trim
+ default zoom level

**To Reproduce**
1. install vscode 1.36 and the remote-ssh extensions
2. login to a remote server, install this extension
3. open a latex folder, build and preview the pdf, ctrl-click on the pdf, not sync back to source
4. change default trim and / or zoom level, open preview again, those settings are not honored

**Expected behavior**
Ctrl-click on the pdf should sync back to the source. Settings of trim and zoom level should be honored.

**Logs [Required]**
Please paste the whole log messages here, not parts of ones. It is very important to identify problems. If you think the logs are unrelated, please say so.

I don't think these logs are related. I deleted part of the path for privacy.

<details>
<summary>LaTeX Workshop Output [Required]</summary>

```
[17:42:37] Initializing LaTeX Workshop.
[17:42:37] Creating LaTeX Workshop http and websocket server.
[17:42:37] LaTeX Workshop initialized.
[17:42:37] Found root file from active editor: main.tex
[17:42:40] Root file changed from: undefined. Find all dependencies.
[17:42:40] Instantiating a new file watcher for main.tex
[17:42:40] Parsing main.tex
[17:42:40] Found .bib file library.bib
[17:42:40] Creating file watcher for .bib files.
[17:42:40] Adding .bib file library.bib to bib file watcher.
[17:42:40] Parsing .bib entries from library.bib
[17:42:40] Parsed 209 .bib entries from library.bib.
[17:42:40] Parsing main.fls to compute dependencies
[17:42:40] Found .bib file library.bib
[17:42:40] .bib file library.bib is already being watched.
[17:42:40] Server created on 127.0.0.1:35196
[17:42:40] Snippet data loaded.
[17:42:40] LaTeX Workshop version: 7.0.2
[17:42:49] VIEW command invoked.
[17:42:49] Found root file from active editor: main.tex
[...]

PR #1501, opened 2019-07-11
Use base64url to encode path

tamuratak:
We use `encodeURIComponent` to encode the path of a pdf file for the request of `src/components/server.ts`. We also use `/pdf:` as the prefix of the pdf file. They are affected by the change of the behavior of VS Code and pdf.js. For example, #1472 and d68195f5484fd206e21bd507653cb607315025ab.

To solve this issue, we should use base64url to encode the path. See [link](https://en.wikipedia.org/wiki/Base64#URL_applications). Only the characters `a-zA-Z0-9`, `_`, and `-` are used to encode it on the scheme. They are guaranteed not to be converted by `encodeURIComponent`.

In this PR, after calling `encodeURIComponent`, we encode the result with base64url.

### DIFF
--- a/src/components/encodePath.ts
+++ b/src/components/encodePath.ts
@@ -1,0 +1,15 @@
+export const pdfFilePrefix = 'pdf..'
+
+export function encodePath(path: string) {
+    const s = encodeURIComponent(path)
+    const b64 = Buffer.from(s).toString('base64')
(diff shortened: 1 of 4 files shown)


James-Yu:
I think the idea is quite good! Any comments? @jlelong

jlelong:
That looks a good idea. Let us give it a try!

Issue #1428, opened 2019-06-10
Mathjax does not render correctly


joulev:
**Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]**
Done.

**Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.**
Done.

**Describe the bug [Required]**
The Mathjax preview  is cropped (even when the equation is short).

**To Reproduce**
This is my example code
```latex
\documentclass{article}
\usepackage{needspace}
\newcommand{\topic}[1]{%
    \needspace{5\baselineskip}%
    \begin{center} {\itshape #1} \end{center}%
}
\begin{document}
Hello world! This is a famous equation: $E=mc^2$.
\end{document}
```
After putting the mouse on `$E=mc^2$`, I see something like the screenshot below.

**Expected behavior**
Of course, it should be *E* = *mc*<sup>2</sup>, not = *mc*<sup>2</sup>

**Logs [Required]**

<details>
<summary>LaTeX Workshop Output [Required]</summary>

```
[09:33:54] Initializing LaTeX Workshop.
[09:33:56] pdflatex is provided by MiKTeX
[09:33:56] Creating LaTeX Workshop http and websocket server.
[09:33:56] LaTeX Workshop initialized.
[09:33:56] Found root file from active editor: e:\LaTeX Projects\check\check.tex
[09:34:00] Root file changed from: undefined. Find all dependencies.
[09:34:00] Instantiating a new file watcher for e:\LaTeX Projects\check\check.tex
[09:34:00] Parsing e:\LaTeX Projects\check\check.tex
[09:34:00] Cannot find file e:\LaTeX Projects\check\check.fls
[09:34:00] Server created on 127.0.0.1:52665
[09:34:00] Snippet data loaded.
[09:34:00] LaTeX Workshop version: 7.0.2
[09:35:33] Parsing e:\LaTeX Projects\check\check.tex
[09:35:33] Cannot find file e:\LaTeX Projects\check\check.fls
[09:35:33] File watcher: responding to change in e:\LaTeX Projects\check\check.tex
[09:35:37] Parsing e:\LaTeX Projects\check\check.tex
[09:35:37] Cannot find file e:\LaTeX Projects\check\check.fls
[09:35:37] File watcher: responding to change in e:\LaTeX Projects\check\check.tex
[09:35:46] RECIPES command invoked.
[...]

PR #1500, opened 2019-07-11
TikzPicture Preview

tecosaur:
# TikzPicture Preview

gif preview

## Overview

 - Adds codelense before `\begin{tikzpicture}`
 - when clicked, starts tracking that tikzpicture by a `vscode.Range` using `vscode.workspace.onDidChangeTextDocument` and re-fetching the content of that range
 - when the tikzpicture content changes (and on initialisation)
 - a preamble is generated, and compared to the last preamble generated for the file
   - if the preamble has changed (and on initialisation) a standalone document with that preamble is written and compiled to a `.fmt` file (within a tempdir)
 - the tikzpicture associated file in the tempdir is compiled using the format file
 - on successful compilation the viewer refreshes

Closes #1450

## Needs work
Not quite at the stage where it would be good for merging. However! It is at the stage where you can take it out for a spin and provide feedback (please do).

PR #1506, opened 2019-07-12
Use a LaTeX parser

tamuratak:
I have written a [LaTeX parser](https://github.com/tamuratak/latex-utensils) based on [LaTeX.js](https://github.com/michael-brade/LaTeX.js) and [latex-parser](https://github.com/siefkenj/latex-parser) using [peg.js](https://github.com/pegjs/pegjs). It must be useful for LaTeX Workshop.

As the first step, let's use it to extract commands defined with multi lines in the preamble. Related to #1428.

スクリーンショット 2019-07-13 0 57 28

### DIFF
--- a/package.json
+++ b/package.json
@@ -1368,6 +1368,7 @@
     "fs-extra": "^5.0.0",
     "glob": "^7.1.1",
     "iconv-lite": "^0.4.24",
+    "latex-utensils": "^0.1.4",
     "mathjax": "^2.7.5",
     "mathjax-node": "2.1.1",
     "micromatch": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -1368,7 +1368,7 @@
     "fs-extra": "^5.0.0",
     "glob": "^7.1.1",
     "iconv-lite": "^0.4.24",
-    "latex-utensils": "^0.1.4",
+    "latex-utensils": "^0.1.7",
     "mathjax": "^2.7.5",
     "mathjax-node": "2.1.1",
     "micromatch": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -1368,7 +1368,7 @@
     "fs-extra": "^5.0.0",
     "glob": "^7.1.1",
     "iconv-lite": "^0.4.24",
-    "latex-utensils": "^0.1.7",
+    "latex-utensils": "latest",
     "mathjax": "^2.7.5",
     "mathjax-node": "2.1.1",
     "micromatch": "^3.0.0",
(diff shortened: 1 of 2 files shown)


tecosaur:
Nice! I could also do with this for getting stuff like `\tikzset` contents in #1500

tamuratak:
@James-Yu @jlelong if no objection, I want to merge this PR.

jlelong:
No objection but I have a few questions to better understand the motivation behind this change. I am probably already labelled as the _maintenance_ guy, so I am not afraid of saying to I am mainly concerned with maintainability in this change.
- What is the gain in terms of robustness for LaTeX-Workshop? I have the feeling you can parse `\newcommand` definitions split over several lines. Do you confirm that?
- What is the maturity of `latex-utensils`? 
- You are the only maintainer so all the pressure is on you. Do you fell we could rely more on `latex-utensils` for other things in LaTeX-Workshop?

tamuratak:
> What is the gain in terms of robustness for LaTeX-Workshop? I have the feeling you can parse \newcommand definitions split over several lines. Do you confirm that?

Yes. With latex-utensils, we can reduce the number of codes of parsing with regex. They easily lead to bugs and make maintenance difficult.

> What is the maturity of latex-utensils?

I think it is matured. At least for this PR. 

> Do you fell we could rely more on latex-utensils for other things in LaTeX-Workshop?

It is the motivation why I have developed latex-utensils. However, the scope of application would be limited because a LaTeX document being edited is often not grammatically correct and cannot be parsed without errors. A typical usage would be extracting a part of the document and paring it with latex-utensils.

---

The parser is generated by PEG.js from a file `latex.pegjs`. For example, LaTeX environment is defined as follows:

```js
Environment
  = beginEnv name:groupedEnvname args:(argumentList / Group)*
      skip_space body:(!(endEnv n:groupedEnvname & {return name === n;}) x:Element {return x;})*
    endEnv groupedEnvname
  {
    return { kind: "env", name, args, content: body, location: location() };
  }
```

tamuratak:
We can see AST (abstract syntax tree) returned by latex-utensils as follows:
```
$ cat sample/t.tex         
\documentclass[12pt]{article}
\begin{document}
  abcd
\end{document}
                       
$  ./node_modules/.bin/luparse -i --color sample/t.tex
{ kind: 'ast.root',
  content:
   [ { kind: 'command',
       name: 'documentclass',
       args:
        [ { kind: 'arg.optional',
            content: [ { kind: 'text.string', content: '12pt' } ] },
          { kind: 'arg.group',
            content: [ { kind: 'text.string', content: 'article' } ] } ] },
     { kind: 'env',
       name: 'document',
       args: [],
       content:
        [ { kind: 'text.string', content: 'abcd' },
          { kind: 'parbreak' } ] } ],
  comment: undefined }
```

jlelong:
I guess this is time for merging.

James-Yu:
Really looking forward to more robust syntax parsing by this parser!

PR #1522, opened 2019-07-20
Add tikzpicture environment

tecosaur:
Just adds `tikzpicture` to the environments list.

### DIFF
--- a/data/environments.json
+++ b/data/environments.json
@@ -56,6 +56,7 @@
   "tabular",
   "thebibliography",
   "titlepage",
+  "tikzpicture",
   "columns",
   "column"
 ]


jlelong:
Actually, the `tikzpicture` environment is already defined in `data/packages/tikz_env.json`. However for a strange reason, it does not pop up as a completion for `\begin... \end`. There must be a bug somewhere in the environment completion. I will look at it.

tecosaur:
For me this made it pop up, and it didn't before.

Screenshot from 2019-07-20 20-58-34

jlelong:
You have to load the `tikz` package to have `tikzpicture` defined.

jlelong:
I confirm that it does work but not with the [VSCodeVim](https://github.com/VSCodeVim/Vim) extension.

tecosaur:
I see. I have the issue where
 - I am using tikz
 - `\usapackage{tikz}` is not in my preamble.

Hence I don't receive any suggestion for the environment, which I find somewhat annoying.

I notice that this selective logic doesn't seem to be applied consistently. For example `\begin{bmatrix}` has a suggestion for the environment without `\usepackage{amsmath}` in the preamble. Should we make it so that the methodology is the same for environment suggestions?

If we do, I'd personally consider it to be better to provide suggestions for more environments than can be used than less (up to a point, for the common environments/packages at least).

tecosaur:
@jlelong I'd appreciate it if you could give me your thoughts on my previous comment.

jlelong:
I agree with you that the completion logic lacks a bit of coherence. This is mainly due to the way intellisense used to be provided prior the selection package mechanism.

I ma not against adding `tikzpicture` to the list of default environments. However, I feel a bit concerned with autocompletion for the commands provided by the `tikz` package if it (or any derivative) is not included somehow in the preamble.

tecosaur:
Another approach to this — would it be worth having a setting that manually marks packages as always "loaded" to get the completions for them? It may not be many users but I imagine I am one of some number of users that don't actually have all the packages in the preamble.

tecosaur:
Another approach to this — would it be worth having a setting that manually marks packages as always "loaded" to get the completions for them? It may not be many users but I imagine I am one of some number of users that don't actually have all the packages in the preamble.

jlelong:
For sure! It would help people who actually define their own packages/classes and need a way to declare the packages which are loaded internally.

jlelong:
Done. Use the new setting `latex-workshop.intellisense.package.extra`.

tecosaur:
Fantastic! I'll check it out.

Issue #1430, opened 2019-06-11
Math Brackets Match


chenr12:
I am wondering if it is possible to match
["\\{", "\\}"], ["\\left(", '\\right)'],  ["\\left[", '\\right]'],  ["\\left\\{", '\\right\\}']
as bracket pairs? 

It seems that currently only "(", "[" and "{" are recognized as brackets. So when paired brackets get highlighted, "\\left" and "\\right" will be left out.

I suppose we should be able to customize/extend bracket definitions for latex but haven't figured out how to do that. This may be a nice feature to have so that these math brackets would get highlighted as a whole, which would be even more useful for editing in combination with bracket coloring extensions like [Bracket Pair Colorizer 2](https://marketplace.visualstudio.com/items?itemName=CoenraadS.bracket-pair-colorizer-2).

jlelong:
We already match `["{", "}"], ["[", "]"], ["(", ")"]`. It is currently not possible to also match `["\left(", '\right)'], ["\left[", '\right]']` because of Microsoft/vscode/issues/58619.
I suggest to declare `"\left", "\right"` as a pair.

chenr12:
Thank you for the response (and also showing me how to modify the bracket definition)!

Although we can't get those math brackets as "autoClosingPairs", for those who are using bracket colorizer like me it is still useful to have them declared as "brackets". This is how they look like with  [Bracket Pair Colorizer 2](https://marketplace.visualstudio.com/items?itemName=CoenraadS.bracket-pair-colorizer-2), which it to find matching pairs.
<img width="679" alt="Screen Shot 2019-06-12 at 10 48 11 AM" src="https://user-images.githubusercontent.com/35321665/59366232-a0924380-8cff-11e9-8d3a-88caff1c79dc.png">

However, some people might not want to pair up `"\\left", "\\right"` as its scope would not be nested with other brackets. It is good to have it as an option though.

stone-zeng:
The pair `["\\left", "\\right"]` will often  make the rendering incorrect, e.g.

```tex
% See #1430
\left( \right)
```

![image](https://user-images.githubusercontent.com/18233088/76209739-8dd07f00-623d-11ea-8f6e-e004174ed87b.png)

```tex
% See #1641
{\leftmark }
```

![image](https://user-images.githubusercontent.com/18233088/76209667-64175800-623d-11ea-97d5-e5508e355775.png)

I use the extension Bracket Pair Colorizer 2 too.

[...]

PR #1534, opened 2019-07-24
Fix Math Brackets Highlight

### LINKED ISSUE #1529: Math Brackets Highlight
**Describe the bug [Required]**
probably related to James-Yu/LaTeX-Workshop#1430 and `Bracket Pair Colorizer 2`
The `\left` delimiter is not highlighted as in 1430's screenshot, 

**Expected behavior**
The `\left` delimiter pair is highlighted.

**Screenshots**
Screen Shot 2019-07-23 at 13 54 48
Screen Shot 2019-07-23 at 13 52 06

**Desktop [Required]:**
 - OS: macOS 10.14.5
 - VS Code version: 1.36.1
 - Extension version: 7.2.0

toooonyy:
close #1529, the code is from @chenr12

### DIFF
--- a/syntax/syntax.json
+++ b/syntax/syntax.json
@@ -6,7 +6,9 @@
 		["{", "}"],
 		["[", "]"],
 		["(", ")"],
-		["\\left", "\\right"]
+        ["\\left(", "\\right)"],
+        ["\\left[", "\\right]"],
+        ["\\left\\{", "\\right\\}"]
 	],
 	"autoClosingPairs": [
 		["{", "}"],


jlelong:
From the discussion in #1430.

> We already match `["{", "}"], ["[", "]"], ["(", ")"]`. It is currently not possible to also match `["\left(", '\right)'], ["\left[", '\right]']` because of Microsoft/vscode/issues/58619.
I suggest to declare `"\left", "\right"` as a pair.

Your patch is only good for people using `Bracket Pair Colorizer 2`, which has been reported to create some incompatibilities with LaTeX-Workshop. For other people, it is good to match `\left` and `\right` as a pair

Capture d’écran 2019-07-24 à 09 38 20

See the small boxes around `\left` and `\right`, they are rendered by vscode itself when `editor.matchBrackets` is set. This does not work with your patch. Because of this I do not think this is a good idea to merge your patch. Sorry.

WasabiFan:
@jlelong Can you help me understand the scenario here?

I locally applied this patch onto recent master (7f4a7786de0f343c48a486ee1d8a1d330943f6fb) and tried the example shown above. VSCode applies the box highlighting similar to how your screenshot shows it, however it includes the following parenthesis as well:

image

It seems to me like this works as expected. I would personally prefer this behavior, since the parenthesis after the `\left` or `\right` is semantically part of the same "brace pair".

Perhaps the behavior of vscode has changed since this was originally tested?

It's worth noting that this PR doesn't include `<` brackets, which would probably be good to include. Auto-closing braces' functionality is unchanged from its current behavior, as far as I can tell: it operates on the parenthesis alone unless you use the automatic `\left(` command suggestion.

For completeness, I tried enabling the Bracket Pair Colorizer 2 extension which was mentioned in some other threads; it now correctly highlights left/right pairs (because before the syntax definition would have suggested interleaved pairs of brackets) and no longer partially-colors commands which begin with `\left` or `\right` as a prefix, e.g. `\leftarrow`:

image

I tried out a few other "bracket"-based extensions (https://marketplace.visualstudio.com/items?itemName=chunsen.bracket-select, https://marketplace.visualstudio.com/items?itemName=2gua.rainbow-brackets) and they seem to work together with this extension as well.

All in all, I would appreciate this change being revisited. It's unfortunate VSCode doesn't seem to provide more intelligent handling of bracket pairs (e.g. allow a wildcard for the character following `\left` or `\right`); in the absence of that functionality, I think the change in this PR would improve the semantics that LaTeX Workshop reports to VSCode.

jlelong:
@WasabiFan Thanks for digging up this PR! You are right, this PR works fine now! Probably due to some changes in VSCode. I am merging it!

PR #1542, opened 2019-07-26
Tikzpreview fixes (2)

tecosaur:
One more sorry. I forgot to merge @trevorgunn's fix.


tecosaur:
Wow. That was speedy. 2 minutes has got to be a record.

Issue #1538, opened 2019-07-26
Consistency/linting roadmap


jlelong:
Following the long discussion in #1475, I think it would be nice to write a little roadmap to improve code consistency. 

### Before moving to `eslint`

- [x] Remove unnecessary escape in regex: `{`, `}` need no escape and neither do `[`, `]` when used inside character class. Rule `no-useless-escape`.
- [x] Strings should single quotes. Backticks should be reserved for strings using substitution. Rule: `quotes`.
- [x] Statements should not end with a `;`.
- [x] Remove all `// @ts-ignore` and fix the errors. These are trickier so I will wait until we merge #1543. See PR #1545.

### Breaking rules compared to `tslint`.

- [x] Remove extra space before `:` when giving explicit type. Rule `@typescript-eslint/type-annotation-spacing`.

### Non default `eslint` rules we want to enforce

- [x] Prefer `type` over `interface`. See [here](https://github.com/typescript-eslint/typescript-eslint/issues/433#issuecomment-487174104).
- [x] Ignore everything in `src/lib`.
- [x] Properties can be declared in the constructor.  set `"@typescript-eslint/no-parameter-properties": "off"`.

### Extra changes

- [x] Convert `.plist` syntax files to `.tmLanguage.json`. I find `.json` files much easier to edit and maintain and they allow for more elaborate constructions not available in the `.plist` (say `.xml`) format. The [vscode-tmlanguage](https://github.com/Togusa09/vscode-tmlanguage) extension can be used as a conversion tool.

Such changes may induce a lot of modifications and tend to blur file history. So, I want to make sure there is no objection before making any change. I will try to update this post with any further discussion.

The **Done** ticks refer to the developments in branch [format](https://github.com/James-Yu/LaTeX-Workshop/tree/format).

tamuratak:
I prefer `type` to `interface`. To my understanding, the difference is little. However, the declaration merging of `interface` might become a trouble-maker. See [link](https://github.com/typescript-eslint/typescript-eslint/issues/433#issuecomment-487174104).

tamuratak:
We can use [prefer-arrow-callback](https://eslint.org/docs/rules/prefer-arrow-callback) as an alternative to `only-arrow-functions`.

I prefer 
```json
prefer-arrow-callback: [ "error", { "allowUnboundThis": false } ]
```

This setting

> prohibits the use of function expressions as callbacks or function arguments entirely, without exception.

[...]

PR #1543, opened 2019-07-26
Formatting towards eslint

jlelong:
See #1538 to know what has already been done.

It has implied quite a lot of changes, so I would rather have a few people test before being merged.


tamuratak:
Please commit and push `package.json` also.

jlelong:
If everyone is fine with this, I suggest to merge these changes and move on to more stylistic rules later.

tamuratak:
Please revert the changes to files `src/lib/**/*.ts` too.

jlelong:
Everything should be fine now. @James-Yu  @tamuratak  are you ok with merging?

tamuratak:
Warnings `warning  Unexpected any` make us miss other warnings. Let's turn off `@typescript-eslint/no-explicit-any` in the first stage and consider what we can do later.

jlelong:
OK I have turned `@typescript-eslint/no-explicit-any` off but we should take some time to investigate this. Any more comments before merging?

tamuratak:
No. It seems to work well.

Issue #1512, opened 2019-07-15
BibTeX folder volume mappings for docker


rmrnc:
I appreciate your support for running latex in a docker container, but am missing support for BibTeX folders located on the docker host. This would require additional volume mappings in --AFAICT at least -- your `scripts/latexmk` script. E.g., 

```bash
docker run --rm --interactive \
    --env="BIBINPUTS=/usr/share/texmf/bibtex/bib" \
    --volume="${HOME}/bib:/usr/share/texmf/bibtex/bib" \
    --volume="${PWD}:${PWD}" \
    --workdir="${PWD}" \
    $LATEXWORKSHOP_DOCKER_LATEX latexmk $@
```

Let me add some observations:

1. Although the `BIBINPUTS` environment variable(s) could be assigned in the corresponding `Dockerfile` instead, I'd prefer it to be passed into the container at runtime thereby leveraging a VSCode specific configuration (see below). 
Why am I assigning a `BIBINPUTS` variable for the standard path `texmf/bibtex/bib` in the first place, you might ask? Pointing to the default TeXLive installation folder used by the Debian distro, i.e., `/usr/share/texmf`, by way of the `TEXMFROOT` environment variable (in `Dockerfile`), breaks LaTeX-Workshop's `latexmk` build, e.g., `! LaTeX Error: File 'book.cls' not found.` (at least on my box). I have no idea why that is.

2. The existing `"latex-workshop.latex.bibDirs"` configuration appears to provide a useful starting point for a volume mapping configuration. Though it would have to be declared as a mapping of host folders to docker folders instead of an array of host folders only, i.e,

```json
"latex-workshop.latex.bibDirs": [ 
    { "/path/to/host/folder1": "/path/to/mapped/docker/folder1" }, 
    ...
],
```

Thank you for your consideration.

tamuratak:
VS Code supports Docker with [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) now. Please try it. You can also use [Remote - WSL](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl).

PR #1551, opened 2019-07-30
Add ability to use custom docker wrapper scripts

garyng:
Rationale behind this functionality:

Currently LaTeX-Workshop only allow users to specify the image file to be used - but in some cases the users might want to pass extra args to `docker` (like volume mount used by [miktex](https://miktex.org/howto/miktex-docker), or use case like https://github.com/James-Yu/LaTeX-Workshop/issues/1512). This PR allows user to specify their own `docker` wrapper script via `latex-workshop.docker.wrapper`.

Todo:
- [ ] should the old option for specifying docker image be remained? If not, this might be a breaking change
- [x] `latexmk`
- [x] `texcount`
- [ ] `latexindent` (not working on my side, seems like its related to the generated temporary file)
- [ ] `synctex` (not working on my side either)
- [ ] some documentation

_I'm very new to both LaTeX and vscode extension development - so I might did some wrong here_ 😅

### DIFF
--- a/package.json
+++ b/package.json
@@ -305,7 +305,7 @@
       },
       {
         "command": "latex-workshop.showSnippetPanel",
-        "title": "Show Snippet Panel", 
+        "title": "Show Snippet Panel",
         "category": "LaTeX Workshop"
       }
     ],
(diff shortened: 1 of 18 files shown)


tamuratak:
Have you tried [VS Code Remote Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)? VS Code supports Docker now. I do not think we have to do more supports on the extension side.

garyng:
I thought it still needs vscode insiders... I'll give it a try, thanks!

tamuratak:
Feel free to reopen this PR if you want.

PR #1554, opened 2019-07-30
Multi file parser

James-Yu:
As name implies. This PR is a major refactor of the original messy multi-file parsing and watching logic. The main logic does not change a lot, only code re-aligned into various functions with more meaningful names. In addition, the original `texTree` is discarded by a new `cachedContent`, which serves both as a cache of the dirty tex files and the subfiles hierarchy. Lastly, a new function `getContent` is implemented to provide the full text with subfiles expanded from a given tex file, typically the root.

As this PR involves a re-write of the foundation of this extension, I would like to invite extensive tests on the robustness, especially looping `\inputs` and file-changes in all means. This PR also aims to re-implement all features provided by the original code. So please alert me if something is missing or edging cases are not considered. Thanks in advance 😆


James-Yu:
Additionally, I am really worried about the maintainability of the aging code without comments. I'll try to interpret them and add explanatory comments progressively.

jlelong:
@James-Yu I am currently away from my computer for a few days of vacation, I will not be able to look at all this in details before the week-end. 
Anyway this is a very initiative of yours! This is the core of the extension and maintenance is sometimes difficult (I remember when implementing a new `subfiles` logic), so congrats.

James-Yu:
@tamuratak do you have some time to do a test on this PR? Thx!

src/components/manager.ts
```diff
@@ -78,46 +85,50 @@ export class Manager {
...
-            }
-        }
-        if (wsroot) {
-            if (wsroot !== this.workspace) {
-                this.workspace = wsroot
-            }
+    private workspaceRootDir: string = ''
+    private findWorkspace() {
+        // Fallback to infer workspace from the root file
+        let uri = vscode.Uri.parse(this.rootFile ? this.rootFile : 'file:///')
```
tamuratak: Why fallback to 'file:///'? I think there is a security concern.

src/components/manager.ts
```diff
@@ -200,155 +211,139 @@ export class Manager {
...
+            if (reload) {
+                break
             }
+            if (path.relative(cachedFile, file) !== '') {
+                continue
+            }
+            return this.cachedContent[cachedFile].content
         }
-        return false
+        const fileContent = utils.stripComments(fs.readFileSync(file, 'utf-8').toString(), '%')
```
tamuratak: Why can we assume utf-8?

src/components/manager.ts
```diff
@@ -1,75 +1,82 @@
...
+                  .replace(/%DIR%/g, docker ? './' : path.dirname(texPath).split(path.sep).join('/'))
+                  .replace(/%TMPDIR%/g, this.extension.builder.tmpDir)
     }
 
     get rootDir() {
         return path.dirname(this.rootFile)
     }
 
+    private rootFiles: { [key: string]: string } = {}
     get rootFile() {
```
tamuratak: Please write a document on the difference between `rootFile` and `findRoot`. How can  we use them?

tamuratak:
I do not know much about `manager.ts`. @jlelong is appropriate for the reviewer. 

I think it has higher priority to refactor `roofFile` and `findRoot`. At least, we should document that `rootFile` can be `undefined`, and that calling `findRoot` is needed to ensure `rootFile` is not undefined.

src/components/manager.ts
```diff
@@ -1,75 +1,82 @@
...
+                  .replace(/%DOCFILE%/g, docfile)
+                  .replace(/%DIR%/g, docker ? './' : path.dirname(texPath).split(path.sep).join('/'))
+                  .replace(/%TMPDIR%/g, this.extension.builder.tmpDir)
     }
 
     get rootDir() {
         return path.dirname(this.rootFile)
     }
 
+    private rootFiles: { [key: string]: string } = {}
```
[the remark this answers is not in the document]
James-Yu: Because `undefined` cannot be saved in `rootFiles`. The only assignment statement is L126->L65, which is protected by the undefined check at L121.

src/components/manager.ts
James-Yu: I don't know 🤣 It was as is. Will remove it.

src/components/manager.ts
James-Yu: Will add comments. In short, `rootFile` is an interface to `rootFiles`, which holds cache of roots w.r.t. workspaces. It is edited by the method `findRoot`.

Want to re-find the root file? -> `findRoot()`. It is costly so cannot invoke upon keystroke.
Want to just obtain the save root file location? -> `rootFile`

James-Yu:
`findRoot` loops through many related files on the disk/in the memory to search for root. The root, after found, is saved in `rootFile`. That is the main difference. The former is too costly that we cannot invoke it whenever we need the root.

James-Yu:
It is possible to have `undefined` rootFile. Will do refactor on it.

src/components/manager.ts
```diff
@@ -200,155 +227,139 @@ export class Manager {
         const rootFilesExcludePatterns = configuration.get('latex.search.rootFiles.exclude') as string[]
         const rootFilesExcludeGlob = rootFilesExcludePatterns.length > 0 ? '{' + rootFilesExcludePatterns.join(',') + '}' : undefined
         try {
-            const urls = await vscode.workspace.findFiles(rootFilesIncludeGlob, rootFilesExcludeGlob)
-            for (const url of urls) {
-                const content = utils.stripComments(fs.readFileSync(url.fsPath).toString(), '%')
+            const files = await vscode.workspace.findFiles(rootFilesIncludeGlob, rootFilesExcludeGlob)
+            for (const file of files) {
+                const content = utils.stripComments(fs.readFileSync(file.fsPath).toString(), '%')
```
tamuratak: This change does not solve the problem since the default encoding of `buf.toString` is utf-8. See [link](https://nodejs.org/api/buffer.html#buffer_buf_tostring_encoding_start_end).

James-Yu: `stripComments` is implemented by @jlelong several months ago. Need to wait for him to answer this.

James-Yu: Additionally, we intensively use `fs.readFileSync(someFile).toString()` throughout the extension. I am not aware of other approaches to read a text file into string. All in all we need to guess an encoding, or find a way (I have no idea) to determine one.

tamuratak: I do not know why this code does not cause problems.  As long as no issue is reported by users, I think it is OK.

However, if you expect the newly introduced API, `getContent`, will be used frequently in LaTeX Workshop, it would cause problems.

tamuratak:
This PR has several errors on eslint. It seems that the restyled bot does not warn that.

```
/Users/tamura/src/github/LaTeX-Workshop/src/components/manager.ts
   49:13  error  'out' is never reassigned. Use 'const' instead  prefer-const
   60:78  error  Trailing spaces not allowed                     no-trailing-spaces
  110:23  error  Trailing spaces not allowed                     no-trailing-spaces
  243:68  error  Unexpected a space before the ':'               @typescript-eslint/type-annotation-spacing
  351:78  error  Unexpected a space before the ':'               @typescript-eslint/type-annotation-spacing

✖ 5 problems (5 errors, 0 warnings)
  5 errors and 0 warnings potentially fixable with the `--fix` option.
```

James-Yu:
Interestingly, the `eslint` extension in vscode also did not warn about the errors.

.eslintrc.json
```diff
@@ -53,7 +53,10 @@
         "prefer-const": "error",
         "quotes": ["error", "single", {"avoidEscape": true}],
         "@typescript-eslint/semi": ["error", "never"],
-        "@typescript-eslint/member-delimiter-style": {"delimiter": "semi", "requireLast": false},
```
tamuratak: Do not change  `.eslintrc.json` at the same time as other changes to keep the history clean.

James-Yu: It's not a change, it's a fix. The previous one is buggy for eslint v6. Obviously, the error level was missing.

tamuratak: You must open a separate PR. This change causes the following errors:

```
$ ./node_modules/.bin/eslint --ext .ts .

/Users/tamura/src/github/LaTeX-Workshop/dev/createSymbolSvgs.ts
   5:18  error  Expected a comma          @typescript-eslint/member-delimiter-style
   6:23  error  Expected a comma          @typescript-eslint/member-delimiter-style
   7:20  error  Expected a comma          @typescript-eslint/member-delimiter-style
   8:21  error  Expected a comma          @typescript-eslint/member-delimiter-style
   9:23  error  Expected a comma          @typescript-eslint/member-delimiter-style
  10:18  error  Expected a comma          @typescript-eslint/member-delimiter-style
  11:22  error  Unexpected separator (;)  @typescript-eslint/member-delimiter-style
  43:47  error  Unexpected separator (;)  @typescript-eslint/member-delimiter-style
  44:11  error  Unexpected separator (;)  @typescript-eslint/member-delimiter-style
  62:52  error  Expected a comma          @typescript-eslint/member-delimiter-style
  63:55  error  Expected a comma          @typescript-eslint/member-delimiter-style
  64:51  error  Expected a comma          @typescript-eslint/member-delimiter-style
  65:60  error  Expected a comma          @typescript-eslint/member-delimiter-style
  66:51  error  Unexpected separator (;)  @typescript-eslint/member-delimiter-style

✖ 14 problems (14 errors, 0 warnings)
  14 errors and 0 warnings potentially fixable with the `--fix` option.

```

src/components/manager.ts
jlelong: The `stripComments` function simply remove everything from a comment sign to the end of line by taking care of escaped comment signs. My 2 cents on the encoding issue
- we mostly only search strings of the form `\ + ascii_char` and this is not sensitive to the encoding
- `utf-8` has become widely used
- We may troubles with labels and file names (but I am not sure `latex` handles non ascii file names well).

Is `getContent` not restricted to getting the content or an editor or can it read a file from the disk?

src/commander.ts
```diff
@@ -204,10 +209,10 @@ export class Commander {
             return
         }
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
-        let pdfFile: string
+        let pdfFile: string | undefined
```
jlelong: I guess we should set `pdfFile` as `undefined`.

src/components/manager.ts
```diff
@@ -200,155 +231,139 @@ export class Manager {
...
+       async parseFileAndSubs(file: string, onChange: boolean = false) {
+        this.extension.logger.addLogMessage(`Parsing ${file}`)
+        if (this.filesWatched.indexOf(file) < 0) {
+            // The file is first time considered by the extension.
+            this.extension.logger.addLogMessage(`Adding ${file} to file watcher.`)
+            this.fileWatcher.add(file)
+            this.filesWatched.push(file)
+            // So update completer w.r.t. this file. Subsequent updates will be
+            // handled in 1) onDidChangeTextDocument (in vscode) and 2) file
+            // watcher (changed externally)
```
jlelong: A long time ago, we had two handlers as you describe: one for changes inside vscode and one for external change. Then, we only kept the external one, which actually watches when a file changes on disk and therefore responds when a file is saved inside vscode. I am wondering if you really need to keep both. If I remember well, `onDidChangeTextDocument` is called after every keystroke or so.

PR #1561, opened 2019-08-01
Remove TikzPreview feature (move to LU)

tecosaur:
Does what the title says.


src/main.ts
```diff
@@ -141,40 +141,42 @@ function conflictExtensionCheck() {
 }
 
 function newVersionMessage(extensionPath: string, extension: Extension) {
```
James-Yu: Instead of wrapping a promise, you may simply put `async` before `function` to promisify the result. It is also git-log-friendly.

tecosaur: Does that work with the callback function? I haven't done that before.

tecosaur: i.e.
```
async function example() {
    functionwithcallback((data) => {
        await data
    })
}
````

because I thought you had to do promises for that?

**Comment by tecosaur** [comment]
Not sure what went on with f133d7728cd2a079814ea3c383823cc9d932e220. For some reason when starting the extension I would get the error `...can't get "version" of undefined` (paraphrasing) referencing line 363 in `main.ts`

```ts
console.log(`LaTeX Workshop version ${extension.packageInfo.version} activated.`)
```

I had a look at `newVersionMessage` where `extension.packageInfo` is set and noticed that it was set inside the callback for `fs.readFile`, and so since it is called in an async function it looks like `packageInfo.version` was accessed on line 363 before the callback of `fs.readFile`.

James-Yu:
If that is the only warning, the related changes can be reverted. That line of debug should be removed in `master`, but I mistakenly removed it in #1554. Will do so soon.

tecosaur:
Yea, that's it. I'll revert it.

tecosaur:
Accidently commited me commenting out that problematic version line. With the added api bits this should be fine to merge now.

src/main.ts
```diff
@@ -367,7 +361,17 @@ export async function activate(context: vscode.ExtensionContext) {
     return {
         getRootFile: () => extension.manager.rootFile,
```
James-Yu: It seems that this is duplicating the `rootFile` below. What about removing this and making corresponding changes in LU, specifically in word counter? We may also need to have a general rule of thumb in what structure api should be roughly arranged.

tecosaur: The reason that's there is because of these lines in `LU`.
```ts
if (this.extension.workshop.manager.rootFile() === undefined) {
    await this.extension.workshop.manager.findRoot()
}
```
If that's unnecesary I'll just remove that.

Re: api structure, what've I've been going with so far is just mirroring the structure in LW, for example, see above.

James-Yu: Yes, please remove. I will handle the word counter changes.

tecosaur: done

PR #1564, opened 2019-08-01
Make toggleSelectedKeyword work for multi-selections

tecosaur:
Instead of just using the first selection, this PR makes it so that every selection is looped though, and any resulting increase or decrease in content is saved as an offset which is used for future loop iteration edits.

There will be no line offset with the current keywords we're using, but keeping track of that as well is very small adition and makes this slightly more robust going forward.

### DIFF
--- a/src/commander.ts
+++ b/src/commander.ts
@@ -552,7 +552,7 @@ export class Commander {
 
         const actions: ('added' | 'removed')[] = []
 
-        selectionLoop: for (let i = 0; i < selections.length; i++) {
+        for (let i = 0; i < selections.length; i++) {
             const selection = selections[i]
             const selectionText = selectionsText[i]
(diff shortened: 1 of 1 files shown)


James-Yu:
I am fine with the new logic, yet the `continue selectionLoop` makes the control flow a bit messy. It could be replaced by common controls instead of a GOTO style control.

tecosaur:
~~The~~ An alternative would be to make it so that the `continue` is replaced with setting a flag + `break`, and then in the outer loop `if flag == true; continue`. However to me that's messier than continuing a labeled loop. Would you rather that way, or have another approach you'd preffer?

tecosaur:
@James-Yu have you any thoughts on my last comment?

tecosaur:
@James-Yu *cough

PR #1569, opened 2019-08-02
run eslint against viewer/latexworkshop.js

tamuratak:
Run eslint against `viewer/latexworkshop.js`.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,7 +3,8 @@ dev/**/*.js
 resources/**/*.js
 out
 node_modules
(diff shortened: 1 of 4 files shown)


tamuratak:
Restyled seems useless.

tecosaur:
Wow. That was a lot of changes. I have to say if it's not using our `.eslint` I don't think it's that useful. It also seems like it's prettier config directly contradicts some of our eslint rules (e.g. quote style)

PR #1582, opened 2019-08-06
Use AST to parse reference instead of regex

James-Yu:
This is a first attempt to adopt @tamuratak 's LaTeX parser in providing completion suggestions. All original functionalities should be preserved. TeX files are parsed on being added to the file watcher, changes saved to disk, and calling `\ref` (only for active editor). As the incremental update is not quite computationally expensive, the refresh timer is remove. Yet reinclude that is trivial if needed.

### DIFF
--- a/src/providers/completer/reference.ts
+++ b/src/providers/completer/reference.ts
@@ -1,114 +1,109 @@
 import * as vscode from 'vscode'
 import * as fs from 'fs'
 import * as path from 'path'
-import { stripComments } from '../../utils'
+import {latexParser} from 'latex-utensils'
 
 import {Extension} from '../../main'
 
-export type ReferenceEntry = {
-    item: {
-        reference: string,
-        text: string,
-        position: vscode.Position,
-        atLastCompilation?: {refNumber: string, pageNumber: string}
-    },
-    text: string,
-    file: string
+export interface Suggestion extends vscode.CompletionItem {
+    file: string, // The file that defines the ref
+    position: vscode.Position, // The position that defines the ref
+    prevIndex?: {refNumber: string, pageNumber: string} // Stores the ref number
 }
 
 export class Reference {
     extension: Extension
-    suggestions: vscode.CompletionItem[]
-    referenceData: {[id: string]: ReferenceEntry} = {}
-    refreshTimer: number
+    // Here we use an object instead of an array for de-duplication
+    suggestions: {[id: string]: Suggestion} = {}
 
     constructor(extension: Extension) {
         this.extension = extension
     }
 
-    reset() {
-        this.suggestions = []
-        this.referenceData = {}
-        this.refreshTimer = 0
-    }
-
     provide(args: {document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken, context: vscode.CompletionContext}): vscode.CompletionItem[] {
-        if (Date.now() - this.refreshTimer < 1000) {
-            return this.suggestions
-        }
-        this.refreshTimer = Date.now()
-        const suggestions: {[key: string]: ReferenceEntry['item']} = {}
-        Object.keys(this.referenceData).forEach(key => {
-            suggestions[key] = this.referenceData[key].item
-        })
+        // Update the dirty content in active text editor
         if (vscode.window.activeTextEditor) {
-            const items = this.getReferenceItems(vscode.window.activeTextEditor.document.getText())
-            Object.keys(items).forEach(key => {
-                if (!(key in suggestions)) {
-                    suggestions[key] = items[key]
+            const content = vscode.window.activeTextEditor.document.getText()
+            const refs = this.getReferenceFromAST(latexParser.parse(content), content.split('\n'))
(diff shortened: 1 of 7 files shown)


src/providers/completer/reference.ts
```diff
@@ -1,114 +1,109 @@
...
-                items[result[1]] = {
-                    reference: result[1],
-                    text: `${noELContent.substring(prevContent.lastIndexOf('\n') + 1, result.index + followLength)}\n...`,
-                    position: new vscode.Position(positionContent.length - 1, positionContent[positionContent.length - 1].length)
+    update(file: string, content: string) {
+        const refs = this.getReferenceFromAST(latexParser.parse(content), content.split('\n'))
+        this.extension.manager.cachedContent[file].element.reference = refs
+    }
+
+    private getReferenceFromAST(ast: latexParser.LatexAst | latexParser.Node, lines: string[], nextNode?: latexParser.Node): vscode.CompletionItem[] {
```
tamuratak: `LatexAst` and `Node` are completely different. Please split `getReferenceFromAST` defining `getReferenceFromNode`.

src/providers/completer/reference.ts
```diff
@@ -1,114 +1,109 @@
...
+        let refs: vscode.CompletionItem[] = []
+        let label = ''
+        if (ast.kind === 'command' && ast.name === 'label') {
```
tamuratak: You can use a type guard `isCommand`. See [link](https://github.com/tamuratak/latex-utensils/blob/master/src/latex_parser_types.ts#L18).

James-Yu: No, typescript cannot recognize the data type with `isCommand`-like guard. Try and see.

src/providers/completer/reference.ts
James-Yu: There are two classes: command/string, and others (node and root). So which one is AST?

src/providers/completer/reference.ts
tamuratak: スクリーンショット 2019-08-06 20 23 22

src/providers/completer/reference.ts
```diff
@@ -1,114 +1,109 @@
...
+                label,
+                kind: vscode.CompletionItemKind.Reference,
+                documentation: lines.slice(ast.location.start.line - 2, ast.location.end.line + 4).join('\n'),
+                range: new vscode.Range(ast.location.start.line - 1, ast.location.start.column,
+                                        ast.location.end.line - 1, ast.location.end.column)
+            })
+            // Here we abuse the definition of range to store the location of the reference definition
+            return refs
+        }
+        if (ast.hasOwnProperty('content') && Array.isArray(ast['content'])) {
```
tamuratak: I have added `hasContent` and `hasContentArray`. Please use them. See [link](https://github.com/tamuratak/latex-utensils/blob/master/src/latex_parser_types.ts#L223).

src/providers/completer/reference.ts
tamuratak: How about the following? I have not tested it yet.  [branch](https://github.com/tamuratak/LaTeX-Workshop/tree/completer-rework_2).

```typescript
    update(file: string, content: string) {
        const refs = this.getReferenceFromNodeArray(latexParser.parse(content).content, content.split('\n'))
        this.extension.manager.cachedContent[file].element.reference = refs
    }

    private getReferenceFromNodeArray(nodes: latexParser.Node[], lines: string[]): vscode.CompletionItem[] {
        let refs: vscode.CompletionItem[] = []
        for (let index = 0; index < nodes.length; ++index) {
            if (index < nodes.length - 1) {
                // Also pass the next node to handle cases like `label={some-text}`
                refs = refs.concat(this.getReferenceFromNode(nodes[index], lines, nodes[index+1]))
            } else {
                refs = refs.concat(this.getReferenceFromNode(nodes[index], lines))
            }
        }
        return refs
    }

    private getReferenceFromNode(node: latexParser.Node, lines: string[], nextNode?: latexParser.Node): vscode.CompletionItem[] {
        let refs: vscode.CompletionItem[] = []
        let label = ''
        if (latexParser.isCommand(node) && node.name === 'label') {
            // \label{some-text}
            const group = node.args.filter(latexParser.isGroup)[0]
            label = group ? latexParser.stringify(group.content) : label
        } else if (latexParser.isTextString(node) && node.content === 'label=' && nextNode !== undefined) {
            // label={some=text}
            if (latexParser.isGroup(nextNode)) {
                label = latexParser.stringify(nextNode.content)
            }
        }
        if (label !== '' && (latexParser.isCommand(node) || latexParser.isTextString(node))) {
            refs.push({
                label,
                kind: vscode.CompletionItemKind.Reference,
                documentation: lines.slice(node.location.start.line - 2, node.location.end.line + 4).join('\n'),
                range: new vscode.Range(node.location.start.line - 1, node.location.start.column,
                                        node.location.end.line - 1, node.location.end.column)
            })
            // Here we abuse the definition of range to store the location of the reference definition
            return refs
        }
        if (latexParser.hasContentArray(node)) {
            refs = refs.concat(this.getReferenceFromNodeArray(node.content, lines))
        }
        return refs
    }
```

PR #1587, opened 2019-08-07
Use AST to provide environment suggestions

James-Yu:
Similar to #1582 , but for environment.

### DIFF
--- a/src/providers/completer/environment.ts
+++ b/src/providers/completer/environment.ts
@@ -73,8 +73,9 @@ export class Environment {
     private getEnvFromNode(node: latexParser.Node, lines: string[]): vscode.CompletionItem[] {
         const envs: vscode.CompletionItem[] = []
         let label = ''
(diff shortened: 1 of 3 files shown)


src/providers/completer/environment.ts
```diff
@@ -1,81 +1,103 @@
...
         }
+        return envs
     }
 
-    getEnvironmentsTeX(filePath: string) {
-        Object.assign(this.environmentsInTeX, this.getEnvironmentItems(fs.readFileSync(filePath, 'utf-8')))
+    private getEnvFromNode(node: latexParser.Node, lines: string[]): vscode.CompletionItem[] {
+        const envs: vscode.CompletionItem[] = []
+        let label = ''
+        if ((latexParser.isEnvironment(node) || node.kind.indexOf('env.') > -1) && // e.g., 'env.math.align', 'env.verbatim'
```
tamuratak: How about the following?

```typescript
if (latexParser.isEnvironment(node) || latexParser.isMathEnv(node) || latexParser.isMathEnvAligned(node)) {
```

PR #1594, opened 2019-08-12
Bibtex parser rework

James-Yu:
Similar to the previous rework on reference and environment autocompletion. No features are changed, only engineering-side changes to `latex-utensils`.

### DIFF
--- a/src/providers/completer/citation.ts
+++ b/src/providers/completer/citation.ts
@@ -88,7 +88,7 @@ export class Citation {
         return entries
     }
 
-    private getIncludedBibs(file: string | undefined, visitedTeX: string[] = []) {
+    private getIncludedBibs(file?: string, visitedTeX: string[] = []) {
         if (file === undefined) {
             // Only happens when rootFile is undefined
             return Object.keys(this.bibEntries)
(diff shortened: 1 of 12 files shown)


src/providers/completer/citation.ts
```diff
@@ -1,137 +1,67 @@
...
+    detail: string,
+    fields: {[key: string]: string},
+    file: string,
+    position: vscode.Position
 }
 
 export class Citation {
     extension: Extension
-    suggestions: vscode.CompletionItem[]
-    citationInBib: { [id: string]: {citations: CitationRecord[], rootFiles: string[] }} = {}
```
jlelong: citations used to be tight to a list of LaTeX projects via `rootFiles` making it possible to have different citation completions depending on the active editor without forgetting/reparsing `.bib` files every time the time changes (see #1085). How is it done now?

James-Yu: I missed that 🤣 The latest commit re-implement the feature. Here is a sketch:

First, the included bib file information is now stored in `cachedContent`. Additionally, these bib files are watched by watcher, which updates the entries upon changes.

When the extension provides citation suggestions (inline or browser), it first traces from the root file and investigate all included bib files along the parent-child-tree. Only those included will be used in the suggestion, along with inline `bibitem`s.

Issue #1476, opened 2019-07-04
VSCode 1.36 breaks syntax highlighting with minted environments


autinerd:
**Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]**

You still see this issue?: Yes

**Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.**

**Describe the bug [Required]**
When using a minted environment, the syntax highlighting after the code inside the minted environment doesn't work properly. It still uses the first one for all next minted environments, and LaTeX syntax highlighting doesn't work properly at all (numbers and the word before and after a period are highlighted)

**To Reproduce**
Insert a minted environment into the document.
Examples:
```tex
\begin{listing}[htbp]
    \begin{minted}{python}
def main():
    return [test for test in range(3)]
    \end{minted}
\end{listing}

\begin{listing}[htbp]
    \begin{minted}{c}
struct test {
int test2;
int test3;
int test4;
}
    \end{minted}
\end{listing}
This is a test sentence. Here is a number: 39
```

**Expected behavior**
Everything works like it did in VSCode 1.35

**Logs [Required]**
No messages in both logs, which can say anything about it.

**Screenshots**
![Test 1](https://user-images.githubusercontent.com/27780930/60664231-a643fa80-9e61-11e9-92c0-02b51179d650.png)
![Test 2](https://user-images.githubusercontent.com/27780930/60664247-b3f98000-9e61-11e9-9546-dcce70d01553.png)

**Desktop [Required]:**
 - OS: Linux Mint
 - VS Code version: 1.36
 - Extension version: 7.0.2

**Additional context**
With 1.35 everything was fine.

jlelong:
This issue appears only with `c` and `cpp`, so I think it is a bug in the C/C++ grammar. Using `java` code works fine
```
\begin{listing}[htbp]
    \begin{minted}{python}
def main():
    return [test for test in range(3)]
    \end{minted}
\end{listing}

\begin{listing}[htbp]
    \begin{minted}{java}
class test {
int test2;
int test3;
int test4;
}
    \end{minted}
\end{listing}
This is a test sentence. Here is a number: 39
````
I think a bug report should be filled against vscode repo.

autinerd:
[...]

PR #1604, opened 2019-08-15
Prefer source.cpp.embedded.latex in cpp minted env

matter123:
C++ side PR: https://github.com/jeff-hykin/cpp-textmate-grammar/pull/351

This should completely fix https://github.com/microsoft/vscode/issues/76603 and #1476.

The grammar for the scope `source.cpp.embedded.latex` is a modified version of `source.cpp` that closes any ranges when a positive lookahead for /\\end\{minted\}/ is matched.

### DIFF
--- a/syntax/LaTeX.tmLanguage.json
+++ b/syntax/LaTeX.tmLanguage.json
@@ -138,6 +138,9 @@
           "contentName": "source.cpp",
           "patterns": [
             {
+              "include": "source.cpp.embedded.latex"
+            },
+            {
               "include": "source.cpp"
             }
           ],


jlelong:
Thanks for the PR. I guess we should wait until  jeff-hykin/cpp-textmate-grammar#351 is merged and the new grammar is in use by vscode before merging. Do you confirm?

matter123:
There is no harm in either waiting or eagerly merging. This PR adds `source.cpp.embedded.latex` as a first choice, `source.cpp` will still be used if the grammar doesn't exist.

For example, this fallback behavior is used by the C++ grammar to support multiple flavors of embedded assembly see line 16354 of [the syntax](https://github.com/jeff-hykin/cpp-textmate-grammar/blob/master/syntaxes/cpp.tmLanguage.json#L16354).

There would be no change to existing behavior until both PRs have been merged in, and the order in which that happens is not important.

PR #1605, opened 2019-08-16
Command completer rework

James-Yu:
Similar to #1594, but on command auto-completion. No features are changed, only engineering-side changes to `latex-utensils`.

edit: btw this should be the last PEG-related refactor by me. I plan to make a new release after this one is settled, which might be a new major version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -520,11 +520,11 @@
         "mac": "cmd+l cmd+m",
         "when": "textInputFocus && editorLangId == 'latex'"
       },
-      {	
-        "key": "ctrl+l ctrl+w",	
-        "mac": "cmd+l cmd+w",	
-        "when": "editorTextFocus && !editorReadonly && editorHasSelection && editorLangId == 'latex'",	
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -520,12 +520,6 @@
         "mac": "cmd+l cmd+m",
         "when": "textInputFocus && editorLangId == 'latex'"
       },
-      {
-        "key": "ctrl+l ctrl+w",
-        "mac": "cmd+l cmd+w",
-        "when": "editorTextFocus && !editorReadonly && editorHasSelection && editorLangId == 'latex'",
-        "command": "latex-workshop.surround"
-      },
```
jlelong: Humm, I do not think `latex-workshop.surround` has ever been deprecated. I think it is only the use of `\` to trigger surrounding that has been deprecated, see #844. I would rather have this feature kept. Tell me if I missed something.

James-Yu: You are correct. I mistaken this one with `wrap-env`. Will restore soon.

Issue #1464, opened 2019-06-27
Customizable shortcut for Backward/Reverse synctex (pdf to source)


ataiya:
In the documentation I read:
```
Backward/Reverse synctex (pdf to source) is activated by ctrl + left-clicking (cmd + left-clicking on Mac) on the relevant element of the pdf preview.
```

However, other popular latex editors (e.g. overleaf) enable synctex by a simple double-click on the PDF. It would be good if the extension would allow for the shortcut to be used for this operation to be customized?

trevorgunn:
The shortcut exists in the viewer webpage which is a stand-alone web-document that you could view in any web browser. As such, it exists outside of the VSCode API so it is not as straightforward to access user settings in the viewer.

The shortcut is hardcoded here: https://github.com/James-Yu/LaTeX-Workshop/blob/aeaa583f2c118cbe0c6cd112ef0ee7c85e675616/viewer/latexworkshop.js#L249-L252 and you can patch it to have the shortcut you want easily enough by changing `onclick` to `ondblclick` and removing the `if` block. Here's a diff that you can use to patch:

```diff
 249,253c249
<     canvas_dom.onclick = (e) => {
<         if (!(e.ctrlKey || e.metaKey)) {
<           return
<         }
< 
---
>     canvas_dom.ondblclick = (e) => {
```

One reason why ctrl-click is preferred is because double-clicking will select text. (Overleaf deals with this by unselecting the text after synctex.)

ataiya:
Gotcha! Too bad the *.js cannot be parameterized to receive a shortcut spec on load.
Thanks for the patch!

Issue #662, opened 2018-06-09
SyncTex does not work when double clicking the pdf preview


K9A2:
### Description

This built-in synctex doesn`t work when I try to jump to corresponding position in a .tex file by double clicking the generated pdf file view by command `LaTex Workshop: View LaTex PDF file in VSCode tab`.
But it can jump form .tex file to pdf. See the log if necessary.

### Log messages

<details>
<summary>LaTeX Workshop Output</summary>

</details>

**Here is the full log including compiling, viewing and clicking.**

```
[11:56:45] Root file remains unchanged from: /home/stormlin/Downloads/paper/main.tex.
[11:56:45] Root file remains unchanged from: /home/stormlin/Downloads/paper/main.tex.
[11:56:47] Root file remains unchanged from: /home/stormlin/Downloads/paper/main.tex.
[11:56:51] BUILD command invoked.
[11:56:51] Root file remains unchanged from: /home/stormlin/Downloads/paper/main.tex.
[11:56:51] Building root file: /home/stormlin/Downloads/paper/main.tex
[11:56:51] Build root file /home/stormlin/Downloads/paper/main.tex
[11:56:51] Recipe step 1: latexmk, -synctex=1,-interaction=nonstopmode,-file-line-error,-pdf,/home/stormlin/Downloads/paper/main
[11:56:54] LaTeX log parsed with 2 messages.
[11:56:54] Recipe of length 1 finished.
[11:56:54] Successfully built /home/stormlin/Downloads/paper/main.tex
[11:56:54] No PDF viewer connected for /home/stormlin/Downloads/paper/main.pdf
[11:56:54] Root file remains unchanged from: /home/stormlin/Downloads/paper/main.tex.
[11:56:54] File cleaned: /home/stormlin/Downloads/paper/main.aux
[11:56:54] File cleaned: /home/stormlin/Downloads/paper/main.fls
[11:56:54] File cleaned: /home/stormlin/Downloads/paper/main.bbl
[11:56:54] File cleaned: /home/stormlin/Downloads/paper/main.log
[11:56:54] File cleaned: /home/stormlin/Downloads/paper/main.blg
[11:56:54] File cleaned: /home/stormlin/Downloads/paper/main.fdb_latexmk
[11:57:05] TAB command invoked.
[11:57:05] Root file remains unchanged from: /home/stormlin/Downloads/paper/main.tex.
[...]

Issue #106, opened 2017-04-10
Synctex and MikTeX


zachjweiner:
With the ability to set the synctex path, I think I inched closer to getting synctex to work with LaTeX Workshop with a MikTeX installation, in that I get a different error now :)

First, I downloaded `synctex.exe` and `kpathsea622.dll` and placed them according to the second comment to [this answer](http://tex.stackexchange.com/a/338117/78245).

Then, sensibly, I set `"latex-workshop.synctex.path": "C:\\Program Files\\MiKTeX 2.9\\miktex\\bin\\x64\\synctex.exe"`.

Now, attempting to synctex in either direction produces the error `Cannot synctex, code: 3221225781`, which differs from errors thrown by being unable to find `synctex.exe`. Googling fails me at this point, but I thought I'd crowdsource, since it'd be nice to have a setup for MikTeX+LaTeX Workshop users.

James-Yu:
Does it work well from command line? You may try executing the same command from cmd and see if problem persists.

zachjweiner:
Thanks for the suggestion, @James-Yu! The command line had better errors that lead me to the issue.

I needed `msvcr100.dll`. Downloading and placing this file itself does not work. If anyone happens upon this, just download both versions of Microsoft Visual C++ 2010 (32 and 64 bit - but I think 32 bit did the trick?) a la [here](https://www.faqforge.com/windows/fix-the-program-cant-start-because-msvcr100-dll-is-missing-from-your-computer-error-on-windows/). Then, viola!

zachjweiner:
Followup question: is there any way to set, say, double clicking on the pdf to trigger synctex?

James-Yu:
There is. The problem is, `ctrl`+`lclick` is almost a 'standard' for reverse SyncTeX. There is not much incentive for this extension to 'break' the convention.

zachjweiner:
Ok, good to know. (I'm used to double clicking in ShareLaTeX.) Yea don't change it on your end, was just wondering if there was a setting or whatnot. Thanks.

GitHubRulesOK:
SumatraPDF & PDF file with valid .synctex(.gz) will run its inverse command to open almost ANY valid editor with tex (editor needs a  form of command line call to not open fresh instance)

PR #1623, opened 2019-08-20
Configurable reverse synctex keybinding

### LINKED ISSUE #1596: Double click on PDF to jump to code

Coming from Overleaf, I often try to double click on the PDF to jump to the code. However, I need to use CMD+click. 

Add double click as a trigger for synctex. 

It could be configurable. 

**Additional context**

This came up in https://github.com/James-Yu/LaTeX-Workshop/issues/662.

trevorgunn:
This has come up in several issues now:

#1596 
#1464 
#662
#106

With double clicking being the default on Overleaf and other online editors, I think it is worth having an option to have double clicking as a keybind.

Closes #1596

### DIFF
--- a/src/components/viewer.ts
+++ b/src/components/viewer.ts
@@ -230,7 +230,9 @@ export class Viewer {
                             spreadMode: configuration.get('view.pdf.spreadMode'),
                             hand: configuration.get('view.pdf.hand'),
                             invert: configuration.get('view.pdf.invert'),
-                            keybinding: configuration.get('view.pdf.internal.synctex.keybinding')
+                            keybindings: {
+                                synctex: configuration.get('view.pdf.internal.synctex.keybinding')
+                            }
                         }))
                     }
                     if (configuration.get('synctex.afterBuild.enabled') as boolean) {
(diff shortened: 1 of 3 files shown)


tamuratak:
I do not like this approach. We should avoid that each new key binding requires a new `type` to be defined in the protocol. 

I prefer that, through a new type `keybinding`, key bindings configured by users are send to `latexworkshop.js`.

trevorgunn:
@tamuratak Thank you for your comment. Please let me know if the latest commit resolves your concerns.

src/components/viewer.ts
```diff
@@ -230,6 +230,7 @@ export class Viewer {
                             spreadMode: configuration.get('view.pdf.spreadMode'),
                             hand: configuration.get('view.pdf.hand'),
                             invert: configuration.get('view.pdf.invert'),
+                            keybinding: configuration.get('view.pdf.internal.synctex.keybinding')
```
tamuratak: To make other keybindings configurable, please do not set the value of the keybinding of synctex directly.

I prefer:
```typescript
 keybinding: { synctex: configuration.get('view.pdf.internal.synctex.keybinding') }
```

tamuratak:
It seems reasonable.

domoritz:
Do you think you could support both using an array instead of a single value?

James-Yu:
It seems that reverse synctex fails after the viewer is refreshed by a successful compilation. Any idea? @trevorgunn

PR #1633, opened 2019-08-22
stop users sending bug reports to the PDF.js dev team.

tamuratak:
Because it is difficult for users to distinguish bugs of LaTeX Workshop from the ones of PDF.js, make users report bugs only. We should prevent users from sending false bug reports to the PDF.js dev team and annoying them.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -20,8 +20,6 @@ You still see this issue?: Yes/No
 
 **If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.**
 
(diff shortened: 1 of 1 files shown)


James-Yu:
I'd say a disclaimer is good to reduce noise, though pointing to `pdf.js` is a bit pre-mature. What about changing the description to the follows?
```
**If your issue is with the internal pdf viewer not displaying something correctly, take the pdf and open it in Firefox and check if the same issue exists there. If so, the problem is mostly likely not with this extension.**
```

tamuratak:
There is no reason to discourage users from reporting a bug related to the pdf viewer. It is a kind of bug we have to treat. To my experience, false bug reports of the pdf viewer's displaying are rare.

It would be more helpful to require users to upload the pdf file and the TeX file.

PR #1591, opened 2019-08-08
Fixes: Wrong File Name When Parsing The LaTeX Log File (issue #1566)

sdhdez:
It is a change in the regular expressions (`pathResult` and `mikTeXPathResult`) to retrieve the correct path to files.
Issue #1566

Issue #1566, opened 2019-08-01
Wrong File Name When Parsing The LaTeX Log File


uninyhart:
**Describe the bug**

The extension reports the problem with a wrong file name.

**Condition**

Input a file which needs >=2 pages and has a warning.

**Desktop**
 - OS: Windows 10
 - VS Code version: 1.36.1
 - Extension version: 7.3.0

**Example**

```latex
% This is main file
\documentclass{article}

\newcommand{\reportwarning}{\noindent \\} % Just a trick to report info

\usepackage{lipsum}

\begin{document}

\input{1.tex}

\end{document}
```

```latex
% This is 1.tex
\lipsum % Print many pages

\reportwarning
```

And then compile it.

It will report a problem.

Here is some related log info in `log` file

```
(./1.tex [1

] [2]
LaTeX Font Info:    External font `cmex10' loaded for size
(Font)              <7> on input line 6.
LaTeX Font Info:    External font `cmex10' loaded for size
(Font)              <5> on input line 6.
LaTeX Font Info:    External font `cmex10' loaded for size
(Font)              <8> on input line 6.
LaTeX Font Info:    External font `cmex10' loaded for size
(Font)              <6> on input line 6.

Underfull \hbox (badness 10000) in paragraph at lines 6--7

 []

)
```

And the bug:

The file name is `1.tex [1] [2]`.

---

P.S. It is quite annoying when debugging a huge document because I cannot directly goto that line in the file.

Thanks a lot.

tamuratak:
Please show us log messages. 

![2018-11-07 20 34 24](https://user-images.githubusercontent.com/10665499/48129417-c4565c80-e2cc-11e8-89de-05b36ef1fa76.png)

Thank you for your cooperation.

uninyhart:
Compile command is

```
latexmk -synctex=1 -interaction=nonstopmode -file-line-error -xelatex main.tex
```

The log of *extension* looks pretty normal.

Here is some related message in the LaTeX Workshop Log:

```
LaTeX log parsed with 1 messages.
```

The problem pops when parsing the log of *LaTeX*, the `*.log` file.

Here is some info in the `main.log` file:

```
 (./1.tex [1

]
Underfull \hbox (badness 10000) in paragraph at lines 4--5

 []

) 
```

And then the extension parsing it as (the VSCode problem message)

```json
{
	"resource": "1.tex [1]",
	"owner": "LaTeX",
	"severity": 2,
[...]

PR #1490, opened 2019-07-08
Allow all printable ASCII characters except ",(,) in paths.

trevorgunn:
Fixes #1487 

Attempt number 2. Allowing parentheses and " produced garbage paths the last time I tried this.

I've tested this for a while on both Windows (but TeXLive, not MikTeX) and Linux and haven't seen anything matched by the regex that shouldn't have been.

This requires more testing I think.

Edit: I've tested this with MikTeX now on a large-ish file with a lot of path information contained within it (mostly LaTeX packages and stuff). From what I've seen, as long as `"` and `)` aren't included in the regex, this should work.

PR #1652, opened 2019-08-28
Disallow [ and ] in paths

trevorgunn:
Related: #1490 and #1591 

Pdftex outputs some lines like `./1.tex [1{/var/lib/texmf/fonts/map/pdftex/updmap/pdftex.map}])`

Forbidding `[` and `]` is required to correctly pick up `./1.tex`, otherwise it will match all the way to the end and cause errors.

Previously, we allowed everything except non-printable ascii and `()"`. I've changed the regex group to a negative match so it is easier to read. This will allow non-printable ascii again but I don't think that's going to be a problem.

This will fix the issue that uninyhart had in #1566 for xelatex, but not the same issue for pdflatex as I will try to describe shortly.

### DIFF
--- a/src/components/parser.ts
+++ b/src/components/parser.ts
@@ -266,8 +266,8 @@ export class Parser {
         if (result && result.index !== undefined && result.index > -1) {
             line = line.substr(result.index + 1)
             if (result[1] === '(') {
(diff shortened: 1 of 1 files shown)


trevorgunn:
Previous set: ` !\u0023-\u0027\u002A-\u00FE\u00a1-\uffff`

So this excludes:

- `u0000` to `u001F` (non-printable),
- `u0022` - `"`,
- `u0028` - `(`,
- `u0029` - `)`,

`u00FE` comes after `u00A1`, which I'm just now noticing, so the above list is everything that was excluded before now.

Issue #1639, opened 2019-08-25
Add option to close document in Adobe Reader before compilation (or prevent compilation when output file is not writable)


bersbersbers:
**Is your feature request related to a problem? Please describe.**
I use auto-compile on save, and often I proof-read the compiled document in Adobe Reader. Saving any change to the tex file then triggers compilation, which fails with "I can't write on file \jobname.pdf". That wouldn't be too bad of a problem, but that error usually triggers a complete rebuild (involving some 4-5 iterations of pdflatex) of the whole document.

**Describe the solution you'd like**
I'd like a solution where the file is closed in Adobe Reader, compiled, and then reopened after compilation - ideally at the same scroll position. TeXstudio allows doing that through DDE commands.

**Describe alternatives you've considered**
Alternatively, LaTeX workshop could make sure that the output file is writable before starting compilation; if it is not, there is no point in starting a compilation process. (The longer I think about this, the more this is my preferred option. Also, I believe this is much more easy to implement.)

**Additional context**
n/a

James-Yu:
Thanks for the suggestion. It is not practical to implement pdf-reader-closing actions as there are too many kinds of such readers. Yet checking whether the file is writable is a good improvement.

James-Yu:
Note that this check does not work due to limits on `fs` of node. We cannot fix that.

Issue #1662, opened 2019-08-30
Latex Workshop cannot recompile after first compilation, output pdf file is locked


lucblassel:
**Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue.**

You still see this issue?: Yes

**If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.**  
I can compile manually by calling the docker container with the following command:  
`docker run -w "$(pwd)" -v "$(pwd):$(pwd)" blang/latex:ctanfull latexmk -synctex=1 -interaction=nonstopmode -file-line-error -pdf filename`

**Describe the bug [Required]**
After the first successful build which, when re-building changes made to the document, an error is shown sating that the .pdf file is locked, and the compiler is not called. Any additional calls to the BUILD command is unsuccessful and doesn't show any additional error messages in the notifications. 

**To Reproduce**
1. Create a new folder and a new  dummy .tex document and save to build:
````tex
\documentclass{article}
\begin{document}
    Hello World.
\end{document}
````
2. Modify document and save again additional times. 
3. After the first modification, build will fail with the following error in the notification tray: `/home/lblassel/latex_workshop_test/latex.pdf is locked. Is it opened in other applications?`
4. Any subsequent modifications and saves to build fail without notification, and the compiler is not even called.

**Expected behavior**
The tex document is recompiled and the changes appear in the updated pdf file.

**Logs [Required]**

<details>
<summary>LaTeX Workshop Output [Required]</summary>

````
[13:57:02] Initializing LaTeX Workshop.
[13:57:02] Creating LaTeX Workshop http and websocket server.
[13:57:03] LaTeX Workshop initialized.
[13:57:03] Found root file from active editor: /home/lblassel/latex_workshop_test/latex.tex
[13:57:03] Root file changed from: undefined to /home/lblassel/latex_workshop_test/latex.tex. Find all dependencies.
[13:57:03] Instantiating a new file watcher for /home/lblassel/latex_workshop_test/latex.tex
[13:57:03] Creating file watcher for .bib files.
[13:57:03] Parsing /home/lblassel/latex_workshop_test/latex.tex
[13:57:03] Server created on 127.0.0.1:44644
[13:57:03] Adding /home/lblassel/latex_workshop_test/latex.tex to file watcher.
[13:57:03] Snippet data loaded.
[13:57:03] LaTeX Workshop version: 8.1.0
[...]

PR #1668, opened 2019-09-03
Stop checking locks unnecessarily

tamuratak:
Stop checking locks unnecessarily. Related to #1662.

We do not have to do this on our extension side.

### DIFF
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -156,19 +156,6 @@ export class Builder {
             return
         }
         const releaseBuildMutex = await this.preprocess()
-        let locked = false
-        try {
-            if (fs.existsSync(this.extension.manager.tex2pdf(rootFile))) {
-                fs.accessSync(this.extension.manager.tex2pdf(rootFile), fs.constants.W_OK)
-            }
-        } catch (err) {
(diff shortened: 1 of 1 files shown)


trevorgunn:
Also related to #1639 . The lock check was added in 8.0.7 and the user claimed the problem exists in other versions of 8.x.x although that error message certainly only exists in ^8.0.7.

There are other methods to solve #1639, such as having the user set `latex.autoBuild.cleanAndRetry.enabled = false` or testing if latexmk/pdflatex outputs "I can't write on file \jobname.pdf".

tamuratak:
No, we cannot fix this problem. As @James-Yu  commented in #1639, we cannot fix #1639 on our side because of the restriction of Node.js API on Windows.

If we cannot solve the problem, we should not touch it on our side.

trevorgunn:
I'm referring to this:

> That wouldn't be too bad of a problem, but that error usually triggers a complete rebuild (involving some 4-5 iterations of pdflatex) of the whole document.

I'm saying that it is possible to avoid the clean-and-rebuild without having to check if the file is writable.

tamuratak:
I get the point. We do not have to do anything if the user can fix it by the setting.

PR #1679, opened 2019-09-07
add lualatex to latex-workshop.latex.tools

garykwtsui:
This is aimed to allow users to configure lualatex as part of their latex-build-recipe.  Info about `lua{la}tex`: http://www.luatex.org

When this is properly configured in users' recipes, this will allow latex workshop to compile tex that uses luacode (e.g. `\usepackage{luacode}`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -636,6 +636,7 @@
               "args": [
                 "-synctex=1",
                 "-interaction=nonstopmode",
+                "-file-line-error",
                 "%DOC%"
               ],
               "env": {}
--- a/package.json
+++ b/package.json
@@ -629,6 +629,16 @@
                 "%DOCFILE%"
               ],
               "env": {}
+            },
+            {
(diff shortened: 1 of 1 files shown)


trevorgunn:
Adding a new tool doesn't really hurt but isn't it just as easy for users  to add the `-lualatex` flag to the `latexmk` tool?

Also, I believe you should add the `-file-line-error` flag although I'm not 100% sure why. (Something to do with the parser with multifile projects.)

garykwtsui:
@trevorgunn good catch, sure, i can add `-file-line-error`.   I'm with you on the possibility of adding the extra flag to `latexmk`; however, the initial thought of this change was to provide some _default_ options for new adopters.

jlelong:
I do not think adding a `lualatex` tool will really help. It would be far better to add a `latexmk` tool with the `-lualatex` option.

Issue #1692, opened 2019-09-11
Package documentation suggestions


cojua8:
Hi!,

I use package documentation a lot, so I'd like the "Show package documentation" to suggest the packages called in the document, instead of writing the names every time.

Also, I noticed that when you write a wrong package name, the command opens all previous documentation opened. For example, I was looking at subcaption and hyperref documentation, and tried to open an inexistent hyperreff, both subcaption and hyperref docs were opened again

cojua8:
I understand the problem, and as user I'd prefer a selection list, since I usually want to see documentation of already loaded packages, and look for new packages on google.
Of course, this is just my opinion,

jlelong:
@cojua8 a new command has been introduced for that `latex-workshop.texdocUsepackages`. You can call through the command palette with _Show package documentation actually used_.

PR #1712, opened 2019-09-23
Texdoc completion

tamuratak:
Related to #1692.

- use `showQuickPick` for texdoc completion.
- read `data/packagenames.json` for completion items.
- introduce a new command `Show package documentation actually used`

### DIFF
--- a/package.json
+++ b/package.json
@@ -296,6 +296,11 @@
         "category": "LaTeX Workshop"
       },
       {
+        "command": "latex-workshop.texdocUsepackages",
+        "title": "Show package documentation actually used",
+        "category": "LaTeX Workshop"
+      },
+      {
         "command": "latex-workshop.showCompilationPanel",
         "title": "Show LaTeX Compilation Info",
         "category": "LaTeX Workshop"
--- a/src/components/texdoc.ts
+++ b/src/components/texdoc.ts
@@ -12,21 +12,21 @@ export class TeXDoc {
 
     constructor(e: Extension) {
         this.extension = e
-        setImmediate( () => this.packageItems = this.readPackages() )
+        setImmediate( () => this.readPackages() )
     }
 
     private readPackages() {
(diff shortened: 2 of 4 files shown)


src/components/texdoc.ts
```diff
@@ -0,0 +1,120 @@
...
+            }
+        })
+    }
+
+    texdoc(pkg?: string) {
+        if (pkg) {
+            this.runTexdoc(pkg)
+            return
+        }
+        vscode.window.showQuickPick(this.packageItems).then(selectedPkg => {
```
jlelong: I believe we should stick to the `showInputBox` approach here and we should not restrict ourselves to the packages listed in `packagenames.json`.

package.json
```diff
@@ -295,6 +295,11 @@
         "title": "Show package documentation",
         "category": "LaTeX Workshop"
       },
+      {
+        "command": "latex-workshop.texdocUsepackages",
+        "title": "Show package documentation actually used",
+        "category": "LaTeX Workshop"
+      },
```
jlelong: Good idea to add an extra function to access the documentation of the included packages. This solves #1692.

jlelong:
We should keep in mind to document this in the wiki too.

src/components/texdoc.ts
```diff
@@ -0,0 +1,118 @@
...
+            const pkgs = content && content.element.package
+            if (!pkgs) {
+                continue
+            }
+            names = names.concat(pkgs)
+        }
+        const packagenames = Array.from(new Set(names))
+        for (const name of packagenames) {
+            let item: vscode.QuickPickItem | undefined = { label: name }
+            const pkg = this.pkgs[name]
```
jlelong: I do not think there is a need to get information from `data/packagenames.json`. I guess users know exactly which package documentation they want to look at. I think this part of the code can be removed along with the `readPackages` method.

PR #1725, opened 2019-09-28
Make version numbers in changelog links

tecosaur:
This just makes the version numbers in the changelog link to the diffs. For an example see https://github.com/tecosaur/LaTeX-Utilities/blob/master/CHANGELOG.md it's minor but it's nice

New versions will obviously need similar lines added.


jlelong:
Not sure this is really useful as only advanced users may actually be interested in diffs between releases and for sure they can access directly using `git`.

tecosaur:
Eh. I figured why not. It's a bit nice, and I can't see any drawbacks. Not a big deal either way.
Personally I like being able to just click on a version in the changelog and see all the changed made.

jlelong:
I will leave this to @James-Yu as he is the one who actually writes the `README.md` file.

tecosaur:
@James-Yu sorry to bother you again but also, \*cough

James-Yu:
Thanks for the ping. I consider this is a good addition, yet I personally do not have interest in maintaining such a git diff link when newer versions are published. Any automation tools for this task?

tecosaur:
For keeping something like this up to date, what comes to mind is something like this:

https://github.com/github-changelog-generator/github-changelog-generator
or
https://github.com/conventional-changelog/standard-version
or
https://github.com/CookPete/auto-changelog

PR #1732, opened 2019-09-30
use @types/pdfjs-dist

tamuratak:
use `@types/pdfjs-dist`


tamuratak:
`@types/pdfjs-dist` is not enough matured. should be pending.

PR #1751, opened 2019-10-14
Add option to display progress as a notification and refactor buildinfo.ts

### LINKED ISSUE #1684: Use native progress notification
Currently, LaTeX Workshop shows progress using the status text + a bespoke progress bar. 

VS Code supports progress notifications via `vscode.window.withProgress`, so it would be a nice thing to implement them.

There are no alternatives, to the best of my knowledge.

**Additional context**

N/A

trevorgunn:
Closes #1684 

As discussed in #1684, the native API really only matters if the progress is being displayed as a pop-up notification. 

With the pop-up notification, we can use the native progress bar (the current progress bar is displayed simultaneously, I don't know if we want to change that or not). We also have access to a cancel button, which I've set to the `builder.kill` function.

I have also done some unrelated refactoring.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1402,6 +1402,14 @@
             "█████▙░░░ (4 levels per block)",
             "No Bar"
           ]
+        },
+        "latex-workshop.progress.location": {
+          "type": "string",
+          "default": "Status Bar",
+          "enum": [
+            "Status Bar",
+            "Notification Dialogue"
+          ]
         }
       }
     },
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -160,9 +160,30 @@ export class Builder {
         this.extension.logger.displayStatus('sync~spin', 'statusBar.foreground')
         this.extension.logger.addLogMessage(`Build root file ${rootFile}`)
         try {
-            this.extension.buildInfo.buildStarted()
-            pdfjsLib.getDocument(this.extension.manager.tex2pdf(rootFile, true)).promise.then((doc: any) => {
-                this.extension.buildInfo.setPageTotal(doc.numPages)
+            const configuration = vscode.workspace.getConfiguration('latex-workshop')
+            let location: vscode.ProgressLocation
+            let title: string
+            if ((configuration.get('progress.location') as string) === 'Status Bar') {
+                location = vscode.ProgressLocation.Window
+                title = ''
+            } else {
+                location = vscode.ProgressLocation.Notification
+                title = 'Running build process'
+            }
+            vscode.window.withProgress({
+                location,
+                title,
+                cancellable: true
+            }, (progress, token) => {
+                token.onCancellationRequested(this.kill.bind(this))
+                this.extension.buildInfo.buildStarted(progress)
+                pdfjsLib.getDocument(this.extension.manager.tex2pdf(rootFile, true)).promise.then((doc: any) => {
+                    this.extension.buildInfo.setPageTotal(doc.numPages)
+                })
+                const p = new Promise(resolve => {
+                    this.extension.buildInfo.setResolveToken(resolve)
+                })
+                return p
             })
             // Create sub directories of output directory
             // This was supposed to create the outputDir as latexmk does not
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1402,6 +1402,14 @@
             "█████▙░░░ (4 levels per block)",
             "No Bar"
           ]
+        },
+        "latex-workshop.progress.location": {
+          "type": "string",
+          "default": "Status Bar",
+          "enum": [
+            "Status Bar",
+            "Pop-up Window"
```
tecosaur: Rename to "Dialogue" maybe?

src/components/builder.ts
```diff
@@ -160,9 +160,30 @@ export class Builder {
...
         this.extension.logger.addLogMessage(`Build root file ${rootFile}`)
         try {
-            this.extension.buildInfo.buildStarted()
-            pdfjsLib.getDocument(this.extension.manager.tex2pdf(rootFile, true)).promise.then((doc: any) => {
-                this.extension.buildInfo.setPageTotal(doc.numPages)
+            const configuration = vscode.workspace.getConfiguration('latex-workshop')
+            let location: vscode.ProgressLocation
+            let title: string
+            if ((configuration.get('progress.location') as string) === 'Status Bar') {
+                location = vscode.ProgressLocation.Window
```
tecosaur: Two issues with using the status-bar progress report:
 1. Adds a redundant spinner (too many icons)
image
 2. This spinner resets every time a new page is pushed out, which makes it look glitchy

src/components/buildinfo.ts
```diff
@@ -421,14 +424,17 @@ export class BuildInfo {
...
-            this.status.text = `${runIcon}  Page ${padRight(
-                currentAsString + endpointAsString,
-                this.currentBuild.pageTotal ? this.currentBuild.pageTotal.toString().length * 2 + 2 : 6
-            )} ${barAsString}`
+            this.progress.report({
+                message: `${runIcon}  Page ${padRight(
+                        currentAsString + endpointAsString,
+                        this.currentBuild.pageTotal ? this.currentBuild.pageTotal.toString().length * 2 + 2 : 6
+                    )} ${barAsString}`,
+                increment: this.currentBuild.pageTotal ? current / this.currentBuild.pageTotal * 100 : undefined
```
tecosaur: `this.currentBuild.pageTotal ?` gives `false` when it is `0`, however `0` is a valid value. I'd suggest using `!== undefined` or similar.

trevorgunn: That is to avoid division by zero. I do not think we can do anything better.

tecosaur:
Does a negative increment work?

src/components/buildinfo.ts
tecosaur: Hmm.
 - if `current === 0` you want `0`
 - if `this.currentBuild.pageTotal > 0` you want `1 / this.currentBuild.pageTotal`
 - otherwise undefined

This seems to be satisfied by
```js
this.currentBuild.pageTotal ? 1 / this.currentBuild.pageTotal :
  current === 0 ? 0 : undefined
```

src/components/builder.ts
trevorgunn: I can remove the spinner from the logger status bar item, I think we're stuck with the spinner on the progress bar and I don't think I can do anything about it resetting.

tecosaur: Erm, surely if it works as a status just leave it as a status?

trevorgunn: That is probably best, the more I think about it and work on it. The native progress API only offers improvement for notifications, for the status bar the API only allows us to display a message (which we can already do).

trevorgunn:
While working on this, and thanks in part to @tecosaur , I have decided not to use the native API for the status bar, because that does not give us any functional benefit over the old status bar item and additionally looks worse.

So now this PR adds the option of using the native API for a pop-up notification as a separate option but retains the old functionality.

tecosaur:
This seems good to me. Anyone else have some thoughts?
@tamuratak @jlelong

tecosaur:
Since this seems fine to me and no one else has complained I'll just merge this.

PR #1772, opened 2019-10-27
Add command to call bibtex-tidy

### LINKED ISSUE #1761: Add bibtex-tidy support to format and sort .bib files
[bibtex-tidy](https://github.com/FlamingTempura/bibtex-tidy) is a tool that can tidy bibtex files (alignment, fields sorting etc.)

tidy

It would be great to have this built-in.

trevorgunn:
Resolve #1761


jlelong:
Your PR looks fine to me. 
However, the current approach is to call `bibtex-tidy` with the full configuration each time resulting in a full tidy. It would nice to be able to call the different actions separately, at least sort and align.

trevorgunn:
@jlelong It looks like BibTeX-tidy will always: 1) change the fields to lowercase, 2) align the fields. I've added a command that does as little as possible and one that does that plus sort the entries (by key). 

These new commands pull the relevant config values from the common config item (`latex-workshop.bibtex-tidy.options`).

jlelong:
Hum, it would be great to be able to trigger one action at a time (I am pretty sure this will create false bug reports and/or user headaches). I can think of two solutions:
1. Get in touch with the author of `bibtex-tidy` to see if he is willing to implement this. (but the last update is already 4 months old).
1. `bibtex-tidy` acts as a frontend on [`bibtex-parse`](https://github.com/FlamingTempura/bibtex-parse) (by the same author), which actually parses `bibtex` entries. Maybe, we could directly use `bibtex-parse` to keep control on how we apply the different actions.

I think 2. would make more sense. @tamuratak @James-Yu any comment on that?

trevorgunn:
It looks like `bibtex-parse` doesn't keep the original formatting either so I don't think that helps us. Also if `bibtex-tidy` is built off of `bibtex-parse` then there isn't any way for the author to implement non-reformatting sorting either.

I think having a short tidy function using `bibtex-tidy` is a good feature to have. Anything more complicated is better served using a dedicated bibliography management software like JabRef or others.

tamuratak:
It is not usual keeping the alignment of each line when parsing. Although it is possible, I am not sure the author likes it.  Since `latex-utensils` keeps it, `latex-utensils` is more appropriate if you directly use `bibtex-parse`.

In my opinion, we do not have to provide `sort without align`. `Sort and align` and `align (only)` seem feasible.

tecosaur:
On a relevant note, LaTeX bibliography seems to be mostly using/shifting towards `biber` (as far as I can tell). Personally I have found the following command useful
```bash
$ biber --tool --output_align --output_indent=2 --output_fieldcase=lower $file -O $file
```

trevorgunn:
I need to remember to access `latex-workshop.message.warning.show` before displaying the warning message. I will put this in once I know what commands people want.

jlelong:
@tamuratak Good point. I did not notice that your package `latex-utensils` already parses `bibtex` files. As we already use it for other purposes in the extension, we should think of it twice before addind an extra dependency for parsing `bibtex` files, especially when we do not know much about the package maintenance. 

I guess `sort` and `align` could be very easily implemented using `latex-utensils` and they are probably the most used actions. The other actions look far less useful (at least for me).

Maintenance is indeed very important to me so I may overreact on this. Your thoughts on implementing an in-house solution for sorting and align bib entries?

tamuratak:
I have found that treating comments in a BibTeX file properly is not straightforward. At least, with #1780, all the comments will be lost.

On the syntax of comments in a BibTeX file, please refer to the following.

> New entries always start with @. Anything outside the “argument” of a “command” starting with an @ is considered as a comment. This gives an easy way to comment a given entry: just remove the initial @. As usual when a language allows comments, don’t hesitate to use them so that you have a clean, ordered, and easy-to-maintain database. Conversely, anything starting with an @ is considered as being a new entry

p. 20 [TameTheBeast – A manual about bibliographies and especially BibTeX](https://ctan.org/pkg/tamethebeast/)

trevorgunn:
@tamuratak I do not understand what you mean by "all the comments will be lost." See below for instance.

gif2

tamuratak:
I was wrong. I have missed the behavior.

jlelong:
Based on @tecosaur's comment, I would be in favour of relying on `biber` for formatting `bibtex` files. We could just declare a new formatting tool for the `bibtex` language. This would solve any interrogation on the maintainability of a third-party `js` library.

Sorting and aligning could be available through #1780.

jlelong:
Considering the different comments on both this PR and #1780, I suggest:

- to merge #1780 
- to close this #1772
- to implement `.bib` files formatting using `biber` instead of `latexindent.pl`.

Any comments on that?

trevorgunn:
@jlelong I agree with merging #1780 and closing this one. I don't think the question about how we handle aligning is 100% figured out, since #1780 also adds that feature. There are now 4 possible alignment options: `biber`, `latexindent`, `bibtex-tidy` and what is in #1780.

The one in #1780 is quite basic in what it does so having something more involved like `biber` beside it I think is ok. Some users (such as myself) don't want the alignment function to do much more than put spaces in the appropriate places. `biber` does a lot more such as expand abbreviations, remove comments, sorts fields, deletes unnecessary fields. I think there is value in having both options.

trevorgunn:
Wait, does `latexindent.pl` indent `.bib` files? Because it looks like it formats `.tex` files instead.

jlelong:
Yes `latexindent.pl` operates on `.bib` files, see [here](https://github.com/cmhughes/latexindent.pl/issues/65) but I do know how helpful it can be. It seems it has been primarily designed for `.tex` files.

PR #1740, opened 2019-10-10
embed source.r syntax in R code blocks for .Rnw documents

armarti:
Added syntax highlighting for R code blocks in [Rnw documents](https://yihui.name/knitr/demo/minimal/). Tested with Yuki Oeda's [Ikuyadeu.r](https://marketplace.visualstudio.com/items?itemName=Ikuyadeu.r).

This is a screenshot from my editor:
image

PR #1777, opened 2019-10-31
Declare a new language id for .rnw documents

jlelong:
This is a follow up to #1740 
@armarti and @ChristopheRoyer21 Can you checkout that branch and test it with `.rnw` documents? In particular, I am afraid that the outline mechanism (at the bottom of the TeX panel) crashes.


armarti:
I'm just browsing this on my phone, but @jlelong could you attach a screenshot of what you mean by the outline mechanism? 

I have a bunch of extensions dealing with highlighting so just want to make sure I see the same thing.

jlelong:
The extension does a lot more than syntax highlighting, see the [wiki](https://github.com/James-Yu/LaTeX-Workshop/wiki) for a list of features. In particular, the extension can display the table of contents (commonly called structure or outline in VSCode). It requires parsing the `.tex` files and I am afraid the extra bit of syntax in `.rnw` files breaks it.

Capture d’écran 2019-11-01 à 19 51 54

ChristopheRoyer21:
I personally see no problem with this, but I think it's simply because there are no .rnw documents in my project. How I usually work, is that I do my .rnw separately, I compile them and I include them in my main file with an \input{} as a .tex file after it had been compiled. 

If the inclusion of .Rnw syntax breaks the outine mechanism, maybe the definition of a new language is not the solution... I say that, but I have no experience in extension development so you might be able to find something to fix that easily!

jlelong:
The changes of this PR only apply to `.rnw` files, whose purpose is to use a modified adapted LaTeX syntax highlighting for editing `.rnw` files.

Can you post a complete `.rnw` example along with the generated `.tex` file so that I can test the PR? If this PR is to be merged, I do want to leave it open for too long.

ChristopheRoyer21:
test.rnw
test.tex

Here is a simple example. I simply create a graph in an .rnw file. When I compile it into a .tex file, the code is run and a file (most likely *.pdf) containing the output of the code is generated.  The first file is a test.rnw and he second one is test.tex. 
Of course, this is not a stand alone project, but when I want to use the *.rnw file in a main file, I simply use something like " \include{test}" or "\input{test.tex}". Note that when I include the file, I use the extension ".tex" instead of ".Rnw", which means you have to compile your rnw before compiling your whole project. I usually use a makefile to compile everything at the same time, but the firs command it uses is _R CMD Sweave --encoding="utf-8" test.Rnw_ (it directly ompie the .rnw files to .tex)

jlelong:
Could you post the file contents (`.rnw` and `.tex`) directly instead of a screenshot? I need the original files to do some tests. Thanks for your help.

ChristopheRoyer21:
[testrnw.txt](https://github.com/James-Yu/LaTeX-Workshop/files/3800981/testrnw.txt)
[testTex.txt](https://github.com/James-Yu/LaTeX-Workshop/files/3800982/testTex.txt)

Here are the files. The first one if the rnw file and the second one is the tex file. I had to change the extension to .txt to attach them.

armarti:
Just getting to this now. I guess for the structure to fill out in the current extension, you need a \section{} somewhere in your document? I opened the tex file @ChristopheRoyer21 posted (which has a \section) and it still doesn't fill out. Sorry I've just never used this feature.

@jlelong would you mind posting the tex file from your screenshot?

armarti:
*Actually* I see the outline in the Outline section of the workspace panel, but not in the Structure section of the TeX panel. Is that right?

Essentialy I'm seeing the same results for a plain tex file with and without this branch's changes, and agreed nothing is displayed in any outline section for rnw files.

Below are some screenshots for everyone's reference.

Plain tex file (current extension is in use on the left in these screenshots, and this branch is in use on the right):
Screenshot from 2019-11-07 08-57-02
Screenshot from 2019-11-07 08-56-39

Rnw file (this branch only):
Screenshot from 2019-11-07 09-02-40

jlelong:
I currently do not have much time but I am still on it! Here is the current state

- The `rnw` branch has no impact on `.tex` files. So there is no harm.
- Syntax highlighting works well for `.rnw` files but detecting their associated [root files](https://github.com/James-Yu/LaTeX-Workshop/wiki/Multi-File-Projects#the-root-file) does not work, which in turn breaks a lot of the extensions features such as advanced intellisense, outline, ... I will try to fix this.

jlelong:
Detecting root files for `.rnw` files does work (there was no `\begin{document}` in the previous example.

Consider the following `.rnw` file content
````latex
\documentclass{article}
\def\toto{arg}

\SweaveOpts{prefix.string=figures}

\begin{document}
    \section{test for rnw to tex}

    <<import-packages, echo=false>>=
    library(tidyverse, quietly = T)
    library(factoextra)
    library(FactoMineR)
    library(xtable)
    library(caret)
    @

    some text here


    \begin{figure}
    <<mca-graph, fig=true, echo=false>>=
    plot(1:1000, rnorm(1000))
    @
    \caption{some graph}
    \label{fig:mca-data-perso}
    \end{figure}
\end{document}
````

Capture d’écran 2019-11-07 à 21 44 12

Outline is working fine but not advanced intellisense, which relies on parsing the file content.

jlelong:
I guess everything is working fine now. 
Could you check with some more advanced examples before I merge the PR ?

jlelong:
Up! Anyone to check on a real life `.rnw` document or at least provide such as document?

armarti:
Sorry these weeks are mental for me. I've zipped up a bunch of rnws. The answers are probably wrong but these compile fine in RStudio.
[rnws.zip](https://github.com/James-Yu/LaTeX-Workshop/files/3837680/rnws.zip)

armarti:
This is great. Thank you.

PR #1780, opened 2019-11-01
Add commands to sort and align bibtex files

trevorgunn:
This was talked about in #1772. This isn't all the features we want yet (I think) but I think it is also helpful to know what the function will look like.

The logical structure is

1. Parse the file using `latex-utilities`
2. Create an array of initial locations
3. Sort the entries and create an array of final locations
4. Successively copy the sorted text to the original positions

### DIFF
--- a/package.json
+++ b/package.json
@@ -1446,7 +1446,7 @@
             "left": "{",
             "right": "}",
             "case": "lower",
-            "sort": "key"
+            "sort": ["key"]
           },
           "properties": {
             "tab": {
@@ -1470,9 +1470,9 @@
               "default": "lower"
             },
             "sort": {
-              "description": "What to sort by. Either the entry \"key\", \"author\" or \"title\".",
-              "type": "string",
-              "default": "key"
+              "description": "An array of strings to sort by. Either a bibtex field name (title, author, year, etc.) or \"key\" for the entry key. E.g. [\"author\", \"year\", \"title\"].",
+              "type": "array",
+              "default": ["key"]
             }
           }
         }
--- a/package.json
+++ b/package.json
@@ -1438,6 +1438,43 @@
             "Status Bar",
             "Notification Dialogue"
           ]
+        },
+        "latex-workshop.bibtex-format": {
+          "type": "object",
+          "default": {
+            "tab": "  ",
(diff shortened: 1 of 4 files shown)


jlelong:
It looks great. 
Maybe, we could make the sorting key configurable (via a configuration variable). I assume that adding the `align` action should straightforward too.

trevorgunn:
Align and sort have now been implemented. Sorting and aligning takes about 20-30 ms on both of my development machines on a 244 line file with 15 entries.

Compared to `biber`, this alignment function:
- does not sort fields (which I think is ok)
- does not remove comments (which I think is better)
- does not simplify abbreviations (`@string{...}`) or concatenations (which I think is better)

In terms of options, I'm thinking
- Sort by author, key or title (sorting by date might make sense as well but it's too messy because you have to compare months as well)
- Surround fields by `{...}` or `"..."`
- Upper or lower case field names (e.g. `AUTHOR` or `author`)

jlelong:
This is really great.

For sorting, we could accept a list of keys and no need to worry about months, only considering the `year` filed would be enough. I agree with the options 
- `{....}` to `"...."` and vice versa
- Lower or upper case field names.

sushshring:
I like the implementations at [FlamingTempura/Bibtex-Tidy](https://github.com/flamingtempura/bibtex-tidy) (Also see [website](https://flamingtempura.github.io/bibtex-tidy)). There are a lot of configurable options along with useful tools. Would love to see some of those included with LaTex-Workshop

jlelong:
@trevorgunn Since the `latex-workshop.bibtex-format` configuration variable is a complex object it is not directly editable inside VSCode's setting pane

Capture d’écran 2019-11-09 à 07 20 03

One has to edit your `settings.json` file manually and does not have access the different fields documentation. I think it would be best to split this complex object into elementary configuration items directly editable in VSCode's setting pane

trevorgunn:
@jlelong You can access documentation in `settings.json` by hovering over a given key. But you are right about using more elementary configuration items.

jlelong:
Good point @trevorgunn. I had never noticed the hovering trick in `settings.json`.

PR #1798, opened 2019-11-12
Tweak section level shift

tecosaur:
This mostly has no effect on functionality, so I'll just go through my reasoning behind the changes.

- Allow for section level shift for blank title
  - I've experienced the scenario where I type `SPG 🡒 \paragraph{}` and then think, actually this should be a sub-section.
  - Can't think of a good reason not to allow this
- Rename section shift increment/decrement to promote/demote
  - To me this naming makes more sense, it's not a counter
- Switch section shift keybindings to follow meaning of default indent/deindent keybindings
  - VScode has 
    - `ctrl`+`[` to indent, increases 'importance'/'priority' of line
    - `ctrl`+`]`to de-indent, decreases 'importance'/'priority' of line
  - I recently reconsidered they default keybindings I set for this (after instictively using these keybindings the wrong way around), and came to the conclusion that I got it the wrong way around
      - keybindings should follow the same 'meaning' as VSCode's indentation keybindings
      - `ctrl`+`alt`+`[` should _increace_ priority of section level
      - `ctrl`+`alt`+`]` should _decreace_ priority of section level
    - For affected users, this change is fairly easy to adjust to
- Add section shift to the command palette
  - I think this is a nice feature, perhaps this may help with discoverability

### DIFF
--- a/package.json
+++ b/package.json
@@ -382,25 +382,25 @@
         "key": "ctrl+l ]",
         "mac": "cmd+l ]",
         "when": "editorTextFocus && !editorReadonly && editorLangId == 'latex' && config.latex-workshop.bind.altKeymap.enabled",
-        "command": "latex-workshop.increment-sectioning"
+        "command": "latex-workshop.promote-sectioning"
       },
       {
         "key": "ctrl+l [",
         "mac": "cmd+l [",
         "when": "editorTextFocus && !editorReadonly && editorLangId == 'latex'&& config.latex-workshop.bind.altKeymap.enabled",
-        "command": "latex-workshop.decrement-sectioning"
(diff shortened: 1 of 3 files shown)


jlelong:
I agree with all this. Items 1. and 4. are good improvements.
I am always a little uncomfortable with these keybindings as it is impossible to trigger `cmd + alt + [` on an AZERTY Apple keyboard as the key `[` is already the combination of `shift+alt+(`. I do not if it happens with other keyboard layouts too.

jlelong:
I understand that this PR exchanges the actions of `ctrl+[` and `ctrl+]`. If it is merged, it should be made crystal clear in the release notes (Changelog) to avoid false bug reports.

Issue #1155, opened 2019-01-29
Allow to override build magic


nafur:
I'm working of a somewhat larger project (think book, thesis project, ...) and need need to integrate other stuff in my build process (collection and processing of data, building of externalized tikz pictures, ...). The whole build process is defined in a Makefile and I can build my project just fine by running `make` in the project root folder. Now I'd like to have my stuff build whenever I save a file, so I made `make` issue an incremental build which is reasonably fast. Then I created a *tool* called `make` and a recipe called `make` and overwrite the existing tools in this particular workspace.

This mostly worked, but I still had to work around some issues, mostly because the extension would be unable to determine the root file and then not start the build (for example for externalized tikz pictures or local packages and classes) so I had to add a hint for the proper main file to make that work at all. This sometimes also fails for other files, sometimes I'm not even sure why. Altogether this makes the whole thing pretty fragile and I spend a lot of time figuring out why the automatic build does not work.

Now for my actual proposal:
Please add a setting `latex-workshop.latex.buildCommand` and (if the setting is set) whenever any file that this extension feels responsible for is saved we run this command instead of this whole other magic.

jlelong:
See the external build command section in the [wiki](https://github.com/James-Yu/LaTeX-Workshop/wiki/Compile#external-build-command).

nafur:
Thank you very much!

smtrat:
I finally got to use this "in production" after the last release and unfortunately found an issue when the build is triggered by a file within a subfolder. The command is executed with cwd set to the folder of the "triggering file" (as set [here](https://github.com/James-Yu/LaTeX-Workshop/commit/6c04fffd57d06932a2bca55d437f17c0e2aae214#diff-c7f5bd618dbb862973c3204edf8120d6R41)). In addition the variables (like `%DIR%`) do not seem to work here, hence I have not been able to craft arguments that reliably direct `make` to the proper Makefile in the projects root folder.

My suggestion would be to simply change the cwd to the projects root directory (not necessarily `%DIR%`). This kind of aligns with the general intention of this whole setting, at least in my mind.

nafur:
Sorry for using the wrong account...

My colleague just made me aware of [tasks within vs code](https://code.visualstudio.com/docs/editor/tasks) and it somehow seems like reinventing the wheel here. Maybe we could simply execute a task instead?

[...]

PR #1817, opened 2019-11-17
Replace placeholders in external build arguments

### LINKED ISSUE #1816: External build arguments: no expansion of %DOC%
%DOC% is not expanded when given as an arg in the external build command:

```
"latex-workshop.latex.external.build.command": "lmc",
"latex-workshop.latex.external.build.args": ["%DOC%"]
```

My custom command `lmc` has "echo $1" for diagnostics, which returns "%DOC%" in the latex compiler output.

The log messages:

```
[11:52:37] BUILD command invoked.
[11:52:37] Build using the external command: lmc %DOC%
[11:52:37] LaTeX buid process as an external command spawned. PID: 9254.
[11:52:37] Build returns with error: 1/null. PID: 9254.
```

 - OS: mac 10.14.6
 - VS Code version: 1.40.1
 - Extension version: 8.4.1

trevorgunn:
Closes #1816.

I've changed `commander.ts` so that `buildWithExternalCommand` can only be called if the `rootFile` is defined. I didn't see a reason to allow calls to `buildWithExternalCommand` with an undefined `rootFile`.

### DIFF
--- a/src/commander.ts
+++ b/src/commander.ts
@@ -68,13 +68,13 @@ export class Commander {
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
         const externalBuildCommand = configuration.get('latex.external.build.command') as string
         const externalBuildArgs = configuration.get('latex.external.build.args') as string[]
+        if (rootFile === undefined && this.extension.manager.hasTexId(vscode.window.activeTextEditor.document.languageId)) {
+            rootFile = await this.extension.manager.findRoot()
+        }
         if (externalBuildCommand) {
             const pwd = path.dirname(rootFile ? rootFile : vscode.window.activeTextEditor.document.fileName)
             await this.extension.builder.buildWithExternalCommand(externalBuildCommand, externalBuildArgs, pwd, rootFile)
             return
-        }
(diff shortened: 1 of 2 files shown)


jlelong:
One of the reasons why I introduced the `buildWithExternalCommand` feature was indeed to build when the `rootFile` is not defined. See the discussion in #1155.

jlelong:
Maybe, we could replace placeholders only when the `rootFile` is defined. I guess VScode's tasks are precisely designed for that.

trevorgunn:
I see. In that case probably we only want to call the new `replaceArgumentPlaceholders` command if the `rootFile` is defined. I don't know if that will lead to issues.

jlelong:
I can't think of any situations in which it would lead to issues. I think replacing the placeholders when the `rootFile` is defined is a good improvement.

trevorgunn:
If we don't look for the `rootFile` before calling `buildWithExternalCommand` then it will always be `undefined` when the extension first loads. So we need to look for the `rootFile` first but we can still call `buildWithExternalCommand` even if we can't find a `rootFile`.

PR #1826, opened 2019-11-22
load (and cache) PDF preview directly from disk for \includegraphics completion

sqd:
### Why:
Currently, when \includegraphics requests completions that involve PDFs, these two things happen:

1. PDF is loaded and rendered to SVG
2. This SVG is converted into a data URL, which is passed into `vscode.CompletionItem.documentation` as a `vscode.MarkdownString`
3. VSCode has to parse and render this data URL, which is super slow and jerky.

The actual rendering of PDF into SVG and then into data URL takes around half a second for a typical PDF on a Macbook. But it takes around several seconds to over 10s for VSCode to render this data URL into an actual image for some reason, which also hangs the editor.

### What:
Added a (default) option to fix that. Users can still fallback to the current implementation.

### How:
Store the generated SVGs in actual files on disk (instead of data URLs) in temp dir. VSCode is able to load on-disk image files pretty quickly.

Each time a preview for a PDF is requested, we compare the mtime between the SVG on disk (if it exists) and the mtime of the PDF. And generate a new preview if the SVG has expired. Pretty straightforward.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -319,11 +319,6 @@
       "resolved": "https://registry.npmjs.org/@tamuratak/domstubs/-/domstubs-0.1.1.tgz",
       "integrity": "sha512-UvZYDOp/cpqNttuc6w6RxuoPusA1Alfy3i84dVgy2pPUQ+7fovk09IRuabCATnMZi4phmPFgUlO3z0Y6viSOfw=="
     },
-    "@types/base64-js": {
-      "version": "1.2.5",
-      "resolved": "https://registry.npmjs.org/@types/base64-js/-/base64-js-1.2.5.tgz",
-      "integrity": "sha1-WCskdhaabLpGCiFNR2x0REHYc9U="
-    },
     "@types/braces": {
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/@types/braces/-/braces-3.0.0.tgz",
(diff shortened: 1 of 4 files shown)


tamuratak:
I cannot reproduce the problem. Please upload the pdf file if you do not mind.

> But it takes around several seconds to over 10s for VSCode to render this data URL into an actual image for some reason, which also hangs the editor.

sqd:
@tamuratak Here you go

https://drive.google.com/drive/folders/1DNqKv3Ex_0i1H41dfHhzEFYgDY-QcqVw?usp=sharing

Test on an old unplugged laptop for maximum reproducibility.

tamuratak:
Thank you. I can reproduce the problem.

This PR does not work well with VS Code Remote. I will look into it.

sqd:
I don't think there is a clean way to deal with Remote under the current APIs. The document suggests using two different sets of configs for remote and local: https://code.visualstudio.com/docs/remote/troubleshooting#_local-absolute-path-settings-fail-when-applied-remotely

tamuratak:
- microsoft/vscode/issues/85464
- microsoft/vscode-remote-release/issues/1891

tamuratak:
@sqd, thank you for your contribution.

This works well with VS Code Remote now.

PR #1838, opened 2019-11-27
Split TeX options w.r.t. spaces

James-Yu:
Shall we do so? @jlelong

### DIFF
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -476,7 +476,7 @@ export class Builder {
             this.extension.logger.addLogMessage(`Found TeX program by magic comment: ${texCommand.command}`)
             const res = content.match(regexTexOptions)
             if (res) {
-                texCommand.args = [res[1]]
+                texCommand.args = res[1].split(' ')
                 this.extension.logger.addLogMessage(`Found TeX options by magic comment: ${texCommand.args}`)
             }
         }


jlelong:
I do not know what is the true motivation behind that but it sounds a good idea. In particular, it would enable us to simplify the way we currently spawn commands. When using the magic machinery, we concatenate the command and its argument as a single string and run it through a shell

https://github.com/James-Yu/LaTeX-Workshop/blob/4ec46c51234bc3a4dadd5aaa6b0b2748784e7fc8/src/components/builder.ts#L244-L254

However, I think we should be more careful than that with spaces because file names can contain spaces and then everything would be broken. Consider for instance 
````
% !TEX options = -synctex=1 -interaction=nonstopmode -file-line-error "Path to the main TeX file"
````
You do not want to split the quoted string because otherwise each part of the file name would be considered as an argument on its own. We should only split according to spaces that are not inside quotes.

James-Yu:
The referenced lines are quite interesting. It should work with either `cp.spawn` or `cp.exec`. The current implementation did not work for my use case though. Will close this PR for now and look into the issue.

jlelong:
Apart from options, the command `cp.exec` takes a single argument, namely the command name concatenated with all the arguments (you should take care of escaping spaces properly if needed), whereas the command `cp.spawn` takes two arguments the command name and an array of extra arguments (space escaping is automatically handled).

> The current implementation did not work for my use case though.

What is your issue?

James-Yu:
I was playing with magic comments to execute (xe)latexmk. When arguments are set via `tex options`, `latexmk` complaint about unindentifiable arguments.

Currently the magic command and options are concatenated and considered as `command` for `cp.spawn` with an empty `args`. I'm not sure this is intended by `spawn`. Will look into it.

jlelong:
Note the extra `{shell: true}` option when used with an empty `args`. My understanding is that `cp.spawn` with `{shell: true}` is equivalent to `exec` for what we want to do. Am I wrong?

PR #1840, opened 2019-11-27
Allow latexmk uses magic command to select compiler

SainoNamkho:
James-Yu#371 was once sovled by commit James-Yu@8ee9b52, but the solution was removed in James-Yu@67aa156#diff-e6eedacc0ab6a25fbf11532210610a31, which makes it difficult for users to work around without hard coding.

### DIFF
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -405,12 +405,21 @@ export class Builder {
                     const candidates = tools.filter(candidate => candidate.name === tool)
                     if (candidates.length < 1) {
                         this.extension.logger.showErrorMessage(`Skipping undefined tool "${tool}" in recipe "${recipe.name}."`)
+                        return
                     } else {
-                        steps.push(candidates[0])
+                        tool = candidates[0]
                     }
-                } else {
-                    steps.push(tool)
                 }
+                if (tool.command === 'latexmk' && magicTex) {
+                    if (!tool.args) {
+                        tool.args = []
+                    }
+                    tool.args.push(`-${magicTex.command}`)
+                    if(magicTex.args) {
+                        tool.args.push(`-${magicTex.command}="${magicTex.args.join(' ').replace('"','\\"')}"`)
+                    }
+                }
+                steps.push(tool)
             })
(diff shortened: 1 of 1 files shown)


jlelong:
I have to think of this because this PR changes the precedence of magic comments over the recipe mechanism. Currently, if `% !TEX program` is defined in the root file, then it is used to compile and recipes are not used. I guess you try to combine both. I would go for the following: if `% !TEX program` is not defined but `% !TEX options` is defined, the content of `% !TEX options` replaces the arguments defined in the recipe.

SainoNamkho:
It's very nice of you to take my problem into concern! 
The main problem I faced was sometime I write documents with CJK characters, which must be compiled by xelatex, but mostly I use pdflatex to save time. Then if I use `% !TEX options`, I write all  command-line arguments instead of merely '-xelatex'. And in addition, options not compatible between latexmk and LaTeX engines.
Also,  there's a tricky usage in  the pr to avoid  mechanism change: use `latexmk.exe` instead of `latexmk`, lol. 
Thank you again for the splendid tool!

jlelong:
Under Unix, there is only `latexmk`, no `latexmk.exe`. So your PR does change the precedence of magic comments over the recipe mechanism.

I am not convinced magic comments are the solution to your `pdflatex` vs `xelatex` issue. From the [wiki](https://github.com/James-Yu/LaTeX-Workshop/wiki/Compile#magic-comments): 

> However, it is advised to use the recipe system instead of magic program to define the building process, since the latter is only implemented for backward compatibility.

There are several solutions

1. Define a new recipe with the `-xelatex` flag passed to latexmk
1. Create `.latexmkrc` file in your `$HOME` directory containing `$pdf_mode=5;`. Then, use the recipe `latexmk (latexmkrc)`. You could even set this one as the default.

jlelong:
Is everything sorted out, so that i can close this PR?

jlelong:
I guess everything is fixed here. I am closing this issue but feel free to come back with anything new to add to the discussion.

PR #1842, opened 2019-11-27
Rename intellisense.preview.enabled to something more meaningful

trevorgunn:


jlelong:
I think the range of `intellisense.preview.enabled` goes beyond `\includegraphics`. 

https://github.com/James-Yu/LaTeX-Workshop/blob/f5493b3cec9a4894ad4f67e017b2a9aac8371d32/src/providers/hover.ts#L14-L30

It also governs whether to preview maths environments.

trevorgunn:
That is `hover.preview.enabled` for math environments.

jlelong:
My bad. Maybe, we could consider renaming `intellisense.preview.enabled` into something more meaningful.

PR #1855, opened 2019-12-04
The default color of backgourd of PDF viewer should be white

tamuratak:
With the current default, flickering, white -> black -> white, is much annoying when building.

Dec-05-2019 07-08-13

### DIFF
--- a/package.json
+++ b/package.json
@@ -1082,7 +1082,7 @@
         },
         "latex-workshop.view.pdf.backgroundColor": {
           "type": "string",
-          "default": "#525659",
+          "default": "#ffffff",
           "markdownDescription": "The background color around the document. The string must represent a color in HTML."
         },
         "latex-workshop.synctex.path": {


jlelong:
I cannot reproduce this flickering issue. What is your configuration?

tamuratak:
I have removed `settings.json`. I still can reproduce the problem with the default setting.
 
Dec-05-2019 18-10-21

Version: 1.40.2
Commit: f359dd69833dd8800b54d458f6d37ab7c78df520
Date: 2019-11-25T14:52:45.129Z
Electron: 6.1.5
Chrome: 76.0.3809.146
Node.js: 12.4.0
V8: 7.6.303.31-electron.0
OS: Darwin x64 18.7.0

TeX Live 2018

macOS Mojave 10.14.6 
iMac 3.4 GHz Intel Core i5
RAM 32 GB

jlelong:
That's strange because I do not have this intermediate blink.
To be clear, I really do not care if the default bg-color is white or not. I just would like to understand why we do not see the same behaviours.

jlelong:
Ok I get it now and I do agree this is a bit annoying. Let's set the default bg-color as white.

PR #1862, opened 2019-12-14
Syntax highlight for BibTeX style language (.bst)

zepinglee:
This is a naive implementation of language support for BibTeX style (.bst files).

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,15 @@
         ]
       },
       {
+        "id": "bibtex-style",
+        "aliases": [
+          "BibTeX style"
+        ],
+        "extensions": [
+          ".bst"
+        ]
+      },
+      {
         "id": "latex-expl3",
         "aliases": [
           "LaTeX-Expl3"
@@ -162,6 +171,11 @@
         "language": "bibtex",
         "scopeName": "text.bibtex",
         "path": "./syntax/Bibtex.tmLanguage.json"
+      },
+      {
+          "language": "bibtex-style",
+          "scopeName": "source.bst",
+          "path": "./syntax/BibTeX-style.tmLanguage.json"
       },
       {
         "language": "latex-expl3",
--- a/package.json
+++ b/package.json
@@ -173,9 +173,9 @@
         "path": "./syntax/Bibtex.tmLanguage.json"
       },
       {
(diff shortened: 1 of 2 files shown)


jlelong:
This is a good initiative! I tried the syntax highlighting on `IEEEtran.bst` and in some sections the highlighting is weird. I do not know much about `bst` syntax so I may be wrong.  Could you check that?

Capture d’écran 2019-12-14 à 20 47 21

Capture d’écran 2019-12-14 à 20 49 19

zepinglee:
Thanks for your feedback! The highlighted words in your two screenshots are native BibTeX entry types and fields that are defined in `plain.bst`, `unsrt.bst`. Thus the one additionally defined in 'IEEEtran.bst' are not correctly highlighted. I'll dig into the vscode APIs to get them into consistency.

jlelong:
Syntax highlighting might be a tricky. I have spent a lot of time working on the language definition files for `.tex` and `.dtx`, so I may be able to help if you tell more about the `bst` syntax.  In my screenshost,

- would you expect everything inside `ENTRY` to be highlighted or to remain white?
- same question for the `FUNCTION` lines?

zepinglee:
The `ENTRY` command declares its attributes of an entry: fields, integers and strings. 
```
ENTRY
  { field1 field2 ... }
  { int1 int2 ... }
  { string1 string2 ... }
```
It works like a Python class definition:
```Python
class Entry:
    Field field1, field2
    int int1, int2
    str str1, str2
```
The member attributes are not highlighted in the official python syntax of VS Code. Thus everything in the `ENTRY` command remains white now.

The `FUNCTION` command is a function declaration. It is now correctly highlighted as `storage.type.function.bst` and `entity.name.function.bst`.

jlelong:
Great! There is still an issue with `or` being highlighted as a logic keyword even in the middle or word.

Capture d’écran 2019-12-16 à 18 08 17

zepinglee:
I'm planning to write a set of unit tests like https://github.com/microsoft/TypeScript-TmLanguage/tree/master/tests to avoid such mistakes.

jlelong:
That would be awesome but it is a tough work!

Issue #1857, opened 2019-12-06
Changes to latexindent configuration requires to restart/reload VSCode


moi90:
### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

## Describe the bug [Required]

The arguments set in `latex-workshop.latexindent.args` are not passed to `latexindent`.

### To Reproduce

In `.vscode/settings.json` configure:

```
{
    "latex-workshop.latexindent.args": [
        "-c",
        "%DIR%/",
        "%TMPFILE%",
        "-y=defaultIndent: '%INDENT%'",
        "-l"
    ]
}
```

In the `.tex` file, right-click, "Format Document".

### Expected behavior

latexindent is called with the configured parameters (most of all, -l in my case).

### LaTeX Workshop Output [Required]

Instead, the command remains the default:
```
[15:24:48] Formatting with arguments -c,/home/moi/Work/19-mc-paper/,/home/moi/Work/19-mc-paper/__latexindent_temp.tex,-y=defaultIndent: '    '
[15:24:48] Formatted /home/moi/Work/19-mc-paper/test.tex
```

## Desktop [Required]
 - OS: Ubuntu 18.04
 - VS Code version: 1.40.2
 - Extension version: 8.4.2

### Are you using LaTeX Workshop with VS Code Remote?

No

trevorgunn:
I cannot reproduce this issue. Maybe this "In `.vscode/settings.yaml` configure:" is your problem. The settings should be in a `.json` file which is, by default `~/.config/Code/User/settings.json`. Check that when you open the settings editor in VSCode that the `latexworkshop.args` setting is what it should be.

jlelong:
The settings definitely needs a `.json` extension. See [vscode's documentation](https://code.visualstudio.com/docs/getstarted/settings#_settings-file-locations) for the different places of configuration files.

moi90:
Sorry, `.yaml` was a spelling error. It is actually `vscode/settings.json` and was generated by VS Code. Please re-open.

jlelong:
OK, I get it. Currently, the configuration related to `latexindent` is read only once when the extension is loaded. Any further changes will only be taken into account after restarting/reloading VSCode. So the first `.tex` file you open (or more precisely its workspace) fixes the configuration once and for all.

I will change this.

moi90:
[...]

PR #1863, opened 2019-12-14
Refactor latexindent detection

jlelong:


jlelong:
This PR fixes #1857 and a bit more.
Both `latexindent.path` and `latexindent.arguments` are checked everytime the formatter is called. So there is no need to reload/restart vscode to take into account any change in these variables.

Before merging, it would be nice to test it on Windows and using Docker.

Issue #1873, opened 2019-12-18
Error with 'Format Document' command when path has spaces


MrSinamun:
## Describe the bug

When trying to format a .tex document (Alt+Shift+F) it does not work due to there being a space (punctuation) in the paths. 
I've had a similar issue with the Maven for Java extension: [See the issue here](https://github.com/microsoft/vscode-maven/issues/304)
Might be related to #485, #786 

When building a latex project everything works fine, but there is an error related to an .aux file appearing in the developer console. I'm not sure this is related - But I've included this log under it's own section for just in case.

### To Reproduce

1. Open any .tex document
2. Use the _Format Document_ or _Build LaTex project_ command
3. See error

### Expected behavior

When using the _Format Document_ command, it should reformat the document.

### LaTeX Workshop Output 
```
[00:14:56] Formatting with command c:\Users\Mr Sinamun\.vscode\extensions\james-yu.latex-workshop-8.5.0\scripts\latexindent.bat -c,c:/Users/Mr Sinamun/Projects/web3/scratch/,c:/Users/Mr Sinamun/Projects/web3/scratch/__latexindent_temp.tex,-y=defaultIndent: '    '
[00:14:56] Formatting failed with exit code 1
[00:14:56] stderr: 'c:\Users\Mr' is not recognized as an internal or external command,
operable program or batch file.
```

### Developer Tools Console
`Formatting failed. Please refer to LaTeX Workshop Output for details.`

### LaTeX Workshop Output 
```
[00:47:25] File watcher: responding to change in c:\Users\Mr Sinamun\Projects\web3\scratch\lalala.tex
[00:47:25] Parsing c:\Users\Mr Sinamun\Projects\web3\scratch\lalala.tex
[00:47:25] c:\Users\Mr Sinamun\Projects\web3\scratch\lalala.tex changed. Auto build project.
[00:47:25] BUILD command invoked.
[00:47:25] Building root file: c:\Users\Mr Sinamun\Projects\web3\scratch\lalala.tex
[00:47:25] Build root file c:\Users\Mr Sinamun\Projects\web3\scratch\lalala.tex
[00:47:25] Recipe step 1: c:\Users\Mr Sinamun\.vscode\extensions\james-yu.latex-workshop-8.5.0\scripts\latexmk.bat, -synctex=1,-interaction=nonstopmode,-file-line-error,-pdf,-outdir=./,lalala
[00:47:25] LaTeX build process spawned. PID: 17488.
[...]

PR #1883, opened 2019-12-24
Use cross-spawn to run latexindent

jlelong:
Related to #1873. @MrSinamun, can you test the PR? Any help would be much appreciated.
See [here](https://github.com/James-Yu/LaTeX-Workshop/blob/master/CONTRIBUTING.md) for how to test the PR. Make sure to run `git checkout 1873-space-spawn` after `git clone`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1644,6 +1644,7 @@
   "dependencies": {
     "@tamuratak/domstubs": "^0.1.1",
     "chokidar": "^3.2.3",
+    "cross-spawn": "^7.0.1",
     "fs-extra": "^5.0.0",
     "glob": "^7.1.1",
     "iconv-lite": "^0.4.24",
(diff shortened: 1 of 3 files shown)


jlelong:
If this works, we should consider using `cross-spawn` for building and and synctex (when synctexjs is nto used).

package.json
```diff
@@ -1342,7 +1342,7 @@
           },
           "default": [
             "-c",
-            "%DIR%/",
+            "%DIR%",
```
MrSinamun: On line 1349 it says:
`\n\nNote: At this moment -c option requires trailing slash.`
If the removal of the trailing slash has no consequences, the note should be removed.

jlelong: Your are right. The trailing `/` is still compulsory. See cmhughes/latexindent.pl/issues/180

PR #1898, opened 2020-01-09
More precise error highlighting

peterzeller:
Extended the Latex log parser to retrieve columns for error messages.

Before this PR the plugin would always highlight the whole line:

image

With this PR the parser is able to highlight the correct error positions:

image

For an error like in the example above Latex reports an error like this:

```
/home/peter/kram/latextest/blub.tex:6: Undefined control sequence.
l.6 ... this is \textbf{no error} and this is \xxx
                                                  {another error}.
```

The parser now extracts the text `this is \textbf{no error} and this is \xxx` and tries to find it in the line of the respective text document.
If it finds the text, the last word in the matched text will be highlighted as error.

Tested this with pdflatex (pdfTeX 3.14159265-2.6-1.40.19 (TeX Live 2019/dev/Debian)), not sure if this would also work for other versions. However, it should always fall back to the old behavior when it does not find a line.


src/components/parser/log.ts
```diff
@@ -319,11 +327,35 @@ export class Parser {
...
+                    const line = document.lineAt(item.line - 1)
+                    if (line) {
+                        let pos = line.text.indexOf(item.errorPosText)
+                        if (pos >= 0) {
+                            pos += item.errorPosText.length
+                            // find the length of the last word in the error
+                            // This is the length of the error-range
+                            const len = item.errorPosText.replace(/^.* (.*)/, '$1').length
+                            if (len > 0) {
+                                startChar = pos - len
```
trevorgunn: Doesn't affect performance too much, but `startChar = item.errorPosText.lastIndexOf(' ')` would be better than this regex that has to do a lot of backtracking.

trevorgunn:
These changes look good to me. Based on reading the code and doing my own tests, it doesn't look like this is going to cause any problems.

src/components/parser/log.ts
```diff
@@ -158,7 +158,7 @@ export class Parser {
             }
             // append the read line, since we have a corresponding result in the making
             if (searchesEmptyLine) {
-                if (line.trim() === '' || (insideError && line.match(/^\s/))) {
```
trevorgunn: I don't think we want to remove `(insideError && line.match(/^\s/))` here.

The point of removing it would be to test the line against the `messageLine` regex but anything that matches `^\s` will never match `messageLine` (because `messageLine` starts with `^l`).

lrnv:
Hey

What would be requested for this PR to be finished & merged? I may give it a shot if you guide me a little.

Dominic-DallOsto:
Just wondering what the status of this PR is and what more would need to be done. This change would be really handy.

jlelong:
The logs parser is very complex so any change has to be made with extra care. @trevorgunn made some recommendations to improve the PR but apparently its author never really finished it and now it has conflicts with the `master` branch.

So, the PR basically needs to be reworked on top of the `master` branch while taking into account the recommended changes.

peterzeller:
Sorry for never finishing it.

I have now merged in current master and tried to adress the feedback.

src/components/parser/compilerlog.ts
```diff
@@ -114,11 +114,35 @@ export class CompilerLogParser {
...
         compilerDiagnostics.clear()
         const diagsCollection = Object.create(null) as { [key: string]: vscode.Diagnostic[] }
         for (const item of buildLog) {
-            const range = new vscode.Range(new vscode.Position(item.line - 1, 0), new vscode.Position(item.line - 1, 65535))
+            let startChar = 0
+            let endChar = 65535
+            if (item.errorPosText) {
+                // try to find the errorPosText in the respective line of the document
+                try {
+                    const document = await vscode.workspace.openTextDocument(item.file)
```
jlelong, line 126: This opens `item.file` in a new editor. I think some users will find it annoying. I suggest to read the content of this file by using https://github.com/James-Yu/LaTeX-Workshop/blob/e00be44387417c23c54bedc7897f85f780bc59ca/src/components/manager.ts#L515-L518

Note that for this to work you have
- to add `this.extension = extension` to the `constructor` and use `this.extension.manager.getDirtyContent`
- `item.file` must be a full path

jlelong (changes requested):
Could you also remove the changes to `package-lock.json` from your commits?

jlelong:
@peterzeller Are you still interesting in fixing this PR so that I can merge it?

lrnv:
@peterzeller I would be interested in the fact that you fix this PR :)

src/components/parser/compilerlog.ts
peq: I tested this and `openTextDocument` does not actually open the file in the UI. There is a separate method `vscode.window.showTextDocument` for showing the document in the UI.

I applied your suggestion anyway, but I guess the openTextDocument method has faster access to the string of a line, so maybe you want to merge without the last commit.

peterzeller: I tested this and openTextDocument does not actually open the file in the UI. There is a separate method vscode.window.showTextDocument for showing the document in the UI.

I applied your suggestion anyway, but I guess the openTextDocument method has faster access to the string of a line, so maybe you want to merge without the last commit.

jlelong:
> I tested this and `openTextDocument` does not actually open the file in the UI. There is a separate method `vscode.window.showTextDocument` for showing the document in the UI.

Yes, but once a document is opened with `openTextDocument`, the `didOpen`-event fires (see [here](https://code.visualstudio.com/api/references/vscode-api#openTextDocument)) and it has to be closed manually otherwise it remains open inside VS Code without the user knowing. Hence, I prefer to use `getDirtyContent` as we do everywhere else in the extension. 

I agree that we could try to improve the `getDirtyContent` approach to provide a `.lineAt` method without calling `split('\n')` every time. I leave for future improvement.

lrnv:
Would you mind tagging a new version with this included ?

PR #1905, opened 2020-01-14
Fix keybinding regression for ctrl+alt+[ and +]

trevorgunn:
In 6e1c8ea1a3ce83ac4bbf38532620b6263ff8fea2 both the `ctrl+alt+[` and `ctrl+l ctrl+[` were set to `config.latex-workshop.bind.altKeymap.enabled`.


tecosaur:
Nice spot!

Issue #1940, opened 2020-02-07
Remote docker containers


MagicLegend:
**Is your feature request related to a problem? Please describe.**
I use the Docker support for this plugin with a lot of love; it is one of the best niche-features of this extension. Not having to clutter my desktop with all the LaTeX things is amazing. However, recently I've ran into a minor issue.

**I want to run my texlive container remotely (over SSH).**

**Describe the solution you'd like**

I would prefer it if I could somehow use the `-H` option of `docker`. Using a command like `docker -H ssh://remoteUser@remote ps` for example; one can connect to the remote docker deamon and list the running containers. Looking at the scripts in this repo I think this shouldn't be too difficult to provide as an setting. However; I am really time-constraint right now (graduation), so I don't really have the time to look into it myself.

**Describe alternatives you've considered**

I have thought about other tricks with the remote-container plugin; only to realize that this of course doesn't modify the behaviour or this plugin (which is actually running the docker command). I also have tried modifiying my own docker installation to always point to the remote; but this is less than ideal as this makes me either switch between daemons (local and remote).

This idea would make it possible to compile within docker remotely, whilst keeping the files and editing remotely.

**Aditional context**
I need this because I work with git repo's; and the remote server I can use to compile with doesn't have access to said git repos (they're internal).

tamuratak:
Since VSCode Remote Development is available, we would not like to add additional features to the Docker support of LaTeX Workshop.

MagicLegend:
Sure, but Remote Development requires the LaTeX to be on the remote as well. I don't want that in this scenario. I want to keep the LaTeX on my local machine; and copy it to the remote for the container. This is a scenario that is not possible in the current LaTeX-Workshop.

PR #1941, opened 2020-02-07
Added support for remote Docker daemons

MagicLegend:
Based on my request in #1940 ; I have made a possible solution to my problem. It's crude; but it works. I still need to test it; but I'm not sure how to even compile this to an vscode extension ;-) Because of the time constraints that I mentioned in my original issue, I can't really teach myself how to write vscode plugins :smile:. I would expect this PoC to work; but it might be rougher than you want in your main branch. Any edits and comments are appreciated.

Right now the default setting is also the default in Docker; pointing to the local docker instance. It can be changed to `ssh://remoteUser@10.10.10.10` (provided key-based auth has been set up); and it will use the docker on that machine. This has the advantage that one can use this plugin; and compile on a (more powerful) remote machine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1381,6 +1381,11 @@
           "default": "tianon/latex",
           "markdownDescription": "Define the image for `latexmk`, `synctex`, `texcount`, and `latexindent`."
         },
+        "latex-workshop.docker.location": {
+          "type": "string",
+          "default": "unix:///var/run/docker.sock",
+          "markdownDescription": "Possibility to define an remote docker deamon. Will be passed to the `-H` argument of `docker`."
+        },
         "latex-workshop.showContextMenu": {
           "type": "boolean",
           "default": false,
--- a/scripts/latexindent
+++ b/scripts/latexindent
@@ -1,2 +1,2 @@
 #!/bin/sh
-docker -H $LATEXWORKSHOP_DOCKER_LOCATION run -i --rm -w "$(pwd)" -v "$(pwd):$(pwd)" $LATEXWORKSHOP_DOCKER_LATEX latexindent "$@"
+docker -H $LATEXWORKSHOP_DOCKER_DAEMON_LOCATION run -i --rm -w "$(pwd)" -v "$(pwd):$(pwd)" $LATEXWORKSHOP_DOCKER_LATEX latexindent "$@"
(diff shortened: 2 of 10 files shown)


MagicLegend:
So as I already expected, life is never this easy. There was a minor quirk with this solution that I forgot about: docker doesn't have access to the local filesystem anymore. I wrote a quick-and-dirty solution for this by locally directly editing the `latexmk` script as follows:

`#!/bin/sh
# Copy local latex root folder to remote
scp -r "$(pwd)" REMOTE_SERVER:/tmp/latextmp
# Excecute docker to compile
docker -H ssh://REMOTE_SERVER run -i --rm -w /tmp/latextmp -v "/tmp/latextmp:/tmp/latextmp" $LATEXWORKSHOP_DOCKER_LATEX latexmk "$@"
# Take compilation results and copy them back to the local root folder
scp -r REMOTE_SERVER:/tmp/latextmp "$(pwd)"
# Clean up the remote temp folder
ssh REMOTE_SERVER "rm -rf /tmp/latextmp"`

This works, though this also probably introduces some race-conditions. I really hoped to be able to keep it in a one-liner using Docker; but I can't come up with a different solution. As the PR is right now it should *not* be merged, this needs some more thinking.

MagicLegend:
I have updated the above script to use `rsync` instead of `scp`. It is now fully functioning, but it requires the knowledge of the `uid`. Otherwise the container will write back as `root`, creating issues with `rsync` trying to write those files back to the local machine.

```
#!/bin/sh
# Copy local latex root folder to remote
rsync -az "$(pwd)"/ REMOTE_SERVER:/tmp/latextmp
# Excecute docker to compile
docker -H ssh://REMOTE_SERVER run -i --rm -w /tmp/latextmp -v "/tmp/latextmp:/tmp/latextmp" --user 1000:1000 $LATEXWORKSHOP_DOCKER_LATEX latexmk "$@"
# Take compilation results and copy them back to the local root folder
rsync -az REMOTE_SERVER:/tmp/latextmp/ "$(pwd)"
# Clean up the remote temp folder
ssh REMOTE_SERVER "rm -rf /tmp/latextmp"
```
Also this now only works with two linux machines; I don't know how this would play out using Windows-based machines. Unless this is figured out as well, I'm afraid this solution won't be useable for issue #1940. Perhaps any of you have better, more cross-platform, ideas?

MagicLegend:
In order to prevent the race condition when you save whilst it's still compiling, I excluded my `content/` folder from the second `rsync` operation with `--exclude=content`. However, this would only work if the `.tex` files are all (apart from the master) in the `content/` folder. This solution works for me; but won't work in all cases.

tamuratak:
Since VSCode Remote Development is available, we would not like to add additional features to the Docker support of LaTeX Workshop.

Issue #1919, opened 2020-01-27
Updating the pdf outside from VS Code does not reload the pdf file in the viewer.


stakaz:
### Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: Yes

## Describe the bug [Required]

The PDF-Viewer inside the VS Code does not update the pdf file if it was updated outside of VS Code. 

### To Reproduce

1. Open a pdf file in the pdf viewer
2. Update the pdf file through a command outside VS Code
3. The previewed pdf does not change.

### Expected behavior

The pdf file is reloaded.

The pdf is **not** generated from a tex file but from a markdown file through pandoc. But the problem here is a general one since the Latex-Workshop is **the** way of opening pdf files in VS Code.

## Desktop [Required]
 - OS: arch linux
 - VS Code version: 1.41.1
 - Extension version: 8.6.0

### Are you using LaTeX Workshop with VS Code Remote?

No

jlelong:
You can call the _Refresh all viewers_ command manually `latex-workshop.refresh-viewer`.

Issue #1811, opened 2019-11-15
Add a time-to-settle before building based on filesystem changes.


WeaverThree:
**Is your feature request related to a problem? Please describe.**
On windows, running git commands in Cygwin can cause Latex Workshop to start rebuilding before the git process finishes, and then the git process fails to replace a file because it's locked by the compilation process. (Windows locks files for reading a lot harder than Linux...)

Also sometimes I'd like to save multiple interdependent tex files before a build starts.

**Describe the solution you'd like**
A file-monitoring time-to-settle or quiescence period option in the configuration. E.g., "if a file change is detected, wait until no file changes are detected for this many ms before starting a build."

**Describe alternatives you've considered**
"Build only when file saved by vscode" would be an acceptable alternative, allowing you to still build on ctrl-s but not when external stuff modifies files.

tamuratak:
Closed by #1949.

PR #1949, opened 2020-02-14
The delay is too long.

tamuratak:
@jlelong  The delay before starting builds, 2 sec,  is too long. What is the purpose of this commit, aa5e6ccfbf8d63fb1ff79a1c4597dd3ec99b2e55?

### DIFF
--- a/package.json
+++ b/package.json
@@ -846,6 +846,11 @@
           "default": 300,
           "markdownDescription": "Interval of polling, in milliseconds."
         },
+        "latex-workshop.latex.watch.delay": {
+          "type": "number",
+          "default": 250,
+          "markdownDescription": "Delay before starting bulids, in milliseconds."
+        },
         "latex-workshop.latex.watch.files.ignore": {
           "type": "array",
           "items": {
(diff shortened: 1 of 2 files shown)


jlelong:
The commit aa5e6cc was related to #1811 and #1919 (we watch PDF opened inside VSCode for automatic reload). The idea was to make sure the write process was indeed finished before triggering a `fileChange` event. I simply used the default value for `awaitWriteFinish`(see https://github.com/paulmillr/chokidar). It may not be needed for `.tex` files (except in #1811) but I think this is required for `.pdf` files which typically takes longer to be updated when they are generated by external tools. Maybe, we could change the default 2 seconds for `awaitWriteFinish.stabilityThreshold`. I did notice pay attention to the default 2 second delay, which is too long I agree.

tamuratak:
I am not sure the delay is really necessary. How about making it configurable?

LPTK:
What is the status on this? Is the delay configurable?

For smaller documents, it used to be that hitting `Cmd`+`S` refreshed the PDF preview almost immediately, which was great.

Now, whenever I need to make a tikz figure for instance, where I need quick feedback to see what my changes look like, I have to wait for this huge delay, which quite literally ruins my productivity doing this sort of tasks.

tamuratak:
I will be available in the next release.

PR #1955, opened 2020-02-18
add embedded language support for minted ruby

pantajosef:
Add syntax highlighting for ruby code in minted environments, as it currently existing for python, C, and JavaScript:

```latex
\begin{minted}{ruby}
module Test
  def foo
    :bar
  end
end
\end{minted}
```


pantajosef:
@jlelong Thanks for the info!

Issue #916, opened 2018-10-24
Embedded Languages: lualatex, pythontex, gnuplottex, dot2tex, ...


kiryph:
### Description

Currently, I use pythontex and the python library networkx together with tikz to create a more complex graph. 

I miss that the code with in `\begin{pycode}` and `\end{pycode}` is not marked as python code.

VSCode apparently supports embedded languages:
https://code.visualstudio.com/docs/extensions/themes-snippets-colorizers#_embedded-languages

This would also allow me to use ctrl-/ to comment python code correctly with `#` instead of `%`.

The file `syntax/LaTeX.plist` defines embedded languages for mintex/lstverbatim.

A few examples of embedded languages in a latex document are

```
\begin{luacode}
tex.print('This is lua.')
\end{luacode}

\begin{pycode}
import sys
import networkx as nx

G = nx.read_graphml('mygraph.graphml')

print(r'\begin{center}')
print(r'\textit{A message from Python '+sys.version+'!}')
print(r'\end{center}')

print(r'Number of nodes '+str(len(G.nodes)))
print(r'Number of edges: '+str(len(G.edges)))
\end{pycode}

\begin{gnuplot}[terminal=epslatex, terminaloptions=color dashed]
  set key box top left
  set key width 2
  set key opaque
  set sample 1000
  set xr [-5:5]
  set xlabel '$x$-label'
  set ylabel '$y$-label'
  plot  sin(x) w l lc 1 lw 3 t '$\sin(x)$',\
        cos(x) w l lc 2 lw 3 t '$\cos(x)$',\
        tan(x) w l lc 3 lw 3 t '$\tan(x)$',\
        tanh(x) w l lc 4 lw 3 t '$\tanh(x)$'
\end{gnuplot}

\begin{dot2tex}[neato,mathmode]
  digraph G {
    node [shape="circle"];
    a_1 -> a_2 -> a_3 -> a_4 -> a_1;
    }
\end{dot2tex}
```

PR #1966, opened 2020-02-27
Add \addplot grammar support

### LINKED ISSUE #1963: Gnuplot support for pgfplots

Would it be possible to fix syntax colouring when using gnuplot with pgfplots ? Since we can use the dollar sign to call columns, it breaks the syntax colouring of the whole document (see screenshot below).

It seems that gnuplot syntax colouring is already supported for `gnuplot` environnement (see #916) 

At least disable syntax colouring in the example case so that it does not break the whole document.

**Additional context**
image

jlelong:
Fix #1963

### DIFF
--- a/syntax/LaTeX.tmLanguage.json
+++ b/syntax/LaTeX.tmLanguage.json
@@ -500,6 +500,44 @@
       "end": "(\\\\end\\{gnuplot\\}(?:\\s*\\n)?)"
     },
     {
+      "begin": "((\\\\)addplot)(?:\\+?)((?:\\[[^\\[]*\\]))*\\s*(gnuplot)\\s*((?:\\[[^\\[]*\\]))*\\s*(\\{)",
+      "captures": {
+        "1": {
+          "name": "support.function.be.latex"
+        },
+        "2": {
+          "name": "punctuation.definition.function.latex"
+        },
+        "3": {
+          "patterns": [
+            {
+              "include": "#optional-arg"
+            }
+          ]
+        },
+        "4": {
+          "name": "variable.parameter.function.latex"
+        },
+        "5": {
+          "patterns": [
+            {
+              "include": "#optional-arg"
+            }
+          ]
+        },
+        "6": {
+          "name": "punctuation.definition.arguments.begin.latex"
+        }
+      },
+      "contentName": "source.gnuplot",
+      "patterns": [
+        {
+          "include": "source.gnuplot"
+        }
+      ],
+      "end": "\\s*(\\};)"
+    },
+    {
       "begin": "(?:\\s*)((\\\\)begin)(\\{)((?:fboxv|boxedv|V|v)erbatim\\*?)(\\})",
       "captures": {
         "1": {
@@ -855,7 +893,8 @@
       },
       "end": "((\\\\)end)(\\{)(\\4)(\\})(?:\\s*\\n)?",
       "name": "meta.function.environment.latex.mpost"
-    },{
+    },
+    {
       "begin": "(?:\\s*)((\\\\)begin)(\\{)(\\w+[*]?)(\\})",
       "captures": {
         "1": {
--- a/syntax/LaTeX.tmLanguage.json
+++ b/syntax/LaTeX.tmLanguage.json
@@ -529,8 +529,17 @@
           "name": "punctuation.definition.arguments.begin.latex"
         }
       },
-      "contentName": "source.gnuplot",
       "patterns": [
+        {
+          "begin": "%",
+          "beginCaptures": {
+            "0": {
+              "name": "punctuation.definition.comment.tex"
+            }
+          },
+          "end": "$\\n?",
+          "name": "comment.line.percentage.tex"
+        },
         {
           "include": "source.gnuplot"
         }


jlelong:
@PierreMarchand20 Could you test this branch and let me know if the regex is robust enough
https://github.com/James-Yu/LaTeX-Workshop/blob/dc3da70ca43d3320a29de1909d23c17f575e0ec3/syntax/LaTeX.tmLanguage.json#L503

PierreMarchand20:
Sure, could you just remind me how to use this extension locally ? So far, I have cloned the repository, I did `npm install` in the folder of the repository and then I launched the debugger. But then the local version is not picked up, did I miss something ?

jlelong:
You just need to run

```bash
git clone https://github.com/James-Yu/LaTeX-Workshop.git
cd ./LaTeX-Workshop
git checkout 1963-pgfplots-support
npm install --no-optional
code .
```

Then, launch the debugger. It will run the local version

PierreMarchand20:
Ok then, the example I gave does not work: 
image
How can I check that I am using the local version ?

jlelong:
Strange because obviously your example works. In `LaTeX-Workshop`, make sure you are on the correct branch `1963-pgfplots-support`, then running `code -n .` should open a new VSCode window. Launching the debugger will fire up a `Extension Development Host` window, in which you have to open the `.tex` you want to check.

PierreMarchand20:
I am on the right branch 
image
I have the repo in a new window on the right branch 
image
I run it 
image

But I get the picture I sent you previously. But the very first time I tried to debug, I got an error message about something missing and that it was not able to launch the local version of the extension. But I cannot reproduce it now...

PierreMarchand20:
I deleted the folder and reinstall everything as you said, but same remarks as my previous message.

PierreMarchand20:
Here's the development tool:
image

jlelong:
> I got an error message about something missing and that it was not able to launch the local version of the extension

That may be the issue. Can you run `npm run compile` directly in the extension repository?

PierreMarchand20:
Ok, I did and I got 
```
❯ npm run compile

> latex-workshop@8.7.2 compile /Users/pfcm20/Partage/vscode/LaTeX-Workshop
> tsc -p tsconfig.json && tsc -p viewer/tsconfig.json
```
but then it did not change anything. But according to the development tool it seems that the local version is indeed loaded no ?

PierreMarchand20:
So I checked that I am using the local version (I modified a scope for the `pycode` environment and checked the scope was changed while debugging), and my example does not work.

PierreMarchand20:
I noticed that if I do not have any extension providing `source.gnuplot`, then it breaks. But if I have such an extension, it works fine, here's a more complicated example:
image

But there is something I do not understand following #916 and that you can see in my example. We use the syntax colouring of gnuplot, but the way we need to comment is like latex, so they are not grey as one would expect. So I do not understand what kiryph says in #916 because for me, commenting the python/gnuplot way breaks compilation...

jlelong:
For sure, you need an other extension to provide gnuplot support. I have tries with https://marketplace.visualstudio.com/items?itemName=mammothb.gnuplot I confirm that commenting uses `#` inside `\addplot`. 

> So I do not understand what kiryph says in #916 because for me, commenting the python/gnuplot way breaks compilation...

That is a problem because then the content of `\addplot` is not true gnuplot style. I will try to improve it later today.

PierreMarchand20:
1) What surprised me is that if you do not have a gnuplot extension, the colouring breaks. I would expect that the latex colouring works and that there is no colouring at all for the gnuplot script without gnuplot extension, this is not the case. But why not.

2) The issue about comments I was mentioning is that it takes the gnuplot comment style (using `#`) but it does not compile. It is an issue of `pgfplots`, not of this extension. For example, in a `\begin{gnuplot}...\end{gnuplot}`, it comments automatically with `#` and it compiles because the related package `gnuplottex` provides this functionality. I submitted a feature request to `pgfplots` about it: https://sourceforge.net/p/pgfplots/support-requests/23/

PierreMarchand20:
Thank you for the PR and the quick answer by the way!

jlelong:
1. If you do not have a gnuplot extension, the gnuplot language is unknown and the embedding does not work (`source.gnuplot` is unknown)
1. If we consider the `%` comment approach is a bug from `pgfplots`, there are two options
    1. We use the full embedding, meaning that commenting will not work as it will use `#`. We hope it will be fixed soon by the `pgfplots` guys
    1. We only use gnuplot syntax highlighting and on top of it I declare anything after % as a comment. At least I try. Everything will be fully working, but we will have to come to it later when `pgfplots` is fixed.

   May feeling is that ii. would be better.

PierreMarchand20:
1) strange default behaviour of vscode but ok. 

2) Sure ii) would be better in the meantime, if you can do it. It was just to say that if you do not it would be understandable ^^

jlelong:
I have just implemented ii. Can you give it a try and let me know?

PierreMarchand20:
It works fine, thank you !

jlelong:
So I am merging it. Let me know if/when commenting with `#` is supported.

PierreMarchand20:
FYI: pgf-tikz/pgfplots#346

PR #1968, opened 2020-02-28
latex-workshop.latex.recipe 'pdflatex × 3' added

wollmich:
The 'pdflatex × 3' recipe can be used when no bib-file is present. The existing 'pdflatex ➞ bibtex ➞ pdflatex × 2' receipe is not working without a bib-file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -658,6 +658,14 @@
               ]
             },
             {
+              "name": "pdflatex × 3",
+              "tools": [
+                "pdflatex",
+                "pdflatex",
+                "pdflatex"
+              ]
+            },
+            {
               "name": "Compile Rnw files",
               "tools": [
                 "rnw2pdf"


jlelong:
Just use the `latexmk` recipe, which figures out everything on its own. I cannot really see the point in adding `pdflatex x 3`.

wollmich:
I'm using MikTeX on Windows 10. I've not installed `latexmk` neither `PERL`.
The existing `pdflatex ➞ bibtex ➞ pdflatex × 2` recipe works perfect with MikTeX as long as I've a bib-file.
So I added an additional `pdflatex x3` recipe which skips the bibtex step.

jlelong:
OK, I see your point, but then calling `pdflatex` twice should be enough, shouldn't it?

wollmich:
No calling `pdflatex` is not enough, e.g.: wrong width of column header of a `longtable`. After 3 times running `pdflatex` it looks ok.

JimmyZJX:
MikTex has the `texify` utility.

You may search for "texify" here: https://github.com/James-Yu/LaTeX-Workshop/wiki/Compile

wollmich:
`texify` is well not in the default recipes of LaTeX Workshop. I tried it and it seems to work. But I found. out that running `texify` only once could not be enough, see https://tex.stackexchange.com/questions/78036/texify-do-you-ever-need-to-run-it-more-than-once.
So I would still propose to add the `pdflatex × 3` recipe to the default recipes.

James-Yu:
This is more a personal choice of recipe. As no discussions were made for a long time, I am closing this PR for now.

PR #1969, opened 2020-02-28
Prevent PDF viewer from stealing focus on refresh

rliang:
Does not happen in normal VSCode, but does in [code-server](https://github.com/cdr/code-server).
Shouldn't affect usage, since the `isViewerEmbedded` setting used to be enabled anyway.


tamuratak:
code-server is not supported. See https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ#code-server-is-not-supported.

PR #1989, opened 2020-03-17
Add syntax highlighting for TypeScript in minted

leafac:
This is the same that already happens for JavaScript:

Screen Shot 2020-03-17 at 4 00 59 PM

### DIFF
--- a/syntax/LaTeX.tmLanguage.json
+++ b/syntax/LaTeX.tmLanguage.json
@@ -220,6 +220,25 @@
           "end": "(\\\\end\\{minted\\})"
         },
         {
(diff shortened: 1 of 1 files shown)


jlelong:
FYI, if you want the content of the `minted` environment to be considered as `typescript` and have `typescript` full support (snippets, comments, ....) you also need to add 

https://github.com/James-Yu/LaTeX-Workshop/blob/34a8d9111ba98e2bec5483362c8c3c7463867ef1/package.json#L168-L168

See https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#embedded-languages for detailed explanations

leafac:
Thanks for fixing it and for teaching me 😃

PR #2013, opened 2020-03-29
Enhance build for rnw files

jlelong:
Related to #2005 

This PR implements
- [x] `alt+ctrl+v` to view the `pdf` file
- [x] Accept projects with several `.rnw` files
- [x] `alt+ctrl+b` to trigger the build
- [x] Choose an appropriate default recipe for `.rnw` files.
- [x] The following extensions are associated to the `Rsweave` language Id : `.[rRsS]nw`, `.[sR]tex`
- [x] Report LaTeX errors/warnings in the _Problems_ pane

### DIFF
--- a/package.json
+++ b/package.json
@@ -395,13 +395,13 @@
         "key": "ctrl+l alt+c",
         "mac": "cmd+l alt+c",
         "command": "latex-workshop.clean",
-        "when": "editorLangId == 'latex' && config.latex-workshop.bind.altKeymap.enabled"
+        "when": "editorLangId =~ /latex|rsweave/ && config.latex-workshop.bind.altKeymap.enabled"
       },
       {
         "key": "ctrl+l alt+v",
         "mac": "cmd+l alt+v",
         "command": "latex-workshop.view",
-        "when": "editorLangId == 'latex' && config.latex-workshop.bind.altKeymap.enabled"
+        "when": "editorLangId =~ /latex|rsweave/ && config.latex-workshop.bind.altKeymap.enabled"
       },
       {
         "key": "ctrl+l alt+j",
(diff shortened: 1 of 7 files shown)


jlelong:
This PR provides a quite satisfying support for `.rnw` files. Yet, there are two difficulties related to the same technical problem

- The logs showed in the _Problems_ pane refer to the `.tex` file not the `.rnw` one
- Synctex links the `.tex` and `.pdf` files. We would rather use the `.rnw` file instead of the `.tex` one.

Fixing these points requires an important effort because we have to introduce the mapping between the `.tex` and `.rnw` files defined by the generated `-concordance.tex` file generated by `knit` when setting `knitr::opts_knit$set(concordance = TRUE);`. Once the mapping is available, it must be introduced both in the synctex and log parsing mechanisms. This will undoubtedly introduce a lot of code change. Is it worth it?

@RaghuNaik @maikol-solis @ChristopheRoyer21 @armarti You all showed some interest in editing `.rnw` using LaTeX-Workshop. How useful would the concordance mechanism be?

jlelong:
This PR fiddles with the placeholders and the build mechanism. It requires some heavy testing before merging. 

@tamuratak Could you have a look at it when you have time? Sometimes I am wondering if we (I ?) do not go too far beyond the scope of the extension to offer full `.rnw` support.

tamuratak:
VS Code is supporting Jupyter Notebook. See https://github.com/microsoft/vscode/issues/93265.  I guess that users who want to embed `R` code and graphics to their documents will start using that directly. I am not sure supporting `rnw` is promising.

I never use `R`. So this is just a guess.

I have a question. I am not sure why we cannot detect `languageId` from the suffix of `rootFile`.

https://github.com/James-Yu/LaTeX-Workshop/blob/d454bcfade67847459cec4c8c76d3571fb1331e1/src/components/manager.ts#L178-L194

jlelong:
I do not use `R` that much myself but a lot of colleagues of mine do and they `.rnw` quite a lot to generate several PDF from a single `.tex` file with different data sets for instance. It seems to be quite popular. I think it is a good point for the extension to support `.rnw` file. The real question is how elaborate the support should be (concordance stuff). For the moment, I do not plan to go beyond this PR.

>  I am not sure why we cannot detect `languageId` from the suffix of `rootFile`.

This is a really good question and that was actually my first thought. Then, I thought that it would be more robust to read the `document.languageId` property when available. Looking back on this, I think you are right as the `document.languageId` property is only available in `findRootFromActive`. I will change this.

tamuratak:
I have no reason to be against this PR if you can see real use cases.

ChristopheRoyer21:
Same here, it looks fine to me.

PR #2015, opened 2020-03-31
Use cross-spawn to build and view

### LINKED ISSUE #2010: The placeholder %OUTDIR% on Windows
@jlelong I ask you a question on the placeholder, `%OUTDIR%`, on Windows. In the current `master` branch, backward slashes are used in `%OUTDIR%` on Windows since [path.normalize](https://nodejs.org/api/path.html#path_path_normalize_path) replace forward slashes with backward slashes on Windows. Is this an intended behavior?

https://github.com/James-Yu/LaTeX-Workshop/blob/bce11da246b8ae8b12b2fa0ea43b230fd3113e91/src/components/manager.ts#L82

We can see the behavior of `path.normalize` on Windows by calling `path.win32.normalize` on other platforms.

```node
path.win32.normalize('C:////temp\\\\/\\/\\/foo/bar');
// Returns: 'C:\\temp\\foo\\bar'
```

Related to #1897, #1904, and #1947.

jlelong:
Related to the tests and discussions in #2010.
It must be **extensively tested** before merging (paths with `/`, `\\`, spaces and also custom recipes using system commands).

Close #2010

Edit: **Summary of the PR**

- All calls to `child_process.spawn` are replaced by `cross-spawn.spawn`, which takes care of escaping issues in a cross platform way. 
- We introduce new placeholders `%DOC_W32%`, `%DOC_EXT_W32%`, `%DIR_W32%`, `%OUTDIR_W32%`, which are normalized so that they use `\\` as the path separator on Windows. Placeholders without the `_W32` suffix always use `/` as the path separator. On Unix platforms, placeholders with and without the  `_W32` suffix have the same value.


jlelong:
@tamuratak while I was adding more tests, I realised that the issue on `/`vs `\` on Windows is deeper than I thought. See https://stackoverflow.com/a/10526678 for a series of (non-) working examples.

I finally figured out that the tool

```
      {
          "name": "copyPDF",
          "command": "copy",
          "args": ["%OUTDIR%/t.pdf", "%OUTDIR%/b.pdf"],
          "env": {}
        }
```

works fine as long as `%OUTDIR% = %DIR%` . Remember that the `cwd` of the spawned command is `%DIR%`. The issue arises when the `dirname` of the file to be copied uses `/` and does not match the `cwd`.

I think that instead of introducing `_W32` placeholders, we should add a `normalizing` flag for each tool. Clearly, LaTeX commands must use `/` as separators regardless of the os, but for native commands normalising the args is simpler and even required in some cases. Now, the question is how to achieve this with breaking compatibility.

jlelong:
See https://github.com/James-Yu/LaTeX-Workshop/wiki/Compile#placeholders for the definitions of the new placeholders.

PR #2025, opened 2020-04-03
Render citation completion and preview as markdown

### LINKED ISSUE #2022: Follow URL from citation browser/preview
When in citation preview, make URL "clickable".
image

Would make it easier to use LaTeX document to check the actual content of citations without having to copy-paste links or dig through the compiled PDF. I think it used to work this way.

jlelong:
Close #2022

### DIFF
--- a/src/providers/completer/citation.ts
+++ b/src/providers/completer/citation.ts
@@ -43,6 +43,7 @@ export class Citation {
             item.filterText = `${item.key} ${item.fields.author} ${item.fields.title} ${item.fields.journal}`
             item.insertText = item.key
             item.documentation = new vscode.MarkdownString(item.detail.replace(/\n/g, '  \n'))
+            item.detail = ''
             if (args) {
                 item.range = args.document.getWordRangeAtPosition(args.position, /[-a-zA-Z0-9_:.]+/)
             }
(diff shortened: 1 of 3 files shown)


src/providers/completer/citation.ts
```diff
@@ -43,6 +43,7 @@ export class Citation {
             item.filterText = `${item.key} ${item.fields.author} ${item.fields.title} ${item.fields.journal}`
             item.insertText = item.key
             item.documentation = new vscode.MarkdownString(item.detail.replace(/\n/g, '  \n'))
+            item.detail = ''
```
jlelong, line 46: @tamuratak What do you think of dropping `item.detail`?
It was rendered with a monospace font, whereas `item.documentation` is now markdown and is rendered using a much smaller font.

tamuratak: I do not have a specific opinion. 

`item.detail` will be always visible. See microsoft/vscode/issues/39441. It might be used for another purpose.

jlelong: Thanks for pointing microsoft/vscode/issues/39441. For the moment, `item.detail` and `item.documentation` are always duplicates of each other, so I will keep `item.detail = ''` and I will come back to this once the `API` for `completionItem` is changed.

PR #2029, opened 2020-04-07
Refactor environments snippets 

### LINKED ISSUE #2006: Some environment snippets lack mandatory arguments
This is related to the discussion in #2001 

For instance for the `column` environment in Beamer, we simply provide
```
\begin{column}
	
\end{column}
```

whereas the environment must have a mandatory width argument

```
\begin{column}{<width>}
	
\end{column}
```

We need to refactor the snippet mechanism for environment.

jlelong:
Close #2006
This PR contributes the following features

- [x] Create snippets for environments with extra arguments like

    ```
   \begin{env}<some extra args which can even split on several lines>
      Some starting text
   \end{env}
    ```

- [x] Insert any environments (the default ones + those contributed by the used packages) by using the `\envname` snippet, which gets expanded into

    ```
   \begin{env}<some extra args which can even split on several lines>
      Some starting text
   \end{env}
    ```
   and takes care of the extra arguments. Only when `latex-workshopintellisense.package.env.enabled` is true (its default value).

- [x] Handling extra arguments with the `\begin{...} / \end{...}` snippet -- ie when there is a multi-cursor selection -- is not possible. So we simply provide the list of environments names in this case without taking care of the extra arguments.

### DIFF
--- a/src/providers/completer/environment.ts
+++ b/src/providers/completer/environment.ts
@@ -94,35 +94,42 @@ export class Environment {
         // Extract cached envs and add to default ones
         const suggestions: vscode.CompletionItem[] = Array.from(this.getDefaultEnvs(snippetType))
         const envList: string[] = this.getDefaultEnvs(snippetType).map(env => env.label)
+
+        // Insert package environments
+        if (vscode.workspace.getConfiguration('latex-workshop').get('intellisense.package.enabled')) {
+            this.extension.manager.getIncludedTeX().forEach(tex => {
+                const pkgs = this.extension.manager.cachedContent[tex].element.package
+                if (pkgs !== undefined) {
+                    pkgs.forEach(pkg => {
+                        this.getEnvFromPkg(pkg, snippetType).forEach(env => {
+                            if (!envList.includes(env.label)) {
+                                suggestions.push(env)
+                                envList.push(env.label)
+                            }
+                        })
+                    })
+                }
+            })
+        }
+
+        // Insert environments defined in tex
         this.extension.manager.getIncludedTeX().forEach(cachedFile => {
             const cachedEnvs = this.extension.manager.cachedContent[cachedFile].element.environment
             if (cachedEnvs !== undefined) {
                 cachedEnvs.forEach(env => {
                     if (! envList.includes(env.label)) {
+                        if (snippetType === EnvSnippetType.ForBegin) {
+                            env.insertText = new vscode.SnippetString(`${env.label}}\n\t$0\n\\end{${env.label}}`)
+                        } else {
+                            env.insertText = env.label
+                        }
                         suggestions.push(env)
                         envList.push(env.label)
                     }
                 })
             }
         })
-        // If no insert package-defined environments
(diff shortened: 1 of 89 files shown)


jlelong:
I feel like I am almost done here. Everything I had in mind is now implemented. Some `_env.json` files may still need some manual tweaking.

@tamuratak Could you have a look at it when you have time?

src/providers/completer/environment.ts
```diff
@@ -114,10 +240,60 @@ export class Environment {
...
             envList.push(result[1])
         }
         return envs
     }
+
+    entryEnvToCompletion(item: EnvItemEntry, type: EnvSnippetType): Suggestion {
+        const label = item.detail ? item.detail : item.name
+        const suggestion: Suggestion = {
+            label: item.name,
+            kind: vscode.CompletionItemKind.Function,
```
tamuratak, line 261: Is it intentional to use `vscode.CompletionItemKind.Function` for environments here? Previously and other cases in the same file, `vscode.CompletionItemKind.Module` is used.

src/providers/completer/environment.ts
```diff
@@ -4,59 +4,171 @@ import {latexParser} from 'latex-utensils'
...
+    initialize(envs: {[key: string]: EnvItemEntry}) {
+        this.defaultEnvsAsCommand = []
+        this.defaultEnvsForBegin = []
+        this.defaultEnvsAsName = []
+        const endCompletion: Suggestion = {
+            label: 'Complete with \\end',
+            sortText: ' ',
+            insertText: new vscode.SnippetString('$1}\n\t$0\n\\end{$1}'),
+            command: { title: 'Post-Action', command: 'editor.action.triggerSuggest' },
+            kind: vscode.CompletionItemKind.Function,
```
tamuratak, line 42: Is it intentional to use `vscode.CompletionItemKind.Function` for environments here? Previously and other cases in the same file, `vscode.CompletionItemKind.Module` is used.

src/providers/completer/environment.ts
jlelong: This is a good question. I guess I took that part from 

https://github.com/James-Yu/LaTeX-Workshop/blob/cf0ec7c35f2aa1148f353e361ac5f7f58e09292a/src/providers/completer/command.ts#L455-L460

The value of `vscode.CompletionItemKind` seems a bit. The previous values were

- For true LaTeX commands: `vscode.CompletionItemKind.Function`
- For `\envname` snippet: `vscode.CompletionItemKind.Snippet`

https://github.com/James-Yu/LaTeX-Workshop/blob/a32dfdf/src/providers/completer/command.ts#L51-L58

- For environment completion: `vscode.CompletionItemKind.Module`

To stick to this, I propose to fix the value ``vscode.CompletionItemKind` based on the `EnvSnippetType`

- `AsName`: `vscode.CompletionItemKind.Module`. It just returns a list of environment names
- `AsCommand`: `vscode.CompletionItemKind.Snippet`. It returns commands of the form `\envname`.
- `ForBegin`: `vscode.CompletionItemKind.Module`. It is designed to complete `\begin{` with `envname}<some extra stuff>\n\end{envname}`.

PR #2071, opened 2020-05-06
Enhance LaTeX3 support

### LINKED ISSUE #2051: Intellisense for expl3
`expl3` commands are much longer than that of normal LaTeX. They also contains the underscore, which is a "word-separator" in VSCode, so the automatic "word-auto-suggestion" doesn't help. So, when coding in `expl3`, one often has to type the whole command manually. It is really time-consuming.

The auto-suggestion feature added to `expl3`.

jlelong:
Currently, the extension only has syntax highlighting support for `LaTeX3`. 
This PR implements the following features

- [x] Activate the extension on `latex-expl3` language-id
- [x] Make all mechanisms for `latex` language-id available for `latex-expl3`
- [x] Intellisense for `LaTeX3` commands using
    - [x] the commands provided by the [`l3kernel`](https://github.com/latex3/latex3/tree/master/l3kernel)
    - [x] the user defined commands

Close #2051


jlelong:
_Originally posted by @joulev in https://github.com/James-Yu/LaTeX-Workshop/issues/2051#issuecomment-625156719_

---------

@jlelong It works wonderfully! For now I haven't found any issues. It even works with `l3docstrip`'s `@@` syntax, which I didn't even expect.

However, I guess the list should still be modified even more:
* Five variables with `foo` in it, namely `\c_foo_regex`, `\l_foo_clist`, `\l_foo_int`, `\l_foo_seq` and `\l_foo_str`, are not built-in variables so they should not be in the list (they are only used in the example, that is why they appears in the index).
* Functions that end with `TF`, like `\cs_if_eq:NNTF`, is understood to have three different formats: `\cs_if_eq:NNT`, `\cs_if_eq:NNF` and `\cs_if_eq:NNTF` (page 4-5 of interface3.pdf). Currently only the last of the three is included.

jlelong:
_Originally posted by @joulev in https://github.com/James-Yu/LaTeX-Workshop/issues/2051#issuecomment-625166678_

-------

There is a weak point though, but I don't think this can be resolved in an easy fix: there are *function variants* in LaTeX3. Simply put, they are functions having the same name but different signatures. An [extreme] example:

image

(page 39 of interface3.pdf). `\tl_set:Nn` has 11 variants, so the user can use any of the 12. However only the "original" variant, `:Nn`, is indexed, so a large number of variants is missing in the list.

However, tracing these variants is not trivial. As they are not indexed, there are no lists available to parse.

Here I suggest two solutions:

* Leave it as is. If the user uses a variant, that variant will be automatically included, so the user only has to type manually in the first instance of that variant.
* Parse the sources of interface3.pdf, i.e. all the `*.dtx` files that you finds in `source/latex/l3kernel` of your texmf tree: You look at the part inside `documentation` environment (parsing outside it will result in internal functions being included). Find the `function` environments. They should be something like this:

    ```
    % \begin{function}{\clist_new:N, \clist_new:c}
    %   \begin{syntax}
    %     \cs{clist_new:N} \meta{comma list}
    %   \end{syntax}
    %   Creates a new \meta{comma list} or raises an error if the name is
    %   already taken. The declaration is global. The \meta{comma list}
    %   initially contains no items.
    % \end{function}
    ```
    
    Now take all functions listed in the argument of `function` to the list. In this case, `\clist_new:N` and `\clist_new:c`.

The second is rather a complicated process, so if you uses the first one, it is okay. What do you think?

jlelong:
Thanks for pointing out the missing variants issue. I will closely at the `.dtx` files and write a small Python parser to extract the relevant information. Hopefully, I can get back to you on this within a day or two.

jlelong:
As you suggested, I parsed the `*.dtx` files and got the following [list](https://gist.github.com/jlelong/9a49e2bc806f076905f6030faa052cdc#file-dtx-functions-json). Of course, I may need to apply `uniq` on it but still I have the feeling that it is unreasonably long and more importantly that some entries should not be there.
Could you have a look at it and let me know?

joulev:
Some thoughts:
* I think the json file lists only all function *names*, right (I mean the signatures are not yet taken into account)? Because for example, I use `c` argument specifier quite a lot (`\cs_new:cpn`, etc.), but I can't find any functions having this specifier.
* `l3doc.dtx` is for the `l3doc` class and `l3docstrip`, it is IMHO not necessary.
* As you have successfully parsed the dtx files for functions, perhaps parsing for variables will be easy? (Sorry I forgot this in https://github.com/James-Yu/LaTeX-Workshop/issues/2051#issuecomment-625166678, and IMHO variables are not less important than functions). It is almost identical: instead of parsing `function` environments, you only need to parse `variable` environments.

jlelong:
Here is the new [list](https://gist.github.com/jlelong/9a49e2bc806f076905f6030faa052cdc#file-dtx-functions-variables-json) with

- variables from the `.dtx` files
- the file `l3doc.dtx` being ignored
- all the function signatures (there was a parsing error because of leading spaces i had not stripped)

Let me know if it is better

joulev:
I am not sure what went wrong, but while many frequently-used variants are already included, the list doesn't pass most of my random tests. These are some of them.

* [`\iow_shipout_x:cx`](https://github.com/latex3/latex3/blob/master/l3kernel/l3file.dtx#L428-L432)

    ```
    % \begin{function}[updated = 2012-09-08]
    %   {
    %     \iow_shipout_x:Nn, \iow_shipout_x:Nx,
    %     \iow_shipout_x:cn, \iow_shipout_x:cx
    %   }
    ```

* [`\tl_set:<anything>`](https://github.com/latex3/latex3/blob/master/l3kernel/l3tl.dtx#L162-L165)

  ```
  % \begin{function}
  %   {
  %     \tl_set:Nn, \tl_set:NV, \tl_set:Nv, \tl_set:No, \tl_set:Nf, \tl_set:Nx,
  %     \tl_set:cn, \tl_set:cV, \tl_set:cv, \tl_set:co, \tl_set:cf, \tl_set:cx,
  ```

  `\tl_set`, for some reasons, doesn't even show up in the json file.

* [`\box_autosize_to_wd_and_ht_plus_dp:<anything>`](https://github.com/latex3/latex3/blob/master/l3kernel/l3box.dtx#L687-L693)

  ```
  % \begin{function}[added = 2017-04-04, updated = 2019-01-22]
  %   {
  %     \box_autosize_to_wd_and_ht_plus_dp:Nnn,
  %     \box_autosize_to_wd_and_ht_plus_dp:cnn,
  %     \box_gautosize_to_wd_and_ht_plus_dp:Nnn,
  %     \box_gautosize_to_wd_and_ht_plus_dp:cnn
  %   }
  ```

  This also doesn't show up in the json file.

jlelong:
I know what happened. The specs were not clear to me. I did not understand that `\begin{function}[...]{the, function, list}` could span over several lines. I assumed everything was on a single line. I have to completely refactor my parsing code as i was reading all the files line by line.

jlelong:
I have updated the [list](https://gist.github.com/jlelong/9a49e2bc806f076905f6030faa052cdc#file-dtx-functions-variables-json). 
There are some strange entries such as `\::o`, `\::f_unbraced`, `\:::` and many more.
Could you check the list and list the entries that should be removed?

joulev:
> I have updated the [list](https://gist.github.com/jlelong/9a49e2bc806f076905f6030faa052cdc#file-dtx-functions-variables-json).

It is amazing! At least the list has passed all of my tests.

> There are some strange entries such as `\::o`, `\::f_unbraced`, `\:::` and many more.

The kernel team does document some internal functions that some may find useful, such as [`\:::`](https://github.com/latex3/latex3/blob/master/l3kernel/l3expan.dtx#L913)

```
% \begin{function}{\::n, \::N, \::p, \::c, \::o, \::e, \::f, \::x, \::v, \::V, \:::}
```

or [`\__kernel_chk_flag_exist:n`](https://github.com/latex3/latex3/blob/master/l3kernel/l3debug.dtx#L96)

```
% \begin{function}[EXP]{\__kernel_chk_flag_exist:n}
```

> Could you check the list and list the entries that should be removed?

There are weird functions like `\:::` in the list (there are 17 of them). There are also some functions that are explicitly named as internal functions like `\__kernel_chk_flag_exist:n` (there are 8 of them, all are in `l3debug.dtx`).

They *are* documented by the kernel team. However, I don't think they will be used anywhere outside the kernel itself. (The documentation of the `::` functions even has the following line: "They should not be used outside this module.") So I am not quite sure about whether these should be included.

For now I suggest not removing them yet. I will ask some LaTeX3 expert for advice on this, hopefully I will receive a reply soon.

Other than these 25, I don't think any other functions should be removed.

joulev:
A LaTeX3 package author wrote to me:

> no, I don't because they are just for internal code in the `l3exp` module. Don't touch them as a user. As long as you don't plan to contribute to the kernel you shouldn't use them.

So I guess besides `l3debug` functions, the 17 functions containing `::` should be removed as well.

jlelong:
To summarise all this, I should remove everything starting with `\::` (17 entries) but I keep the ones starting with `\__` (8 entries).

I will update the list and make it available for completion. More news later on today!

joulev:
> I keep the ones starting with `\__`

I suggest removing them as well.

jlelong:
Here is the updated [list](https://gist.github.com/jlelong/9a49e2bc806f076905f6030faa052cdc#file-dtx-functions-variables-json). 

I will integrate it into the intellisense mechanism.

jlelong:
One more question : how to merge the entries coming from the `.dtx` files with those extracted from the index of `interface3`? 

First, I thought that the `dtx` files would include everything, but such is not the case. For instance,  in the `dtx` files I find `\token_if_space:N` and from the index we came up with

```
 \token_if_space:NT
 \token_if_space:NF
 \token_if_space:NTF
 \token_if_space_p:N
```

Should I concatenate everything (and deduplicate)?

joulev:
image

As I can see, there are three optional arguments that can change the function name.

In the following section, I assume the function is `\function_name:<signature>` (and an accompanying example function `\foo:N`).

* `pTF` indicates four different functions:
  * `\function_name_p:<signature>` &mdash; `\foo_p:N`
  * `\function_name:<signature>(T|F|TF)` &mdash; `\foo:NT`, `\foo:NF`, `\foo:NTF`.
* `TF`: similar to `pTF`, except that the `_p` variant is removed. So it indicates three different functions: `\function_name:<signature>(T|F|TF)` &mdash; `\foo:NT`, `\foo:NF`, `\foo:NTF`.
* `noTF` indicates four different functions:
  * `\function_name:<signature>` &mdash; `\foo:N`
  * `\function_name:<signature>(T|F|TF)` &mdash; `\foo:NT`, `\foo:NF`, `\foo:NTF`.

So, once again, the parser needs more complication :/ it needs to also parse the optional arguments and make changes to the function name itself. However, I am pretty sure now that this will cover everything available.

jlelong:
I was running out of time these last two days but I have finally managed to generate a [new completion list](https://gist.github.com/jlelong/9a49e2bc806f076905f6030faa052cdc#file-expl3-commands-txt), which takes into account the optional variant parameter of the `function` and `variable` environments. If you prefer the dictionary [version].(https://gist.github.com/jlelong/9a49e2bc806f076905f6030faa052cdc#file-dtx-functions-variables-json)

Now that we parse the `.dtx` files, there is no need to also parse the `interface3.tex` index, which btw also faced the variants issues.

My feeling is that we can start with this completion list.
I will be merging everything shortly. @joulev Of course, feel free to give some feedback anytime.

joulev:
For now, I believe this is the whole list. I cannot find anything in interface3.pdf that is excluded from it.

However, during normal use, if I see some tricky bugs related to it, I will inform you 😃.

Issue #1939, opened 2020-02-07
Running Save Participants takes forever


juanhu96:
A minute ago I was still able to write latex file and save it (and auto-compile etc.), but then suddenly I could not save the latex file and could not compile it (I didn't even close the file or do anything). It says that 

> Running Save Participants for '~/Desktop/document.tex': Formatting

and it seems to take forever. I tried reinstalling VS code as well as the extension but it still doesn't work. Is there anything I could do?
Thanks!

tamuratak:
Fill out all the information required in the issue template. Please paste the whole log messages here, not parts of ones. It is very important to identify problems.

juanhu96:
## Preliminary questions [Required]

### Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: Yes


## Describe the bug [Required]

A minute ago I was still able to write latex file and save it (and auto-compile etc.), but then suddenly I could not save the latex file and could not compile it (I didn't even close the file or do anything). It says that

Running Save Participants for '~/Desktop/document.tex': Formatting

and it seems to take forever. I tried reinstalling VS code as well as the extension but it still doesn't work. Is there anything I could do?

### To Reproduce

Steps to reproduce the behavior:
1. Open a tex file in VS Code with the LaTex-Workshop extension enabled
2. Save the file
3. See error: 

> Running Save Participants for '~/Desktop/document.tex': Formatting

It neither saves the file nor compiles it.

### Expected behavior

A clear and concise description of what you expected to happen.

The tex file would be saved and it would also be complied, and I could view the resulting pdf file.


## Logs [Required]

Nothing

### LaTeX Workshop Output [Required]

```
To access the log, click the 'TeX' icon on the Activity Bar on the left side, select 'View Log Messages', then select 'View LaTeX Workshop extension log'.

[20:44:06] Initializing LaTeX Workshop.
[20:44:06] Creating LaTeX Workshop http and websocket server.
[20:44:07] LaTeX Workshop initialized.
[20:44:07] Server created on 127.0.0.1:57811
[20:44:07] LaTeX Workshop version: 8.7.0
[...]

PR #2080, opened 2020-05-12
Fix double '}}' in env completion

### LINKED ISSUE #2075: Autocomplete for '\begin{}' inserts a double closing '}'
### Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

Invalid syntax from autocomplete for \begin{itemize} or \begin{enumerate} causing compilation error.

### To Reproduce
When typing:
```\begin{i}```
The suggestions for autocomplete will appear. 
In this case, the relevant suggestion will be:
```\begin{itemize}```
If I proceed with this suggestion then I will have what follows:

```
\begin{itemize}
    \item 
\end{itemize}}
```

As you can see, the ending command contains an additional closing curly brace.
It's wrong and will cause a compilation error if not removed (manually).

### Expected behavior

I want the autocomplite to lead to what's below:
```
\begin{itemize}
    \item 
\end{itemize}
```

## Logs [Required]

Please paste the whole log messages here, not parts of ones. It is very important to identify problems.

```

[Paste the log here]
```

```
To access the log, click 'help' -> 'Toggle Developer Tools. Paste anything suspicious.

[Paste the log here]
```

## Screenshots

image

image

## Desktop [Required]
 - OS: Windows 10
 - VS Code version: [e.g. 1.45.0]
 - Extension version: [e.g. 8.9.0]
 - TeX distribution version: [e.g. TeX Live 2020]

Yes, I am using LaTeX Workshop

### PR DESCRIPTION
**Pull request by jlelong**
Close #2075

### CODE REVIEW

**Comment by jlelong** [comment]
Close #2075

File: src/providers/completer/environment.ts [hunk 1]
```diff
@@ -83,14 +83,24 @@ export class Environment {
...
             snippetType = EnvSnippetType.AsName
+        } else {
+            // If a closing '}' after '\begin{' has already been inserted, we need to remove it as it is already included in the snippets
+            const word = args.document.lineAt(args.position).text.slice(args.position.character - '\\begin{'.length, args.position.character + 2)
+            if (word === '\\begin{}') {
+                vscode.window.activeTextEditor.edit(e => {
+                    e.delete(new vscode.Range(args.position, args.position.translate(0, 1)))
+                })
+            }
+
```
tamuratak, line 101: We should use `CompletionItem.range`. 

Calling `TextEditor.edit`, a function with side effects, leads to a bug like #1939.

jlelong:
> We should use `CompletionItem.range`. 

- https://code.visualstudio.com/api/references/vscode-api#CompletionItem

Calling `TextEditor.edit`, a function with side effects, leads to a bug like #1939.

 

Good point! I had completely forgotten these side effect issues. I had already thought of `CompletionItem.range` but it forces us to update the every item in the suggestion list every time completion is triggered. I wanted to avoid that but I agree with you that this is probably the safest way to proceed.

src/providers/completer/environment.ts
jlelong: Good point! I had completely forgotten these side effect issues. I had already thought of `CompletionItem.range` but it forces us to update the every item in the suggestion list every time completion is triggered. I wanted to avoid that but I agree with you that this is probably the safest way to proceed.

jlelong: The new implementation uses `CompletionItem.range`.

PR #2085, opened 2020-05-14
add \includeonly completion

Jieli12:
add \includeonly completion

### DIFF
--- a/src/providers/completion.ts
+++ b/src/providers/completion.ts
@@ -155,7 +155,7 @@ export class Completer implements vscode.CompletionItemProvider {
                 break
 
             case 'input':
-                reg = /\\(input|include|subfile|includegraphics|lstinputlisting|verbatiminput)\*?(?:\[[^[\]]*\])*{([^}]*)$/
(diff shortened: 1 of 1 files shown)


src/providers/completion.ts
```diff
@@ -155,7 +155,7 @@ export class Completer implements vscode.CompletionItemProvider {
                 break
 
             case 'input':
-                reg = /\\(input|include|subfile|includegraphics|lstinputlisting|verbatiminput)\*?(?:\[[^[\]]*\])*{([^}]*)$/
+                reg = /\\(input|include|subfile|includegraphics|lstinputlisting|verbatiminput|includeonly)\*?(?:\[[^[\]]*\])*{([^}]*)$/
```
jlelong, line 158: Full support for `includeonly` is a bit more complicated that as it expects a coma separated list of files and files containing a coma in their names should be enclosed into `{}`

Jieli12: Yes, I noticed that. The items after comma can not be completed when one finished first completion. Would you please add this feature? I am not expert in this field.

jlelong: I have proposed a new implementation of this feature in #2088. So, I am closing this PR

Jieli12: Great! Thank you very much!

Issue #1971, opened 2020-02-29
Autobuild fails in subfolders in 8.7.0+


jmussman:
### Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]

You still see this issue?: Yes, and all other extensions have been left disabled during testing.

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: Yes

## Describe the bug [Required]

On my computer a build will not automatically run when a file in a subfolder is saved. It does work for the master document in the root folder. This happens on any version from 8.7.0 forward. I've checked 8.5.0 and 8.6.0 on my computer and they work properly. The build folder is never created. latex-workshop.latex.autoBuild.run is set to onFileChanged. The file is on the local disk. I've tried setting latex-workshop.latex.watch.usePolling to true, no effect. latex-workshop.latex.search.rootFiles.include is set to the default of **/*.tex. latex-workshop.latex.rootFile.useSubFile is set to true. latex-workshop.latex.rootFile.doNotPrompt is set to true (applies to 8.7.2 only of course).

### To Reproduce

1. Make sure the -cd option is added to the latexmk recipe.
2. Start with my template project: https://github.com/nextstepitt/standard-course-template.
3. Open Workbook/01_Rename_Me/section-0.tex.
4. Change and save the file.

### Expected behavior

I expect a subfile with a .tex extension to be compiled when the file is saved.

## Logs [Required]

Please paste the whole log messages here, not parts of ones. It is very important to identify problems.

### LaTeX Workshop Output [Required]

```
[18:01:53] Initializing LaTeX Workshop.
[18:01:53] Creating file watcher for .pdf files.
[18:01:53] Creating LaTeX Workshop http and websocket server.
[18:01:53] LaTeX Workshop initialized.
[18:01:53] Found root file of this subfile from active editor: /Shared/Projects/Course Projects/NS70000 Software Engineering/NS70500 Software Development Tools/NS70501 Software Development Tools/Workbook/workbook.tex
[18:01:53] Root file changed from: undefined to /Shared/Projects/Course Projects/NS70000 Software Engineering/NS70500 Software Development Tools/NS70501 Software Development Tools/Workbook/workbook.tex. Find all dependencies.
[...]

PR #2094, opened 2020-05-17
Add an option `onSuccess` to latex-workshop.latex.autoClean.run.

### LINKED ISSUE #1419: Clear only when compilation success
Seems the new "Auto Clean" setting only provides options to clean when compilation fails or terminated. Is it possible to add the option to clean when successes? Usually I need the log files to see why the compilation is failed, but I don't need them if it successes.

Also, it would be sweet if I can choose what files to be auto cleaned

tamuratak:
Add an option `onSuccess` to `latex-workshop.latex.autoClean.run`. Close #1419.

### DIFF
--- a/package.json
+++ b/package.json
@@ -985,15 +985,17 @@
           "enum": [
             "never",
             "onFailed",
+            "onSuccess",
             "onBuilt"
           ],
           "enumDescriptions": [
             "Never clean the project",
-            "Clean compilation fails",
+            "Clean when compilation fails",
+            "Clean when compilation succeeds",
             "Clean after build, be it successful or not"
           ],
           "default": "never",
-          "markdownDescription": "When LaTeX auxiliary files should be deleted.\nThe folder to be cleaned is defined in `latex-workshop.latex.outDir`.\n`onFailed` cleans the project when compilation fails. `onBuilt` cleans the project when compilation is terminated, whether successful or failed."
+          "markdownDescription": "When LaTeX auxiliary files should be deleted.\nThe folder to be cleaned is defined in `latex-workshop.latex.outDir`.\n`onFailed` cleans the project when compilation fails. `onSuccess` cleans the project when compilation succeeds. `onBuilt` cleans the project when compilation is terminated, whether successful or failed."
         },
         "latex-workshop.latex.clean.subfolder.enabled": {
           "type": "boolean",
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -376,7 +376,7 @@ export class Builder {
             this.extension.logger.addLogMessage('SyncTex after build invoked.')
             this.extension.locator.syncTeX(undefined, undefined, pdfFile)
         }
-        if (configuration.get('latex.autoClean.run') as string === 'onBuilt') {
+        if (['onSuccess', 'onBuilt'].includes(configuration.get('latex.autoClean.run') as string)) {
             this.extension.logger.addLogMessage('Auto Clean invoked.')
             this.extension.cleaner.clean(rootFile)
         }


jlelong:
I am very much bewildered by the request of an on success cleaning option for at least two reasons

1. The presence of auxiliary files speeds up compilation as it typically spares one or more calls of the compiler.
1. The extension relies on the `.fls` file to fully discover all dependencies as `latexmk` does. So always cleaning the auxiliary files may significantly deteriorate intellisense and build on change features.

I think these two major drawbacks should be clearly highlighted.

tamuratak:
I do not have a specific opinion. I leave aux files for daily use.

We can say the same thing for the `onBuilt` option. Especially if `2` happens, I think it is a bug. To my understanding, `2` does not occur since `parseFlsFile` is called before cleaning files.

https://github.com/James-Yu/LaTeX-Workshop/blob/983de450008711729cff4d1afa7c3b01369a4cc4/src/components/builder.ts#L372

https://github.com/James-Yu/LaTeX-Workshop/blob/983de450008711729cff4d1afa7c3b01369a4cc4/src/components/builder.ts#L363-L383

jlelong:
> I do not have a specific opinion. I leave aux files for daily use.
> We can say the same thing for the `onBuilt` option.

I always keep the auxiliary files too. I agree with you that `onBuild` and `onSuccess` options share the same drawbacks.

You are right about `.fls` files being parsed before the cleaning. Yet, `parseFls` is called when discovering the root file and in `parseFileAndSubs` 

https://github.com/James-Yu/LaTeX-Workshop/blob/aab062f55d5840cade1ea7534425320bfd0d6eb8/src/components/manager.ts#L421

The function `parseFileAndSubs` is called by

https://github.com/James-Yu/LaTeX-Workshop/blob/aab062f55d5840cade1ea7534425320bfd0d6eb8/src/components/manager.ts#L722-L732

before building. So I am afraid `2.` may happen more often than we think in particular when using `subfiles` #1971.

tamuratak:
I agree with you. I close this PR until treating `subfiles` get matured.

jlelong:
I am not saying that this PR should not be merged as all the issues occurring with `onSuccess` already exist with `onBuilt`. My point is just to warn users to avoid false issues.

> I close this PR until treating `subfiles` get matured.

The difficulty with `subfiles` is that users tend to use macros to define paths for inclusion and honestly this is the proper way to do it. So unless we want to perform macro expansion after `latexparser`, which I think is far beyond the scope of the extension, there is no real alternative to using `.fls` files, which are directly generated by the `latex` compiler.

tamuratak:
To our experience, making users avoid opening false issues is extremely difficult. When we find a solution, we should sell it to GitHub Inc at a high price. :)

We should find a sophisticated way to warn users. I can't think of a good idea, so I'll leave this PR as it is for a while.

stephenlevitt:
@jlelong 

>     1. The presence of auxiliary files speeds up compilation as it typically spares one or more calls of the compiler.
> 
>     2. The extension relies on the `.fls` file to fully discover all dependencies as `latexmk` does. So always cleaning the auxiliary files may significantly deteriorate intellisense and build on change features.

Hi, I wasn't aware of these drawbacks. My request for this feature was motivated by the fact that I prefer not having all of the "clutter" of these auxiliary files hanging around after I have successfully produced a document. Another option is to just mark these files as hidden by changing the file attribute, which can easily be done in a recipe. This would avoid the downsides that you mention. I guess this should be left for individual users to customize by themselves.

stephenlevitt:
Just a final comment on this issue. I have modified my build recipes to hide the auxiliary files that are generated so I don't have the clutter plus there is no negative affect on compilation speed/intellisense.

Nevertheless, after thinking about it, I feel the `onSuccess` option should exist for symmetry and for those that are prepared to take the compilation hit (which, as far as I could tell, is very minor for small documents).  And, as mentioned, a general warning should be given about cleaning auxiliary files, perhaps in the explanation in the Settings where users make the selection ("When LaTeX auxiliary files should be deleted. The folder to be cleaned...")

PR #2116, opened 2020-05-26
Fix keys to reflect command in intellisense

### LINKED ISSUE #2112: \left( does not match \left( ... \right) in suggestions.
`\left(` does not match `\left(...\right)` in suggestions. We have to type `\lrparen`.

jlelong:
Close #2112

### DIFF
--- a/data/commands.json
+++ b/data/commands.json
@@ -1,232 +1,232 @@
 {
-  "begin":{
-    "command":"begin",
-    "snippet":"begin{",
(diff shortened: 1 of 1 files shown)


jlelong:
This change restores the behaviour before #2104.
Yet, all this is not completely satisfying as trying to complete `\left(` yields `\\left(\right))`. The expected usage is to type `\left` and choose the correct completion in the list.

PR #2124, opened 2020-05-28
Add auto-close of inline math

tobiasdiez:
Add auto-closing of inline math: typing `\(` expands automatically the inline math snippet and yields `\( | \)` (`|` represents the cursor). Pressing tab again jumps after the closing bracket.

Moreover, I've added a new `begin` snippet which creates a new `\begin{...} ... \end{...}` pair. It's similar to the existing snippet that triggers on `\begin` just that you don't have to type the beginning slash.


tamuratak:
I am against putting any more things into the callback of `onDidChangeTextDocument`.

PR #2125, opened 2020-05-29
Define .bcf and .run.xml as auxiliary extensions

TakingItCasual:
I would define them locally under user settings, but apparently there's no way to append to the default list: https://stackoverflow.com/q/57796423

The file extensions are related to biblatex/biber.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1025,7 +1025,9 @@
             "*.snm",
             "*.synctex(busy)",
             "*.synctex.gz(busy)",
-            "*.nav"
+            "*.nav",
+            "*.bcf",
+            "*.run.xml"
           ],
           "markdownDescription": "Files to clean.\nThis property must be an array of strings. File globs such as *.removeme, something?.aux can be used."
         },


jlelong:
Adding values to a default list is not feasible directly, you have to redefine the whole list in your user settings.

New extensions should be added with care as they are glob patterns. Any file with `.bcf` or `.run.xml` extension in the current workspace will be cleaned regardless of them being related to a LaTeX project. I would like to have feedback on this before considering merging this PR.

TakingItCasual:
I've left a feature request for VS Code (https://github.com/microsoft/vscode/issues/98792). It's been assigned, so hopefully, this PR will become unnecessary.

jlelong:
Instead of dealing with extensions one by one, it would worth considering relying on `latexmk -C` to do the cleaning. That could be a new configuration option. Any thoughts on that?

TakingItCasual:
Running `latexmk -C` leaves behind `.bbl` and `.run.xml` files for me.

jlelong:
What `latexmk` cleans is highly configurable

```
       -c     Clean  up  (remove)  all regeneratable files generated by latex and bibtex or biber except dvi,
              postscript and pdf.  These files are a combination of log files, aux files, latexmk's  database
              file  of  source  file  information, and those with extensions specified in the @generated_exts
              configuration variable.  In addition, files specified by  the  $clean_ext  and  @generated_exts
              configuration variables are removed.

              This cleanup is instead of a regular make.  See the -gg option if you want to do a cleanup then
              a make.

              Treatment of .bbl files:  If $bibtex_use is set to 0 or 1, bbl files are always treated as non-
              regeneratable.  If $bibtex_use is set to 1.5, bbl files are counted as non-regeneratable condi-
              tionally: If the bib file exists, then bbl files are regeneratable, and are deleted in a  clean
              up.   But if $bibtex_use is 1.5 and a bib file doesn't exist, then the bbl files are treated as
              non-regeneratable and hence are not deleted.

              In contrast, if $bibtex_use is set to 2, bbl files are always treated as regeneratable, and are
              deleted in a cleanup.

              Treatment  of  files generated by custom dependencies: If $cleanup_includes_cusdep_generated is
              nonzero, regeneratable files are considered as including those generated by custom dependencies
              and are also deleted.  Otherwise these files are not deleted.


       -C     Clean  up (remove) all regeneratable files generated by latex and bibtex or biber.  This is the
              same as the -c option with the addition of dvi, postscript and pdf files, and  those  specified
              in the $clean_full_ext configuration variable.

              This cleanup is instead of a regular make.  See the -gg option if you want to do a cleanup than
              a make.

              See the -c option for the specification of whether or not .bbl files are treated as  non-regen-
              eratable or regeneratable.

              If $cleanup_includes_cusdep_generated is nonzero, regeneratable files are considered as includ-
              ing those generated by custom dependencies and are also deleted.  Otherwise these files are not
              deleted.
```

In your case, you would probably want to set `$bibtex_use = 2;` and `$clean_ext = "run.xml"` in your `latexmkrc` file.

PR #2154, opened 2020-06-12
Add section numbers in outline

### LINKED ISSUE #2152: Numbered document sections in structure view
### Background
As someone working on a medium-length thesis (100 pages or so), it is always a challenge to find back to where I'm occasionally reading or receiving feedback to where it is in the structure.
One element that makes this a lot easier is using numbered sections, whether they use PDF annotations or highlighting to make them, or if it's a list of sections and their associated comments.

However, this is exceedingly difficult to do without the numbered sections in the structure. It can surprisingly time-demanding and tricky to remember whether section Golf is before or after Hotel when the titles have no alphabetic sort, and to scan for the correct section.

### Proposed Feature
The addition of column-aligned section numbers for the document structure view. The source for it could be the AUX-file in order to accurately parse and extract the applicable tree.

Assuming that the current structure looks for section markers on its own, it can remain a fallback to a more metadata-rich view.

### Alternatives
An alternative approach is to distinguish depths to a higher degree. While a low amount of indentation is neat and tidy for deeper trees --- which is why I'm quite a fan of it --- it is also harder to distinguish between depth levels when quickly scanning a broad structure for _the next section_ or _the next subsection_.
While it doesn't make referencing equally easy, having different colours for different depths would make it visually easier to distinguish parts from chapters, chapters from sections, sections from subsections, and so forth. 

**Additional context**
[See what this can look like when using vim with vimtex](https://user-images.githubusercontent.com/234489/84277034-4525ac00-ab33-11ea-9427-9cdc531c8b3e.png)

jlelong:
This feature can be deactivated by setting `latex-workshop.view.outline.numbers.enabled: false`.

Close #2152

### DIFF
--- a/package.json
+++ b/package.json
@@ -1053,6 +1053,11 @@
           "default": true,
           "markdownDescription": "Show the labels in the outline/structure views. Reload VSCode after changing the value."
         },
+        "latex-workshop.view.outline.numbers.enabled": {
+          "type": "boolean",
+          "default": true,
+          "markdownDescription": "Show the sectioning numbers in the outline/structure views."
+        },
         "latex-workshop.view.autoFocus.enabled": {
           "type": "boolean",
           "default": false,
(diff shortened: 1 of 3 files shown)


thor:
@jlelong thank you! :)

PR #2158, opened 2020-06-15
Add invert PDF in dark mode only.

aterenin:
This creates an option which applies "invert PDF" when the operating system's dark mode setting is enabled. 

This is detected via `window.matchMedia(`(prefers-color-scheme: dark)`).matches`, which is the same mechanism VS Code internally uses for the `window.autoDetectColorScheme` option which auto-switches between light and dark themes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1280,7 +1280,7 @@
             "compat",
             "never"
           ],
-          "default": "never",
+          "default": "compat",
           "markdownDescription": "Enable the CSS invert filter.",
           "enumDescriptions": [
             "Enable the invert filter when using a dark theme.",
--- a/package.json
+++ b/package.json
@@ -1223,6 +1223,11 @@
           "default": 0,
           "markdownDescription": "Define the CSS sepia filter level of the PDF viewer when the invert mode is enabled. Possible values are from 0 to 1."
         },
+        "latex-workshop.view.pdf.invertMode.darkModeOnly": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "Apply CSS invert filter for the PDF viewer in dark mode only.\nDefaults to `false`, which applies invert filter, if it is enabled, in both dark and light mode."
+        },
         "latex-workshop.view.pdf.backgroundColor": {
           "type": "string",
           "default": "#ffffff",
--- a/package.json
+++ b/package.json
@@ -1272,7 +1272,24 @@
           "default": false,
           "markdownDescription": "Define if the hand tool is enabled by default in the PDF viewer."
         },
-        "latex-workshop.view.pdf.invert": {
+        "latex-workshop.view.pdf.invertMode.enabled": {
+          "type": "string",
+          "enum": [
+            "auto",
+            "always",
+            "compat",
+            "never"
+          ],
+          "default": "never",
+          "markdownDescription": "Enable the CSS invert filter.",
+          "enumDescriptions": [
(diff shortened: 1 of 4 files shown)


tamuratak:
With this PR, the PDF Viewer does not get invert-mode when the theme of VS Code is dark and the mode of OS is not dark. It would be a good idea to detect the darkness of VS Code, not OS. 

See
- https://github.com/James-Yu/LaTeX-Workshop/blob/6bb95c2058bb2e5aaefccc37df04b174bab89d95/src/utils/theme.ts#L50 
- https://code.visualstudio.com/api/references/vscode-api#window.activeColorTheme

aterenin:
Theme is now detected via `getCurrentThemeLightness()`. Let me know if this is OK.

tamuratak:
No. `viewer/latexworkshop.ts` is executed in the browser context. You can not import `src/utils/theme`, which uses Node API.

aterenin:
D'oh. So much for quickly trying to hack things together.

The conditional for `darkModeOnly` is now evaluated in the handler, which, depending on `getCurrentThemeLightness` either sends the user's invert value or zero.

aterenin:
@tamuratak Any further comments? Is there still an issue with the current code?

tamuratak:
Sorry for the delay in replying.

I was thinking of making the option, `"latex-workshop.view.pdf.invertMode.darkModeOnly"`, more intuitive. For example, make a new option `"latex-workshop.view.pdf.invertMode.enabled"`, which can be "auto", "always", "compat", and "never". 

`"auto"` is the same as your `darkModeOnly`. `"always"` enables the invert mode even when `invert === 0`. `"compat"` behaves in the same way as now, enabled only when `invert > 0`.

No conclusions have been reached yet. What do you think?

aterenin:
That makes sense, and sounds more intuitive. Should I implement it?

tamuratak:
I would appreciate you if you would implement it.

aterenin:
> I would appreciate you if you would implement it.

Implemented! Let me know how this looks now.

package.json
```diff
@@ -1272,7 +1272,24 @@
...
-        "latex-workshop.view.pdf.invert": {
+        "latex-workshop.view.pdf.invertMode.enabled": {
+          "type": "string",
+          "enum": [
+            "auto",
+            "always",
+            "compat",
+            "never"
+          ],
+          "default": "never",
```
tamuratak, line 1283: The default value should be "compat"

src/components/viewer.ts
```diff
@@ -407,6 +408,10 @@ export class Viewer {
                         continue
                     }
                     const configuration = vscode.workspace.getConfiguration('latex-workshop')
+                    const invertType = configuration.get('view.pdf.invertMode.enabled') as string
+                    const invertEnabled = (invertType == "auto" && (getCurrentThemeLightness() == 'dark')) ||
+                        invertType == "always" ||
+                        (invertType == "compat" && ((configuration.get('view.pdf.invert') as number) > 0))
```
tamuratak, line 414: Run `npm run lint:fix`

package.json
aterenin: Fixed!

src/components/viewer.ts
aterenin: Done!

tamuratak:
Thank you, again.

PR #2177, opened 2020-06-26
Use kpsewhich to resolve bibliography files

### LINKED ISSUE #2163: Can we make "latex-workshop.latex.bibDirs" search BIBINPUTS?
Right now I have the environment variable `BIBINPUTS` set to path.
To make things work in VS Code I need to also set the explicit path.  Can we just make "latex-workshop.latex.bibDirs" automatically search the `BIBINPUTS` path? This seems far more robust and logical. After all the whole point of `BIBINPUTS` is to set one location for all your `.bib` and then forget about their path in other programs.

jlelong:
When a `.bib` cannot be found in the root directory or using `bibDirs`, we use `kpsewhich -format=.bib bib_file_name`.

Fix #2163.

### DIFF
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -2,6 +2,7 @@ import * as vscode from 'vscode'
 import * as os from 'os'
 import * as path from 'path'
 import * as fs from 'fs-extra'
+import * as cs from 'cross-spawn'
 import * as chokidar from 'chokidar'
 import * as micromatch from 'micromatch'
 import {latexParser} from 'latex-utensils'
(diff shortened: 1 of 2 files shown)


src/components/manager.ts
```diff
@@ -865,13 +866,33 @@ export class Manager {
         }
     }
 
+    private kpsewhichBibPath(bib: string): string | undefined {
+        const kpsewhich = vscode.workspace.getConfiguration('latex-workshop').get('kpsewhich.path') as string
+        try {
+            const kpsewhichReturn = cs.sync(kpsewhich, ['-format=.bib', bib])
```
tamuratak, line 872: Add a log message before calling kpsewhich. 

Please consider a case that kpsewhich does not immediately return.

src/components/manager.ts
```diff
@@ -865,13 +866,33 @@ export class Manager {
...
+
     private resolveBibPath(bib: string, rootDir: string) {
         const bibDirs = vscode.workspace.getConfiguration('latex-workshop').get('latex.bibDirs') as string[]
         const bibPath = utils.resolveFile([rootDir, ...bibDirs], bib, '.bib')
 
         if (!bibPath) {
             this.extension.logger.addLogMessage(`Cannot find .bib file: ${bib}`)
-            return undefined
+            this.extension.logger.addLogMessage('Trying to use kpsewhich')
+            return this.kpsewhichBibPath(bib)
```
tamuratak, line 895: I am not sure it is a good idea calling kpsewhich by default.

jlelong: I could add a new setting to use `kpsewhich` and completely ignore the `bibDirs` setting.

src/components/manager.ts
jlelong: What do you have in mind exactly here? Using an asynchronous call with a timeout?

tamuratak: What I am trying to say is that we should add log messages before executing commands. If there is no log message when the command blocks, we have no hint to debug. This is not specific for kpsewhich.

src/components/manager.ts
jlelong: In the end, I think it is better to always use the `bibDirs` setting and to only use `kpsewhich` when `latex-workshop.kpsewhich.enabled` is true (default value is false).

src/components/manager.ts
jlelong: The `cs.sync` function accepts a timeout option. It may be safe to use this option whenever a child process is spawned  synchronously. What do you think of it?

tamuratak: I do not have a specific opinion.

jlelong:
See https://github.com/James-Yu/LaTeX-Workshop/wiki/Intellisense#citations for explanations on how bibliography files are resolved. This PR introduces two new settings

- [`latex-workshop.kpsewhich.enabled`](https://github.com/James-Yu/LaTeX-Workshop/wiki/Intellisense#latex-workshopkpsewhichenabled)
- [`latex-workshop.kpsewhich.path`](https://github.com/James-Yu/LaTeX-Workshop/wiki/Intellisense#latex-workshopkpsewhichpath)

Issue #2126, opened 2020-05-29
Syntax file needs cleanup


tobiasdiez:
### Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: Yes

## Describe the bug [Required]

The "LaTeX" syntax currently has a few problems. These lead to strange colorization and problems in many themes.

- `keyword.control` vs `support.function`
   Some standard commands (such as `\cite`) are tagged as `keyword.control`. This category is usually reserved for keywords related to flow control like continue, while, return, etc. I think `support.function` would be a better scope, e.g. change `keyword.control.cite.latex` to `support.function.cite.latex`. Similar remarks apply to `keyword.control.label.latex`, `keyword.other.item.latex`, `keyword.control.ref.latex`

- Commands in math
   Commands in math mode are tagged as `constant.character.math.tex` or `constant.other.general.math.tex`. For some one could argue that this is indeed the case (`\pi` etc) but for others (`\bigr`) this is definitely the wrong category. Moreover, I feel that `support.function` might be the better category anyway (since for Latex there is no distinction between `\pi` and `\cite`). To provide special colorization of math environments one could add a `meta.math` tag.

- Usage of `support.class`
Latex does not know the concept of a class, and thus this should be converted to `support.function`.

- Incontinence between `tex` and `latex` ending
For example there are `punctuation.definition.function.tex` and `punctuation.definition.function.latex`

- Command arguments
   Sometimes they are tagged as `variable.parameter.function.latex` (e.g `\begin{...}`) and sometimes as `constant.other.reference.citation.latex` (e.g. `\parencite{...}`). It would be good to add (different) meta tags for optional and manditory arguments as well.

- Special scopes missing
  - equals sign, minus, underscore and comma in math (punctation?)
  - braces as command argument delimiters
  - quotes in text

- Wrong classification
   `\,` yields a space in latex but is tagged as `constant.character.escape.tex`. Moreover, everything (even commands environments etc) are tagged as `text.tex.latex`.

### To Reproduce

1. Open an tex file
2. Use "Inspect editor tokens" to view the scope.

### Expected behavior

[...]

PR #2192, opened 2020-07-02
Improve scope names

jlelong:
**Warning: This is PR must not be merged.** It is basically intended for discussing scope names. 

This PR is related to #2126. 

The commit messages should be sufficiently self explanatory to make clear what the changes are.

The most noticeable change is that math text is no special highlighting anymore. This is induced by the use of `meta.math` instead of `support.class.math`.

Before: 
Capture d’écran 2020-07-02 à 15 18 16

After: 
Capture d’écran 2020-07-02 à 15 18 02

### DIFF
--- a/syntax/LaTeX.tmLanguage.json
+++ b/syntax/LaTeX.tmLanguage.json
@@ -1524,13 +1524,14 @@
               "name": "punctuation.definition.optional.arguments.begin.latex"
             },
             "2": {
-              "name": "variable.parameter.latex"
+              "name": "variable.parameter.function.latex"
             },
             "3": {
               "name": "punctuation.definition.optional.arguments.end.latex"
             }
           },
-          "match": "(\\[)([^\\[]*?)(\\])"
+          "match": "(\\[)([^\\[]*?)(\\])",
+          "name": "meta.parameter.optional.latex"
         }
       ]
     },
--- a/syntax/LaTeX.tmLanguage.json
+++ b/syntax/LaTeX.tmLanguage.json
@@ -553,11 +553,11 @@
           "begin": "%",
           "beginCaptures": {
             "0": {
-              "name": "punctuation.definition.comment.tex"
+              "name": "punctuation.definition.comment.latex"
             }
           },
           "end": "$\\n?",
-          "name": "comment.line.percentage.tex"
+          "name": "comment.line.percentage.latex"
         },
         {
           "include": "source.gnuplot"
@@ -646,11 +646,11 @@
         {
           "captures": {
             "1": {
-              "name": "punctuation.definition.function.tex"
+              "name": "punctuation.definition.function.latex"
             }
           },
           "match": "(\\\\)[A-Za-z]+",
-          "name": "support.function.general.tex"
+          "name": "support.function.general.latex"
         }
       ]
     },
@@ -673,7 +673,7 @@
           "name": "punctuation.definition.arguments.end.latex"
         }
       },
-      "contentName": "punctuation.definition.comment.tex",
+      "contentName": "punctuation.definition.comment.latex",
       "end": "((\\\\)end)(\\{)(\\4)(\\})",
       "name": "meta.function.verbatim.latex"
     },
@@ -961,7 +961,7 @@
           "name": "punctuation.definition.begin.latex"
         },
         "4": {
-          "name": "support.function.general.tex"
+          "name": "support.function.general.latex"
         },
         "5": {
           "name": "punctuation.definition.function.latex"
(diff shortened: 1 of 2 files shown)


tamuratak:
What is the reason you think the latter of highlighting is better?

jlelong:
I am not sure this is better. The rendering is closer to the one in Sublime.
I think some commits are good improvements

- Use .latex as suffix scope for latex syntax
- Use support.function for standard commands
- Match \, and \; as as support.function
- Add string.other.math to highlight math delimiters

For the rest, this is a question of taste and it is unclear to me.

tamuratak:
As a user, I do not want highlighting to change without specific reasons. The consistency of the names of tokens does not justify the change. For a user getting used to the current highlighting, it is the correct one, even if it is not correct in semantics.

jlelong:
@tamuratak I think you are right. I will only keep the following changes 

- Use `.latex` as suffix scope for the latex syntax (no visual changes here)
- Match `\,` and `\;` as as support.function

I personally prefer the second screenshot using `meta.math` instead of `support.class.math` but I do agree that this purely a matter of taste and we should not change highlighting for this reason. So I will not keep this commit for now.

tamuratak:
Thank you.

tobiasdiez:
The syntax highlighting strongly depends on the chosen theme. Because of the inconsistency of the scope names, latex text looks rather awful in many themes. Moreover, some other functions and extensions (e.g. spellchecker) rely on the scope. For these reasons I would argue it is important that the scope names match with their intended use.

As far as I understand the VS code architecture, the extension should specify the scope names and the code styling is delegated to themes. Thus, if you would like to keep (or change) the way latex code looks, then a designated latex theme would be a good choice I guess.

jlelong:
> As far as I understand the VS code architecture, the extension should specify the scope names and the code styling is delegated to themes. Thus, if you would like to keep (or change) the way latex code looks, then a designated latex theme would be a good choice I guess.

The user chooses its preferred theme and there is no way to keep the old rendering if we modify the scopes.

tobiasdiez:
What I meant is that (to my understanding) a language extensions shouldn't worry about the code styling and instead specify the correct scopes. The styling is the job of the theme. 

Anyway, if the user prefers the old styling then he can specify something along the following lines in his settings:
```
    "editor.tokenColorCustomizations": {
        "[Default Light+]": {
            "textMateRules": [
                {
                    "scope": "meta.math",
                    "settings": {
                        "foreground": "#267F99"
                    }
                }
            }
     }
```

jlelong:
I have updated my comment to make it clearer and indeed only the user can customize the theme colors.

jlelong:
> What I meant is that (to my understanding) a language extensions shouldn't worry about the code styling and instead specify the correct scopes. The styling is the job of the theme.

From a design perspective I do agree with you but legacy makes things less clearer. If the styling changes, we will get a lot of issues complaining or at least discussing these changes, which I really do not want. 

From your original post, here is what I suggest to address

- [ ] Fix the `.tex` vs `.latex` scope suffix
- [ ] Add the `meta.math` scope on top of `support.class.math` for maths regions. This will not change highlighting unless the user explicitly customize the `meta.math` scope
- [ ] Fix classification of `\,` and match `\;`
- [ ] Highlight maths delimiters
- [ ] Add a meta tag for optional arguments of standard commands
- [ ] Do not highlight file names in `\input, `\include`, `\VerbatimOut` to remove the use of `support.class.latex`. Any suggestion of an other scope?
- [ ] Highlight `+,-,*,/,_,^` in math mode

What I do not plan to address for now

- `keyword.control` vs `support.function`
- `constant.character.math.tex` vs `constant.other.general.math.tex`
- arguments of cite commands using `constant.other.reference.citation.latex`

jlelong:
I would love to have feedback on these changes. There should be very little visible differences for the user. I would like to merge/close this PR and the related issue by the end of the week as I do not really have more time to devote to it.
@tamuratak What is your opinion on the changes? 

It seems that when assigning two scopes to an element, the first one (going from left to right) that is actually defined by the theme takes precedence on the other. However, I did not find any documentation on that.

tamuratak:
I am against changing the color of `$`. 

With `Default Light+`.

Before:

スクリーンショット 2020-07-08 17 37 55

After:

スクリーンショット 2020-07-08 17 36 04

jlelong:
OK, the commit is reverted. I am afraid we cannot do better without altering the current styling.

@tobiasdiez If one specific scope really creates an issue when using other extension, please open a new issue for this one and will try yo figure out what we can do. For the moment, I am merging this PR and closing the related issue.

PR #2193, opened 2020-07-04
Add command to toggle autobuild

mkvoya:
Hi everyone,

This PR adds a new command `latex-workshop.toggleAutoBuild` which can temporarily disable (and re-enable) the auto build functionality.

The feature can be useful when someone has a bunch of modifications to apply in a large LaTeX project. He/she may need to temporarily disable the auto build feature until he/she has finished all the modifications.

Hope this helps. :)

### DIFF
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -52,6 +52,11 @@ interface Content {
     }
 }
 
+const enum BuildEvents {
+    never = 'never',
+    onFileChange = 'onFileChange'
+}
+
 export class Manager {
     /**
      * The content cache for each LaTeX file.
@@ -870,7 +875,7 @@ export class Manager {
 
     private buildOnFileChanged(file: string, bibChanged: boolean = false) {
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
-        if (configuration.get('latex.autoBuild.run') as string !== 'onFileChange') {
+        if (configuration.get('latex.autoBuild.run') as BuildEvents !== BuildEvents.onFileChange) {
             return
         }
         if (this.disableAutoBuild) {
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -70,6 +70,7 @@ export class Manager {
     private jlweaveExt: string[] = ['.jnw', '.jtexw']
     private weaveExt: string[] = []
     private pdfWatcherOptions: chokidar.WatchOptions
+    private disableAutoBuild: boolean = false
 
     constructor(extension: Extension) {
         this.extension = extension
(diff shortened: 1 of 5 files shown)


jlelong:
Why not waiting for all the modifications to be done before saving all the files at once (`cltr-alt-s`)?

mkvoya:
Not all modifications can be done in vscode. For example, figures are edited externally.
Also, after I sync my git repo, I probably want to change something before the build process starts.
Finally, isn't frequent `ctrl-s` a good habit to prevent data loss caused by power failures? :)

src/components/manager.ts
```diff
@@ -840,11 +841,20 @@ export class Manager {
         }
     }
 
+    // This function toggles the autobuild state and returns whether autobuild is enabled.
+    toggleAutoBuild() {
+        this.disableAutoBuild = !this.disableAutoBuild
+        return !this.disableAutoBuild
```
jlelong, line 847: I think having two settings to control the same feature is a bad idea.
You should rather change the value of `latex.autoBuild.run` directly.

jlelong:
FYI, vscode has an internal backup mechanism to save your edits every time you stop typing for more than 1 second. If you restart vscode after a crash or power failure, your files will be marked as dirty because the non saved edits were restored from the backups.

src/components/manager.ts
mkvoya: Agreed. But there are two problems that hinder the use of `latex.autoBuild.run`. 1) Currently `latex.autoBuild.run` is a string and `onFileChange` is the only option that is meaningful. But I don't know whether other new options will be added to `latex.audoBuild.run` and conflict with the auto build toggle. 2) `latex.autoBuild.run` is a user configuration retrieved by `vscode.workspace.getConfiguration`. It seems that it's not possible to change user configurations via code (at least I didn't get an approach when I googled last few days).

jlelong: For 2), you could use the `update` method, see https://code.visualstudio.com/api/references/vscode-api#WorkspaceConfiguration

For 1), you have to assume that the current two values will not be changed.
The proper way to handle this would be turn `latex-workshop.latex.autoBuild.run` into a boolean `latex-workshop.latex.autoBuild.enabled`. Let me think about it.

jlelong: As suggested in #2209, you could define

```ts
const enum BuildEvents {
    never = 'never',
    onFileChange = 'onFileChange'
}
```
Let me know when your code is refactored.

mkvoya: Just finished my refactoring.
By the way, I rebased the commits to the latest master branch and chose to use the clock icon in the sidebar.
😄

jlelong:
The new option `onSave` has been added to `latex-workshop.latex.autoBuild.run` (#2211) to trigger build only file save. Changes outside VSCode do not trigger build with this configuration. If for some reason you want to save a file without triggering a build, use `latex-workshop.saveWithoutBuilding`, see https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ#I-use-build-on-save-but-I-occasionally-want-to-save-without-building.

This may actually solve your problem, does it?

mkvoya:
> The new option `onSave` has been added to `latex-workshop.latex.autoBuild.run` (#2211) to trigger build only file save. Changes outside VSCode do not trigger build with this configuration. If for some reason you want to save a file without triggering a build, use `latex-workshop.saveWithoutBuilding`, see https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ#I-use-build-on-save-but-I-occasionally-want-to-save-without-building.
> 
> This may actually solve your problem, does it?

Mostly, except for the frequent `ctrl-s`. But I am personally okay with that.
We could close this if you think it's okay.

jlelong:
Note that you can assign your own keybinding to `latex-workshop.saveWithoutBuilding`.

PR #2203, opened 2020-07-13
Refactor and add above/below to view.pdf.tab.editorGroup

### LINKED ISSUE #2200: Allow "top"/"above" and "bottom"/"below" with view.pdf.tab.editorGroup setting.
I'm using Code on a monitor in portrait mode which is nice for programming. However this renders the values `left` and `right` of `latex-workshop.view.pdf.tab.editorGroup` useless. The remaining value `current` is not what I want, I'd prefer to open the PDF viewer in a manner that keeps my current editor focused.

I'd like the values  `"top"` (or `"above"`) and `"bottom"` (or `"below"`) added to the `latex-workshop.view.pdf.tab.editorGroup` setting, splitting the current tab vertically instead of horizontally.

None.

**Additional context**
None.

jlelong:
This PR adds two new positions for the viewer tab: `above` and `below`.

This PR changes how to handle the viewer placement. So far, a new editor group was created to hold the viewer and then we tried to fiddle with the group placement. Now, we create the viewer in the same group as the `.tex` file and we rely on vscode internal functions to move the viewer. This leads to a far more robust placement when several there are editor groups.

Close #2200

### DIFF
--- a/package.json
+++ b/package.json
@@ -1166,13 +1166,17 @@
           "enum": [
             "current",
             "left",
-            "right"
+            "right",
+            "above",
+            "below"
           ],
           "markdownDescription": "The editor group in which to open the tab viewer.",
           "enumDescriptions": [
             "Use the current editor group",
             "Put the viewer tab in a new group on the left of the current one",
-            "Put the viewer tab in a new group on the right of the current one"
+            "Put the viewer tab in a new group on the right of the current one",
+            "Put the viewer tab in a new group above the current one",
+            "Put the viewer tab in a new group below the current one"
           ]
         },
         "latex-workshop.view.pdf.ref.viewer": {
--- a/src/components/viewer.ts
+++ b/src/components/viewer.ts
@@ -203,7 +203,7 @@ export class Viewer {
      * @param respectOutDir
      * @param tabEditorGroup
      */
-    openTab(sourceFile: string, respectOutDir: boolean = true, tabEditorGroup: string) {
+    async openTab(sourceFile: string, respectOutDir: boolean = true, tabEditorGroup: string) {
         const url = this.checkViewer(sourceFile, respectOutDir)
         if (!url) {
             return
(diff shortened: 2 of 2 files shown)


jlelong:
@tamuratak Can you test this new approach?

tamuratak:
This PR has a timing issue.

Jul-14-2020 11-00-47

Version: 1.47.0
Commit: d5e9aa0227e057a60c82568bf31c04730dc15dcd
Date: 2020-07-09T08:01:54.115Z
Electron: 7.3.2
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Darwin x64 17.7.0

See a fix:
- https://github.com/tamuratak/LaTeX-Workshop/commit/800359beba4e0b33e4a2acd1501ec8e6489be8a1

jlelong:
Calling

```ts
await vscode.window.showTextDocument(editor.document, editor.viewColumn, false)
```

instead of 

```ts
await vscode.window.showTextDocument(editor.document, editor.viewColumn, false)
```

fixes the tests but it does not work when latex-workshop.view.pdf.tab.editorGroup is set to left or above because it opens the .tex file in the wrong editor group because the editor groups are reordered.

jlelong:
It seems that relying on the `workbench.action.focusXXXXGroup` action to move the focus back to the `.tex` editor fixes everything but I do not really understand why.

tamuratak:
I have confirmed that it works well.

jlelong:
And yet, the viewer tests are still broken under Windows 
https://github.com/James-Yu/LaTeX-Workshop/pull/2203/checks?check_run_id=868175722

tamuratak:
The test on windows sometimes fails. Rerunning the test gets it to succeed.

tamuratak:
The test still fails. Using `executeCommand('workbench.action.moveEditorToAboveGroup')` and others might be too tricky. Since I'm not interested in this feature, I don't care if it doesn't get merged.

jlelong:
It seems that 6799c3a fixes it even under Windows.
I am not that much interested in adding `above/below` either but rather in refactoring to remove the dirty hack I was using for the `left` value.

tamuratak:
That's definitely interesting.

> It seems that 6799c3a fixes it even under Windows.

Because we can skip tests only on windows if necessary, I am not against merging.

jlelong:
> It seems that 6799c3a fixes it even under Windows. I am not sure how reliable that is.

Unreliable! That commit helps passing some of the viewer tests but not all. The tests eventually break, not always on the same one but always on a timeout error. 

I am currently testing to explicitly call `vscode.window.showTextDocument(editor.document, vscode.ViewColumn.Active)` once all the editors are properly placed.

PR #2232, opened 2020-08-05
Fix XML & HTML scopes

vfcp:
`LaTeX.tmLanguage.json` is referring to incorrect scope for XML. Correct XML scope is `"source.xml"`.

Also, `package.json` lists `"source.xtml": "xtml"` but there's no reference for `xtml` inside `LaTeX.tmLanguage.json`. Embedded language should be `"source.xml": "xml"`


vfcp:
Thanks! I've changed the grammar to correct scope `text.xml`.

jlelong (approved):

PR #2261, opened 2020-09-01
Declare a formatter provider for bibtex

### LINKED ISSUE #2250: Formatting bibtex files
LaTeX-Workshop has a set of commands format Bibtex files (https://github.com/James-Yu/LaTeX-Workshop/wiki/Format#Bibtex-files). However, they are only functions you can manually call, not a formatting feature you can use as a formatter (by VSCode formatting interface, e.g. on saving file). LaTeX-Workshop sets themselves as a default formatter for `.bib` files, but formats them with regular LaTeX formatter which can be unexpected (e.g. breaks aligning).

I propose to set e.g. `latex-workshop.bibalignsort` as a command used to formatting `.bib` file. 

Since there are three commands that can be used to formatting, it would be nice to have a configuration which will be used to formatt `.bib` files. If it's possible, it could be a separated formatter entry for Bibtex files, so user could decide which (regular LaTeX as now or Bibtex-specyfic) formatter want to use by selecting it in VSCode formatter configuration window.

jlelong:
Close #2250

### DIFF
--- a/src/providers/bibtexformater.ts
+++ b/src/providers/bibtexformater.ts
@@ -64,7 +64,17 @@ export class BibtexFormatter {
         const lineOffset = range ? range.start.line : 0
         const columnOffset = range ? range.start.character : 0
 
-        const ast = await this.extension.pegParser.parseBibtex(document.getText(range))
+        const ast = await this.extension.pegParser.parseBibtex(document.getText(range)).catch((error) => {
+            if (error instanceof(Error)) {
+                this.extension.logger.addLogMessage('Bibtex parser failed.')
+                this.extension.logger.addLogMessage(error.message)
+                this.extension.logger.showErrorMessage('Bibtex parser failed with error: ' + error.message)
+            }
+            return undefined
+        })
+        if (! ast) {
+            return []
+        }
         // Create an array of entries and of their starting locations
         const entries: bibtexParser.Entry[] = []
         const entryLocations: vscode.Range[] = []
--- a/src/providers/bibtexformater.ts
+++ b/src/providers/bibtexformater.ts
@@ -58,6 +58,8 @@ export class BibtexFormatter {
             sort: config.get('bibtex-format.sortby') as string[]
         }
 
+        const lineOffset = range ? range.start.line : 0
+
         // Create an array of entries and of their starting locations
         const entries: bibtexParser.Entry[] = []
         const entryLocations: vscode.Range[] = []
@@ -118,13 +120,13 @@ export class BibtexFormatter {
                 } else { // 'Comment Duplicates'
                     // Log duplicate entry since we aren't highlighting it
                     this.extension.logger.addLogMessage(
-                        `BibTeX-format: Duplicate entry "${entries[i].internalKey}" at line ${entryLocations[i].start.line + lineDelta + 1}.`)
+                        `BibTeX-format: Duplicate entry "${entries[i].internalKey}" at line ${entryLocations[i].start.line + lineDelta + 1 + lineOffset}.`)
                     text = text.replace(/@/,'')
                 }
             }
 
             // Put text from entry[i] into (sorted)location[i]
-            edits.push(new vscode.TextEdit(entryLocations[i], text))
+            edits.push(new vscode.TextEdit(new vscode.Range(entryLocations[i].start.translate(range?.start.line, range?.start.character), entryLocations[i].end.translate(range?.start.line)), text))
 
             // We need to figure out the line changes in order to highlight properly
             lineDelta += (sortedEntryLocations[i].end.line - sortedEntryLocations[i].start.line) - (entryLocations[i].end.line - entryLocations[i].start.line)
--- a/src/providers/bibtexformater.ts
+++ b/src/providers/bibtexformater.ts
@@ -18,7 +18,12 @@ export class BibtexFormatter {
     }
 
     async bibtexFormat(sort: boolean, align: boolean) {
-        if (vscode.window.activeTextEditor === undefined || vscode.window.activeTextEditor.document.languageId !== 'bibtex') {
+        if (!vscode.window.activeTextEditor) {
+            this.extension.logger.addLogMessage('Exit formatting. The active textEditor is undefined.')
+            return
+        }
+        if (vscode.window.activeTextEditor.document.languageId !== 'bibtex') {
+            this.extension.logger.addLogMessage('Exit formatting. The active textEditor is not of bibtex type.')
             return
         }
         const doc = vscode.window.activeTextEditor.document
@@ -59,7 +64,10 @@ export class BibtexFormatter {
         }
 
         const lineOffset = range ? range.start.line : 0
+        const columnOffset = range ? range.start.character : 0
 
+        this.extension.logger.addLogMessage('Start bibtex formatting.')
+        const ast = await this.extension.pegParser.parseBibtex(document.getText(range))
         // Create an array of entries and of their starting locations
         const entries: bibtexParser.Entry[] = []
         const entryLocations: vscode.Range[] = []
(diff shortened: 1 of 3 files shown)


jlelong:
The bibtex parser logs some information/warning while parsing the `.bib` content. In 8e6a2bd, these logs are further turned into diagnostics and showed in the _Problems pane_. However, even it is fixed, the warning remains in the _Problems pane_, which is a bit weird. This is exactly as when using the _Sort bibliography_ command from the TeX Panel
@tamuratak do you have any idea of to handle this?

tamuratak:
I need a minimal working example.

src/providers/bibtexformater.ts
```diff
@@ -67,56 +96,75 @@ export class BibtexFormater {
...
+export class BibtexFormatterProvider implements vscode.DocumentFormattingEditProvider, vscode.DocumentRangeFormattingEditProvider {
+    private formatter: BibtexFormatter
+    extension: Extension
+
+    constructor(extension: Extension) {
+        this.extension = extension
+        this.formatter = new BibtexFormatter(extension)
+    }
+
+    public provideDocumentFormattingEdits(document: vscode.TextDocument, _options: vscode.FormattingOptions, _token: vscode.CancellationToken): vscode.ProviderResult<vscode.TextEdit[]> {
```
tamuratak, line 154: We should make `provideDocumentFormattingEdits` async.

src/providers/bibtexformater.ts
```diff
@@ -67,56 +96,75 @@ export class BibtexFormater {
...
+        return this.formatter.formatDocument(document, true, true).then(edits => {
+            this.formatter.duplicatesDiagnostics.set(document.uri, this.formatter.diags)
+            this.extension.logger.addLogMessage('BibTeX action successful.')
+            return edits
+            })
     }
+
+    public provideDocumentRangeFormattingEdits(document: vscode.TextDocument, range: vscode.Range, _options: vscode.FormattingOptions, _token: vscode.CancellationToken): vscode.ProviderResult<vscode.TextEdit[]> {
```
tamuratak, line 162: We should make `provideDocumentRangeFormattingEdits` async too.

tamuratak (changes requested):

src/providers/bibtexformater.ts
jlelong: It does not seem possible to declare a function with return type `vscode.ProviderResult<T>` async. We would need to change the return type to `Promise<T>` but https://code.visualstudio.com/api/references/vscode-api#DocumentFormattingEditProvider

Btw, we do not use `async` for the latex formatter 
https://github.com/James-Yu/LaTeX-Workshop/blob/2ad71efb30fdd6a9b03f4ab3d7bb24b567928cc2/src/providers/latexformatter.ts#L210-L219

jlelong:
Consider the badly formatted `.bib` given https://gist.github.com/jlelong/bdfe3cd93c7bca81cd7a8064251f5aa8#file-test-bib. 

Open it in vscode and run _Format Selection_ or _Format Document_, you obtain

Capture d’écran 2020-09-08 à 15 02 35

The warning message does not disappear even when you fix the duplicate entry. To make it disappear you have to run the format command again. I am not convinced that this is really what we want.

tamuratak:
Linting and formatting are different things. We should not do different things at the same time.

src/providers/bibtexformater.ts
tamuratak: What I am saying is that the following is easier to read and maintain:
```ts
    public async provideDocumentFormattingEdits(document: vscode.TextDocument, _options: vscode.FormattingOptions, _token: vscode.CancellationToken) {
        const edits = await this.formatter.formatDocument(document, true, true)
        this.formatter.duplicatesDiagnostics.set(document.uri, this.formatter.diags)
        this.extension.logger.addLogMessage('BibTeX action successful.')
        return edits
    }
```

jlelong: I feel a bit lost here. According to  https://code.visualstudio.com/api/references/vscode-api#DocumentFormattingEditProvider, `provideDocumentFormattingEdits` must have return type `vscode.ProviderResult<T>`, which is not possible for an `async` function.

jlelong: With the diagnostics removed (see 6d27c1b), I think it is fine now?

tamuratak:
I think sorting bib items always goes too far. It should be configurable.

src/providers/bibtexformater.ts
```diff
@@ -67,56 +96,67 @@ export class BibtexFormater {
...
+    private formatter: BibtexFormatter
+    extension: Extension
+
+    constructor(extension: Extension) {
+        this.extension = extension
+        this.formatter = new BibtexFormatter(extension)
+    }
+
+    public provideDocumentFormattingEdits(document: vscode.TextDocument, _options: vscode.FormattingOptions, _token: vscode.CancellationToken): vscode.ProviderResult<vscode.TextEdit[]> {
+        return this.formatter.formatDocument(document, true, true)
```
tamuratak, line 155: I think sorting bib items always goes too far. It should be configurable.

jlelong: I am not sure to understand exactly.
Do you suggest to let the user configure whether he wants to sort the content or just align it?

tamuratak: Yes. The current formatter, `latexindent.pl`, have not sorted items. Just aligning the content by default, not sorting, would be what users expect for a formatter by default.

tamuratak: We should add to log messages why formatting starts. VS Code executed it as the formatter, or the user executed the command.

src/providers/bibtexformater.ts
```diff
@@ -67,56 +96,71 @@ export class BibtexFormater {
...
+
+    constructor(extension: Extension) {
+        this.extension = extension
+        this.formatter = new BibtexFormatter(extension)
+    }
+
+    public provideDocumentFormattingEdits(document: vscode.TextDocument, _options: vscode.FormattingOptions, _token: vscode.CancellationToken): vscode.ProviderResult<vscode.TextEdit[]> {
+        const sort = vscode.workspace.getConfiguration('latex-workshop').get('bibtex-format.sort.enabled') as boolean
+        this.extension.logger.addLogMessage('Start bibtex formatting on behalf of VSCode\'s formatter.')
+        return this.formatter.formatDocument(document, sort, true)
```
tamuratak, line 157: We should catch errors of formatting and add them to log messages, especially for parsing errors.

tamuratak (changes requested):

tamuratak:
`Close #2261` is self-referencing. It should be #2250.

src/providers/bibtexformater.ts
jlelong: This is resolved by 27c50b1

https://github.com/James-Yu/LaTeX-Workshop/blob/27c50b140ab32b1c1d28dead0eb3311e869ed6a1/src/providers/bibtexformater.ts#L67-L77

src/providers/bibtexformater.ts
```diff
@@ -35,7 +61,20 @@ export class BibtexFormater {
...
+        const ast = await this.extension.pegParser.parseBibtex(document.getText(range)).catch((error) => {
+            if (error instanceof(Error)) {
+                this.extension.logger.addLogMessage('Bibtex parser failed.')
+                this.extension.logger.addLogMessage(error.message)
+                this.extension.logger.showErrorMessage('Bibtex parser failed with error: ' + error.message)
+            }
+            return undefined
+        })
+        if (! ast) {
+            return []
```
tamuratak, line 76: With the error handling, when calling `latex-workshop.bibalign`, the message `BibTeX action successful. ` will be logged even if it fails.
```
[21:47:22] Start bibtex formatting on user request.
[21:47:22] Bibtex parser failed.
[21:47:22] Expected "=", "\r\n", or [ \t\n] but "a" found.
[21:47:22] BibTeX action successful. Took 33.06112098693848 ms.
```

tamuratak (changes requested):

src/providers/bibtexformater.ts
jlelong: Thanks for all this. This is a kind of a headache to allow both a direct usage and the formatter approach.

tamuratak (approved):

PR #2264, opened 2020-09-04
Add support for relative file paths in SyncTeX.

markusbrandt:
Hi,

I added support for SyncTeX files with relative paths. This is useful when accessing documents from different machines.


tamuratak:
Not convinced.

Issue #2206, opened 2020-07-15
PDFviewer broken in VS Code 1.47.0 over remote ssh


thegregyang:
### Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: Yes

## Describe the bug [Required]

The PDFviewer does not display the PDF in VS Code 1.47.0 over remote ssh.

### To Reproduce

1. Open a latex project on a remote server via remote ssh plugin
2. Click on the "View Latex PDF file" button
3. A new tab opens but nothing is displayed

### Expected behavior

The pdf file should be displayed in the new tab.

## Logs [Required]

Please paste the whole log messages here, not parts of ones. The log should start with `Initializing LaTeX Workshop`. It is very important to identify problems.

### LaTeX Workshop Output [Required]

```
To access the log, click the 'TeX' icon on the Activity Bar on the left side, select 'View Log Messages', then select 'View LaTeX Workshop extension log'.

[01:53:41] Initializing LaTeX Workshop.
[01:53:41] Creating PDF file watcher.
[01:53:41] Creating LaTeX Workshop http and websocket server.
[01:53:41] LaTeX Workshop initialized.
[01:53:41] Found root file of this subfile from active editor: /home/gregyang/sambashare/tex-workspace/TensorProgramFullPower/main.tex
[01:53:41] Root file changed: from undefined to /home/gregyang/sambashare/tex-workspace/TensorProgramFullPower/main.tex
[01:53:41] Start to find all dependencies.
[01:53:41] Instantiating a new file watcher for /home/gregyang/sambashare/tex-workspace/TensorProgramFullPower/main.tex
[01:53:41] Creating Bib file watcher.
[01:53:41] Parsing a file and its subfiles: /home/gregyang/sambashare/tex-workspace/TensorProgramFullPower/main.tex
[01:53:41] Parsing a file and its subfiles: /home/gregyang/sambashare/tex-workspace/TensorProgramFullPower/NetsortTMain.tex
[01:53:41] Parse fls file.
[...]

PR #2270, opened 2020-09-11
Use the polyfilled version of pdfjs

James-Yu:
In `PDF.js v2.4.456`, the default `pdf.js` is changed to non-ployfilled version to cater with modern browsers. However, nodejs does not have several necessary polyfills implemented to support the pdf viewer used in the extension.

When using WSL/SSH remote, the extension essentially cannot work due to unknown causes which lead to the use of non-ployfilled version, even all references were pointing to `es5/build` instead of `build`. Therefore, this PR propose a somehow dirty but effective workaround by replacing the non-polyfilled pdf.js with the polyfilled one. I believe this won't cause any issues, and the performance impact shall be minuscule if not nothing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1936,7 +1936,7 @@
     "compile": "tsc -p tsconfig.json && tsc -p viewer/tsconfig.json",
     "lint": "eslint --cache --ext .ts,.js .",
     "lint:fix": "eslint --fix --cache --ext .ts,.js .",
-    "release": "npm run clean && npm run lint && npm run compile && vsce package",
+    "release": "cp -r node_modules/pdfjs-dist/es5/build/ node_modules/pdfjs-dist && npm run clean && npm run lint && npm run compile && vsce package",
     "test": "node ./out/test/runTest.js",
     "watch": "run-p watch-src watch-viewer",
     "watch-src": "tsc -watch -p tsconfig.json",


James-Yu:
While I'd merge this PR, the solution is only a workaround. I am still playing with all `pdf.js` references, but cannot find a more elegant solution.

tamuratak:
Please revert this PR. I have fixed the issue in a5ae5dfc36a8c9a64aff24816bb3e1801fd0a223 already. We don't have`require("pdfjs-dist")` in `src/components/builder.js` of the master.

https://github.com/James-Yu/LaTeX-Workshop/blob/8855301a04bd0f7cdc0266543024c64033ed67b5/src/components/builder.ts#L1-L9

In `james-yu.latex-workshop-8.12.0/out/src/components/builder.js` of the package `v8.12.0`, we still see `require("pdfjs-dist")`.

```
const tmp = __importStar(require("tmp"));
const pdfjsLib = __importStar(require("pdfjs-dist"));
const await_semaphore_1 = require("../lib/await-semaphore");
```

Something was wrong with building the release package.

tamuratak:
cc: @jlelong

James-Yu:
Your commit a5ae5df cannot resolve the issue when using with WSL. I tested it, though I don't know the reason. Even changing all occurrence of `build/pdf` to `es5/build/pdf` does not fix the issue.

Please also note that the issue only exists with WSL and SSH. In normal case it won't happen. I have totally no idea why it happens.

James-Yu:
Tested again and confirmed the issue. The test is quite simple:
1. remove the `build` folder from `node_modules/pdfjs-dist`.
2. Open vscode, try to open the viewer and build project.

The result is quite interesting. On Windows host, everything works alright as expected. In WSL, the viewer lead to an empty page. It seems that when deployed to WSL by vscode remote, something unexpected happened.

The test is conducted with `8.13.0`, in which the previous `8.12.0` orphaned `import` was removed.

tamuratak:
>  In WSL, the viewer lead to an empty page. 

Duplicate of #2206

tamuratak:
Please create a `vsix` file with the following steps,
```
git checkout v8.12.0
rm -rf node_modules/
npm i --no-optional 
npm run release
```
and then, test with the `vsix`file. You can install the file on the remote host too.

tamuratak:
Moved to #2272

PR #2278, opened 2020-09-12
CSP: Allow websocket connections from 'self'

sleiner:
For reasons (TM), I would like to view the browser preview on another device than the one that VS Code is running on.
As far as I see, this is not supported per se (or am I overlooking something?) - anyway, I got the basics working using a reverse proxy on the VS Code machine.

However, when using Safari for the preview page on the other device, synctex is not working and the following error is produced:
```
Refused to connect to ws://foobar.local:8010/ because it does not appear in the connect-src directive of the Content Security Policy.
```

I was able to remove the error and get syntax working by removing the restriction that websocket connections are only allowed from `127.0.0.1` and `localhost`. 
Note that this **does not** allow websocket connections to arbitrary host, but only to the ones specified in `connect-src`, so in this case only to `self` 👍

### DIFF
--- a/viewer/viewer.html
+++ b/viewer/viewer.html
@@ -26,7 +26,7 @@ See https://github.com/adobe-type-tools/cmap-resources
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
     <meta name="google" content="notranslate">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self' 'unsafe-eval'; connect-src 'self' ws://localhost:* ws://127.0.0.1:*; style-src 'self' 'unsafe-inline'; img-src 'self' data: blob:;">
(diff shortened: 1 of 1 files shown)


tamuratak:
> For reasons (TM), I would like to view the browser preview on another device than the one that VS Code is running on.

Please use VS Code Remote for the purpose. If you cannot use VS Code on your device, it is out of the scope of LaTeX Workshop.

sleiner:
Thanks for the quick reply, @tamuratak!

> If you cannot use VS Code on your device, it is out of the scope of LaTeX Workshop.

Indeed this is a device where this is not possible (an iPad). I understand very much that this use case it out of scope for LaTeX workshop which is why I'm perfectly happy with rolling my own custom reverse proxy setup for this (admittedly a bit obscure) use case.

---

But I think this patch should be considered for merging indecently of the use case being unsupported:
 - For the supported use case of localhost access only it changes nothing about the behaviour of the `content-src` tag - as far as I can see, **the effective policy stays the same**
 - It is highly unlikely to cause any maintenance effort because no new code is introduced 😃 
 - It makes my obscure use case possible by just not hardcoding `localhost` and `127.0.0.1

What Do you think?

PR #2298, opened 2020-09-23
Glossary completion

### LINKED ISSUE #516: IntelliSense for Glossaries
Having autocompletion for the defined glossaries entires would be great. 

There seems to be an atom package for this: https://atom.io/packages/autocomplete-glossaries

trevorgunn:
Closes #516 

Still a work-in-progress but the core functionality is ready for review I think.

**TODO**:

- [x] Write a fallback regex-based parser
- [x] Review the LaTeX glossaries package to see if there are any LaTeX commands the parser needs to be aware of.


src/providers/completion.ts
```diff
@@ -85,7 +88,7 @@ export class Completer implements vscode.CompletionItemProvider {
         const line = document.lineAt(position.line).text.substr(0, position.character)
         // Note that the order of the following array affects the result.
         // 'command' must be at the last because it matches any commands.
-        for (const type of ['citation', 'reference', 'environment', 'package', 'documentclass', 'input', 'subimport', 'import', 'includeonly', 'command']) {
+        for (const type of ['citation', 'reference', 'environment', 'package', 'documentclass', 'input', 'subimport', 'import', 'includeonly', 'command', 'glossary']) {
```
jlelong, line 91: `command` must be the last type of completion because its regex matches any command starting with `\`. Put `glossary` before it.

jlelong:
@trevorgunn I am glad you gave it a try as this was something I wanted to do but I never found the time.

jlelong:
I think that listing existing glossary entries and acronyms is fairly complete. We have still to figure out the citing commands

From the sections 6 and 13 of [glossaries package manual](https://mirrors.chevalier.io/CTAN/macros/latex/contrib/glossaries/glossaries-user.html), I can list the following commands

- `(gls|GLs|GLS)(pl|text|first|plural|firstplural|name|symbol|desc|user(i|ii|iii|iv|v|vi))?`: all (almost all?) these commands seem to work both acronyms and glossary entries.
- `Acr(long|full|short)?(pl)?` or `[Aa]c[slf]?p?`: only to be used for acronyms.

Are there any glossaries users to give feedback on this?

trevorgunn:
There are some additional commands for providing glossary items like `\provideglossaryentry` and `\longprovideglossaryentry`. I'll add those into the code soon.

trevorgunn:
I changed the completion regex to be case insensitive so we don't need to test for `(gls|Gls|GLS|acr|Acr|ACR)` all separately but that can probably still be improved to support the other commands you listed.

jlelong:
I think this PR is now ready for review. There remains to automatically trigger glossary completion when a glossary reference command is entered. For citations and references, this is done in 

https://github.com/James-Yu/LaTeX-Workshop/blob/a5c4604654bba82329c2b690d1d977a8db71e097/src/providers/completer/command.ts#L467-L469

jlelong:
I am now merging this PR. We can improve the feature when we have users' feedback once a new release is published.

Issue #2326, opened 2020-10-13
Using LaTeX Workshop in a webbrowser


kromsam:
I have a question to what would be the recommended way to have a remote installation of vscode to run LaTeX Workshop in. Dockerized preferably. I see that code-server is not supported, but does this plugin offer any alternatives? All the best.

tamuratak:
Out of the scope of the extension.

PR #2341, opened 2020-10-25
Allow configurable viewer URL

lukas-hetzenecker:
this supports e.g. LaTeX-Workshop running in code-server

I know similar things were previously discussed - e.g. in #2326, #1265, #1271
I am not sure what the original PR in #1271 was as it seems to have been force-pushed to remove the configurable URL, but the approach here should have minimal security implications.

---

LaTeX-Workshop is already working reasonably well with code-server, with most of the features working out-of-box.
Only the integrated PDF viewer does not - for the simple reason that its iframe is loaded from `http://localhost:random_port`.

This pull request changes nothing in terms of security (e.g. embedded http server is _still_ only bound to 127.0.0.1). 
The only change is that the viewer can be loaded from a different **URL**. 
Then the [reverse proxy feature of code-server](https://github.com/cdr/code-server/blob/v3.6.1/doc/FAQ.md#sub-domains) can be leveraged for accessing the viewer.

## Example

`latex-workshop.viewer.pdf.internal.url` is set to `https://%p.codeserver.com`
and code-server started with `code-server --proxy-domain codeserver.com`

This request gets now sent to code-server, which proxies it to localhost:%p. All security implications of that are outsourced to code-server (and its proxy-domain feature) and/or the loadbalancer before.

Please let me know if you are willing to accept such a feature request.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1254,6 +1254,11 @@
           "default": "0",
           "markdownDescription": "Define the port to listen on for communicating with the internal viewer. The default value \"0\" means the port is chosen randomly by the application."
         },
+        "latex-workshop.viewer.pdf.internal.url": {
+          "type": "string",
+          "default": "http://localhost:%p",
+          "markdownDescription": "URL used for connecting to the internal viewer. \"%p\" is replaced by the listening port of the viewer."
+        },
         "latex-workshop.view.pdf.internal.synctex.keybinding": {
           "type": "string",
           "default": "ctrl-click",
--- a/src/components/server.ts
+++ b/src/components/server.ts
@@ -14,15 +14,18 @@ export class Server {
     private wsServer: ws.Server
     address?: string
     port?: number
+    url?: string
 
     constructor(extension: Extension) {
         this.extension = extension
         this.httpServer = http.createServer((request, response) => this.handler(request, response))
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
         const viewerPort = configuration.get('viewer.pdf.internal.port') as number
+        const viewerURL = configuration.get('viewer.pdf.internal.url') as string
         this.httpServer.listen(viewerPort, '127.0.0.1', undefined, () => {
             const {address, port} = this.httpServer.address() as AddressInfo
             this.port = port
+            this.url = viewerURL.replace('%p', port.toString())
             if (address.includes(':')) {
                 // the colon is reserved in URL to separate IPv4 address from port number. IPv6 address needs to be enclosed in square brackets when used in URL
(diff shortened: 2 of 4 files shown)


tamuratak:
Out of the scope of the extension.

James-Yu:
@modaresimr May I invite you to introduce your steps of usage of this patch on code-server, so that the extension will work? Is the proposed port placeholder `%p` a must?

modaresimr:
%p placeholder is a need since we need to be able to distinguish between different ports in vscode server. For example, when we open two repositories we will have two ports.

based on the documentation there is two ways to access internal ports in vscode server: (subdomain) and (path) https://coder.com/docs/code-server/latest/guide#accessing-web-services

However,  the problem is sth else:

Latex-Workshop viewer uses an absolute path (starts with /). 
However, VSCode server makes a proxy for an internal port to a path like /proxy/port/
It means that if my vscode server URL is vscode.mydomain.com, and the port for the pdf viewer is 1000, the viewer URL can be vscode.mydomain.com/proxy/1000/ 
Therefore, vscode.mydomain.com/proxy/1000/viewer.html can be opened successfully. However, since its resources like js files have absolute paths, the browser tries to open vscode.mydomain.com/*. for example, for /viewer.js it will open vscode.mydomain.com/viewer.js instead of vscode.mydomain.com/proxy/1000/viewer.js

This is why I use https://%p.vscode.mydomain.com in my [fork](https://github.com/modaresimr/LaTeX-Workshop) 

However, I think https://vscode.mydomain.com/proxy/%p/ is a better option

James-Yu:
Thanks @modaresimr for your very detailed explanation? May I interpret the message as that, as long as absolute paths are changed to relatives, the viewer will work for code-server without the need of customizing `localhost:port`? Or the latter is a must and prerequisite?

modaresimr:
From my point of view, there is no need.
Simply, if you return this URL: `window.location.protocol+"//"+window.location.host+"/proxy/{{port}}/viewer.html?file=....`, the pdf viewer should work (after removing the / at the beginning of resources).

Normally, [asExternalUri](https://github.com/coder/code-server/pull/5624) should do the same thing but in reality, it will not add the path to it (it returns only myserver.com/proxy/port/). Therefore, we need to handle it by ourselves.

James-Yu:
I'm inclined on not touching the server url part as it is highly error-prone. As I do not use code-server, a PR on this config item change is welcomed and I will review it.

modaresimr:
I have fixed the problem  in https://github.com/James-Yu/LaTeX-Workshop/pull/3639

PR #2380, opened 2020-11-19
Add toggle math preview

### LINKED ISSUE #2379: Ctrl+Alt+M to close the math preview panel
As far as I'm concerned, since ctrl+alt+M is opening the math preview panel, then intuitively ctrl+alt+M when the panel has been opened should be closing the panel. Just like switch on and off a light.

Actually, I think that an average latex user is not likely to use the math preview panel all the time, because one usually doesn't need the preview while typing simple equations, and long and tough equations usually show up less than text. But the preview panel take up lots of space so it's not so comfortable while typing text. Therefore, I think the best way to use the preview panel is to call it out while I need it, and close it after use in order to save the screen space. 

Imagine that you suddenly need to type a terribly long equation, you could just press ctrl+alt+M to call out the panel, and then type the equation carefully, and after that you could just press ctrl+alt+M to quickly close the panel. Then you can go on your writing, and the math preview panel won't take up the space of the screen. That's really convenient and easy.

Thus a quick way to close the preview panel is quite significant.  However, in the latest version, in order to close the preview panel one needs to ctrl+shift+p and type a command. This costs too much attention and too many steps. I'd appreciate it a log if the developers could consider my suggestions.

jlelong:
Fix #2379.
I would even be tempted to remove close and open math preview panel commands as they become useless once we have a toggle action.

### DIFF
--- a/package.json
+++ b/package.json
@@ -467,13 +467,18 @@
         "command": "latex-workshop.closeMathPreviewPanel",
         "title": "Close Math Preview Panel",
         "category": "LaTeX Workshop"
+      },
+      {
+        "command": "latex-workshop.toggleMathPreviewPanel",
+        "title": "Toggle Math Preview Panel",
+        "category": "LaTeX Workshop"
       }
     ],
     "keybindings": [
       {
         "key": "ctrl+alt+m",
         "mac": "cmd+alt+m",
-        "command": "latex-workshop.openMathPreviewPanel"
+        "command": "latex-workshop.toggleMathPreviewPanel"
       },
       {
         "key": "ctrl+l alt+b",
--- a/package.json
+++ b/package.json
@@ -476,9 +476,10 @@
     ],
     "keybindings": [
       {
-        "key": "ctrl+alt+m",
-        "mac": "cmd+alt+m",
-        "command": "latex-workshop.toggleMathPreviewPanel"
+        "key": "ctrl+l alt+m",
+        "mac": "cmd+l alt+m",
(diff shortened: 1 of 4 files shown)


tamuratak:
> I would even be tempted to remove close and open math preview panel commands as they become useless once we have a toggle action.

I am against removing the commands.

package.json
```diff
@@ -467,13 +467,18 @@
...
+        "category": "LaTeX Workshop"
       }
     ],
     "keybindings": [
       {
         "key": "ctrl+alt+m",
         "mac": "cmd+alt+m",
-        "command": "latex-workshop.openMathPreviewPanel"
+        "command": "latex-workshop.toggleMathPreviewPanel"
       },
```
tamuratak, line 482: We should add the `when` clause for cases `altKeymap` are enabled and disabled.  I have missed it.

tamuratak (changes requested):

package.json
jlelong: > I am not sure whether the clause should include the condition of `editorLangId`.

We need it for the `altKeymap` shortcut because it uses `ctrl-l` as prefix, which is also the standard vscode shortcut for expanding line selection. Also, `ctrl-alt-m` is used by the [JSON tools extension](https://marketplace.visualstudio.com/items?itemName=eriklynd.json-tools). So I added the `editorLangId` condition in both shortcuts.

tamuratak (approved):

PR #2392, opened 2020-11-24
Keep selected text in autocompletions/snippets

### LINKED ISSUE #2386: Autocomplete overwrites selected text while snippets do not
Selecting text and typing a snippet like `FTT` surrounds the selected text with the snippet content `\texttt{selected text}`.
The same with an autocompletion like `\section*{text}` which shows up when typing `\sec...` overwrites the selected text and just shows up as `\section*{}`.

It would be great if normal autocompletions would also preserve the selected text.

Provide every current autocompletion as a snippet but that may be a lot.
Alternatively provide more generic snippets like an environment snippet which puts the selected text inside the environment body and moves the cursor to the environment type (the brackets after the begin/end statement).

jlelong:
Close #2386

- [x] **environments** This works when inserting environments with `\envname` or the [`B` snippets](https://github.com/James-Yu/LaTeX-Workshop/wiki/Snippets#Environments) and when using the `\begin/\end` snippet. 
  https://github.com/James-Yu/LaTeX-Workshop/blob/7eb309ec1ae6521f1f8887bccae3f1f02c557cc5/data%2Fcommands.json#L8-L13

  Note that the autocompletion on `\begin{`, which inserts `envname` and `\end{envname}` does not keep the selected text because the environment is actually inserted by a second call to autocompletion. See https://github.com/James-Yu/LaTeX-Workshop/blob/7eb309ec1ae6521f1f8887bccae3f1f02c557cc5/data%2Fcommands.json#L2-L7

- [x] **commands**

With this PR, the `wrap-env` function becomes useless, we should mark it as deprecated.

### DIFF
--- a/data/commands.json
+++ b/data/commands.json
@@ -5,6 +5,12 @@
     "detail": "Begin a new environment",
     "postAction": "editor.action.triggerSuggest"
   },
+  "beginend": {
+    "command": "beginend",
+    "snippet": "begin{$1}\n\t${0:${TM_SELECTED_TEXT}}\n\\end{$1}",
+    "detail": "Insert \\begin/\\end",
+    "postAction": "editor.action.triggerSuggest"
+  },
   "left(": {
     "command": "left(",
     "snippet": "left(${1}\\right)",
--- a/data/commands.json
+++ b/data/commands.json
@@ -9,6 +9,7 @@
     "command": "beginend",
     "snippet": "begin{$1}\n\t${0:${TM_SELECTED_TEXT}}\n\\end{$1}",
     "detail": "Insert \\begin/\\end",
+    "documentation": "Insert a `\\begin{} / \\end{}` pair and wrap the selected text if any. Autocompletion is automatically triggered to complete the environment name",
     "postAction": "editor.action.triggerSuggest"
   },
   "left(": {
--- a/snippets/latex.json
+++ b/snippets/latex.json
@@ -6,12 +6,12 @@
 	},
 	"subscript": {
 		"prefix": "__",
-		"body": "_{$1}",
+		"body": "_{${1:${TM_SELECTED_TEXT}}}",
 		"description": "subscript"
 	},
 	"superscript": {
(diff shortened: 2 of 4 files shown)


tamuratak:
This PR does not work well with the command `latex-workshop.surround`.

jlelong:
The command `latex-workshop.surround` is fixed. However, it becomes useless with this PR and should be marked as deprecated.

jlelong:
I think we should remove https://github.com/James-Yu/LaTeX-Workshop/blob/57f01831d4d73913390dd37ddac4dc255a6c324c/src%2Fconfig.ts#L110-L116

tamuratak:
I prefer `latex-workshop.surround`.

The behavior of surrounding a text with TM_SELECTED_TEXT is not intuitive to me since the selected text is removed for a while during the operation.

Nov-25-2020 23-42-40

jlelong:
Do you mean that you prefer not to mark `latex-workshop.surround` as deprecated? Personally, I do not care.

tamuratak:
Yes. I am against marking `latex-workshop.surround` as deprecated.

jlelong:
This should be fine now.

tamuratak:
Because the `beginend` command is LaTeX-Workshop specific, we should explain it a bit more in the `documentation` field. 

スクリーンショット 2020-11-26 16 08 14

For this purpose, we can define the `beginend` command in `src/providers/completer/command.ts`, not in `data/commands.json`.

jlelong:
No need to resort to `src/providers/completer/command.ts`, the `documentation` field can be directly defined in `commands.json`.

I think the following has become useless https://github.com/James-Yu/LaTeX-Workshop/blob/7eb309ec1ae6521f1f8887bccae3f1f02c557cc5/src%2Fproviders%2Fcompleter%2Fenvironment.ts#L38-L45

To insert a `begin/end` pair, we should simply use `\beginend`.

tamuratak:
Looks good to me.

PR #2370, opened 2020-11-10
Deactivate autoclosing for snippet prefixes

jlelong:
Close #2368.
This should also fix #1423 and #1820

@tamuratak Can you test this ?

PR #2293, opened 2020-09-19
Fix double }} in env completion

jlelong:
Improve #2080
Close #2290

PR #2394, opened 2020-11-25
Refactor fix for double }} in completion

jlelong:
This commit basically reverts #2293 and #2080. Instead it uses the same approach as #2370.
The new approach is less intrusive and much cleaner.
@tamuratak can you test it?


jlelong:
The `}}` issue also appears when inserting snippets ending with `(`, `[`, `{`.

Nov-25-2020 18-51-33

The same occurs with all the `\big`'s commands. It can be fixed by adding all these commands to 

https://github.com/James-Yu/LaTeX-Workshop/blob/142ff2dfd1e0e044faf214cfb48fe7d381abbcff/syntax%2Fsyntax.json#L14-L15

tamuratak:
The issue #2075 occurs again.

Nov-26-2020 08-45-53

If you are saying that stopping auto-closing with `}` for `\begin{` is enough, I agree it is enough.

jlelong:
#2075 was related to the auto-closing of `{`. With this PR, typing `\begin{item` and completing works fine

Nov-26-2020 09-32-03


However, it is true that completing `\begin{item}` fails

Nov-26-2020 09-33-05

But I cannot really see the purpose of such as completion. Instead, 

Nov-26-2020 09-36-24

What could when completing `\begin{item}` with the cursor inside `{}` is to only complete the environment name and not to add the `\end{}` part.

Are you fine with this approach

tamuratak:
Yes, I am fine with it.

PR #2397, opened 2020-11-26
Autocomplete big paren commands: 'left(' -> '\left( ... \right)'

### LINKED ISSUE #2112: \left( does not match \left( ... \right) in suggestions.

jlelong:
This is a follow up on #2116 and #2394

The following commands get a unique autocompletion : `(\\left|\\[Bb]ig{1,2}l|\\)[({[]`

Nov-26-2020 15-52-34

Note that to insert `\left\{ ... \right\}`, you have to type `\left{` without the `\` before the `{`. This is because autocompletion considers that the command to complete starts at the last `\` and we cannot change that.

Close #2112. I think this PR finally (and hopefully) solves all the auto-closing issue in autocompletion.


tamuratak:
1. Type `\le`
2. Enter `Ctrl+Space`

Suggests include only one entry:

Nov-27-2020 14-17-29

jlelong:
The filtering was a little too aggressive. This is fixed.

tamuratak (approved):

Issue #2395, opened 2020-11-26
The extension is interefering with c++ doxygen syntax highlighting


mehdi-benallegue:
### Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]
Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

Yes

## Describe the bug 
Activating this extension removes all syntax highlighting of doxygen comments in c++ files. 

## Screenshots

Without the extension
![Screenshot from 2020-11-26 17-57-23](https://user-images.githubusercontent.com/435367/100329612-3e232500-3011-11eb-84b6-e929eb1bf3d6.png)

With the extension
![Screenshot from 2020-11-26 17-57-55](https://user-images.githubusercontent.com/435367/100329621-44190600-3011-11eb-9733-38281282bfdc.png)

## Desktop
 - OS: Ubuntu 18.04
 - VS Code version: [. 1.51.1]
 - Extension version: [ v8.14.0]

### Are you using VSCodium?

No

### Are you using LaTeX Workshop with VS Code Remote?

No

### Additional context
I haven't even started to compile latex documents...

tamuratak:
I can reproduce. The following is the cause:

https://github.com/James-Yu/LaTeX-Workshop/blob/a5621669636e47912557a6281e8db39288b0b469/package.json#L216-L223

PR #2398, opened 2020-11-27
Stop overriding the cpp grammar

tamuratak:
We should stop overriding the `cpp` grammar. We can avoid overriding by defining a new language id. Related to #2395.

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,6 +152,9 @@
           ".Snw"
         ],
         "configuration": "./syntax/syntax-weave.json"
+      },
+      {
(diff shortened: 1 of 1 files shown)


jlelong:
Overriding the `cpp` grammar was suggested by `cpp` extension maintainers jeff-hykin/cpp-textmate-grammar/pull/351. This was introduced to fix #1476 and microsoft/vscode/issues/76603

I have to look at all this closely.

jlelong (approved):

PR #2420, opened 2020-12-05
Fix a focus issue when opening PDF files through the explorer.

tamuratak:
Fix a focus issue when opening PDF files through the explorer.


tamuratak:
I will rewrite it.

PR #2465, opened 2021-01-07
Check if git is available on OSX to prevent dev tools pop up

### LINKED ISSUE #2457: A dialog pops up on macOS if git is not installed and search.useIgnoreFiles is true
A dialog pops up on macOS if `git` is not installed and `search.useIgnoreFiles` is true.

The dialog says `the "git" command requires the command line developer tools. Would you like to install the tools now?`

We should run `which git` on macOS before executing `git`.

https://github.com/James-Yu/LaTeX-Workshop/blob/d375dfc7e9e75b911ac96b350520dd5234678894/src/providers/completer/input.ts#L25-L29

jlelong:
Close #2457

@tamuratk can you test the PR on your version of OSX? I want to make the `which` command has the same behaviour as on mine.

### DIFF
--- a/src/providers/completer/input.ts
+++ b/src/providers/completer/input.ts
@@ -4,6 +4,7 @@ import * as path from 'path'
 import * as micromatch from 'micromatch'
 import * as cp from 'child_process'
 import * as utils from '../../utils/utils'
+import * as os from 'os'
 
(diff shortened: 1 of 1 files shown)


tamuratak (approved):
It works well on may mac.

macOS Mojave 10.14.6

PR #2476, opened 2021-01-14
Enable converting ChkTeX outputs' column numbers.

### LINKED ISSUE #829: Wrong column number from chktex (Tab) + character/byte skipping in actions
Hi,

Here is the interesting screenshot showing two issues:
image

If I got the things correctly, chktex detected a common `$...$` -> `\(...\)` TeX vs LaTeX thing starting at 13th character but it reported column number is 21 (because of its own setting `TabSize = 8`). When I remove the leading tab character, this is what happens:
image

As you can see, codeaction suggestion is shifted by one character to the right-hand side, and clicking it does not make any change. The root cause for the misplacement is the character `č` in UTF8-encoded input file. Once I remove it, codeaction suggestion is placed correctly and the action successfully replaces `$P(x,y)$` by `\(P(x,y)\)`.

So, to conclude:
1) [UX suggestion] Could LaTeX-Workshop consider `TabSize = 8` setting from chktex?
2) [Bug?] Is it really true that non-one-byte-characters are the root cause of the presented issue and can that be fixed?

Kind regards, Ivan

tamuratak:
Enable converting ChkTeX outputs' column numbers. Close #829.

スクリーンショット 2021-01-14 20 18 53

```latex
\documentclass{article}
\usepackage{amsthm,amsmath,amssymb}
\begin{document}

	漢字 $(a$ $(a$ $(a$

	 Ozna─Źimo s $P(x,y$ uvr┼ítavanje 	 字 vrijednosti $(x$ i $(y$ u po─Źetnu jednad┼żbu.

\end{document}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1512,9 +1512,9 @@
           "markdownDescription": "Enable converting ChkTeX outputs' column numbers for non-ASCII characters."
         },
         "latex-workshop.chktex.convertOutput.column.chktexrcTabSize": {
-          "type": "string",
-          "default": "auto",
-          "markdownDescription": "Write the `TabSize` number in `.chktexrc`. If the value is `auto`, LaTeX Workshop will try to read the value from `.chktexrc`. The file in the home directory will not be read."
+          "type": "number",
+          "default": -1,
+          "markdownDescription": "Write the `TabSize` number from `.chktexrc`. The default value \"-1\" means that LaTeX Workshop will try to find `.chktexrc` and to read the value from it."
         },
         "latex-workshop.texcount.path": {
           "type": "string",
--- a/package.json
+++ b/package.json
@@ -1506,6 +1506,16 @@
           "default": 500,
           "markdownDescription": "Defines the delay in milliseconds for chktex to wait after stopped typing.\nThis config only matters when `latex-workshop.chktex.run` is set to `onType`."
         },
+        "latex-workshop.chktex.convertOutput.column.enabled": {
+          "type": "boolean",
+          "default": true,
+          "markdownDescription": "Enable converting ChkTeX outputs' column numbers for non-ASCII characters."
+        },
+        "latex-workshop.chktex.convertOutput.column.chktexrcTabSize": {
+          "type": "string",
+          "default": "auto",
+          "markdownDescription": "Write the `TabSize` number in `.chktexrc`. If the value is `auto`, LaTeX Workshop will try to read the value from `.chktexrc`. The file in the home directory will not be read."
+        },
         "latex-workshop.texcount.path": {
           "type": "string",
           "default": "texcount",
--- a/src/components/linter.ts
+++ b/src/components/linter.ts
@@ -193,7 +193,9 @@ export class Linter {
         const reg = /^\s*TabSize\s*=\s*(\d+)\s*$/m
         const match = reg.exec(rcFile)
         if (match) {
-            return Number(match[1])
+            const ret = Number(match[1])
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1506,6 +1506,16 @@
...
         },
+        "latex-workshop.chktex.convertOutput.column.enabled": {
+          "type": "boolean",
+          "default": true,
+          "markdownDescription": "Enable converting ChkTeX outputs' column numbers for non-ASCII characters."
+        },
+        "latex-workshop.chktex.convertOutput.column.chktexrcTabSize": {
+          "type": "string",
+          "default": "auto",
+          "markdownDescription": "Write the `TabSize` number in `.chktexrc`. If the value is `auto`, LaTeX Workshop will try to read the value from `.chktexrc`. The file in the home directory will not be read."
```
jlelong, line 1517: It took me a while to understand the description. Maybe something like "Write the value of `TabSize` from `.chktexrc`" would be clearer. 

Why not reading the gobal `$HOME/.chktexrc` at least on Unix systems? Of course, https://github.com/James-Yu/LaTeX-Workshop/blob/61b3bbe1522428638baba7a81dfaf403f31c53b1/src/components/linter.ts#L18 would need to be updated.

src/components/parser/linterlog.ts
```diff
@@ -61,11 +65,66 @@ export class LinterLogParser {
...
+        const filePath = convertFilenameEncoding(filePathArg)
+        if (!filePath){
+            this.extension.logger.addLogMessage(`Stop converting chktex's column numbers. File not found: ${filePathArg}`)
+            return column
+        }
+        const lineString = fs.readFileSync(filePath).toString().split('\n')[line-1]
+        let tabSize: number | undefined
+        const tabSizeConfig = configuration.get('chktex.convertOutput.column.chktexrcTabSize', 'auto')
+        if (tabSizeConfig !== 'auto') {
+            tabSize = parseInt(tabSizeConfig)
```
jlelong, line 82: Instead of parsing strings here, why not typing `chktex.convertOutput.column.chktexrcTabSize` as a `number` and use `-1` for the auto value for instance? This would avoid the issue with strings not representing integers.

src/components/linter.ts
```diff
@@ -129,7 +130,24 @@ export class Linter {
...
+        const filePath = this.rcPath
+        if (!filePath) {
+            this.extension.logger.addLogMessage('The .chktexrc file not found.')
+            return
+        }
+        const rcFile = fs.readFileSync(filePath).toString()
+        const reg = /^\s*TabSize\s*=\s*(\d+)\s*$/m
+        const match = reg.exec(rcFile)
+        if (match) {
+            return parseInt(match[1])
```
jlelong, line 147: Maybe, we should enforce base `10` here. The default value for the `radix` argument of `parseInt` seems to depend on the platform.

tamuratak: I have chosen to use `Number`.

See https://stackoverflow.com/questions/4090518/what-is-the-difference-between-parseint-and-number

src/components/parser/linterlog.ts
tamuratak: Edited as suggested.

package.json
tamuratak: I have edited as suggested. For paths to search, I referred to the subsection "6.1.3 The chktexrc file" in `texdoc chktex`.

jlelong (approved):

Issue #2484, opened 2021-01-18
Killing the build process makes vscode crash when pkill is provided by proctools


alexreg:
### Disable all the other extensions except for LaTeX Workshop, restart VS Code, and check that you still see this issue.

You still see this issue?: Yes (sporadically)

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: No, but this is precisely the point

## Describe the bug [Required]

Sometimes a build hangs, but when I do "Kill LaTeX Compiler Process", it crashes VS Code.

### To Reproduce

1. Create a TeX document that hangs.
2. Observe never-ending build-in-progress status in bottom-right of window: <img width="130" alt="Screenshot 2021-01-18 at 03 35 32" src="https://user-images.githubusercontent.com/67036/104869548-395a5d80-593e-11eb-9d9c-0889ed0ba925.png">.
3. Perform command "Kill LaTeX Compiler Process".
4. Observe VS Code crash (immediate, no dialog or information screen).

### Expected behavior

Simply a peaceful termination of the LaTeX compiler process -- not a crash of the entire VS Code app, naturally.

### LaTeX Workshop Output [Required]

https://gist.github.com/0052a9d49640b2082303d4402df4640f

### LaTeX Compiler Output

https://gist.github.com/53d6e54ebc347f2bdcb82297bd547299

### Output from manually running `latexmk` command in terminal

https://gist.github.com/285257e52d062373d4f46327744041d1

### Developer Tools Console [Required]

https://gist.github.com/abf948c0194689c6e8426a916746170f

## Screenshots

See above.

## Desktop [Required]
 - OS: macOS 11.1
 - VS Code version: 1.52.1
 - Extension version: 8.15.0
 - TeX distribution version: TeX Live 2020

### Are you using VSCodium?

No

### Are you using LaTeX Workshop with VS Code Remote?

No

### Additional context

[Sample TeX document that causes hanging](https://gist.github.com/477f11b7e0896dc866ea2ae5640a866d)

alexreg:
Last time there was no crash screen, but this time I got this one:
[...]

PR #2493, opened 2021-01-24
Handle killing of build process properly

alexreg:
Fixes issue #2484. I've tested on macOS Big Sur, of course, and it's working nicely. I've also slightly modified the Windows method. On all platforms now, it should try a graceful termination before a hard kill (after a one second timeout).

I also took the liberty to rename the "Kill" commmand, but obviously you can choose to not use that commit. See the commit log for the reason, however.

CC @jlelong

### DIFF
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -48,9 +48,9 @@ export class Builder {
     }
 
     /**
-     * Kill the current build process.
+     * Stop the current build process.
      */
-    async kill() {
+    async stop() {
         if (this.currentProcess) {
             const envVarsPosix: ProcessEnv = {PATH: '/bin:/usr/bin'}
             const envVarsWindows: ProcessEnv = {PATH: '%SystemRoot%\\system32;%SystemRoot%;%SystemRoot%\\System32\\Wbem'}
@@ -253,7 +253,7 @@ export class Builder {
                     title: 'Running build process',
                     cancellable: true
                 }, (progress, token) => {
-                    token.onCancellationRequested(this.kill.bind(this))
+                    token.onCancellationRequested(this.stop.bind(this))
                     this.extension.buildInfo.buildStarted(progress)
                     const p = new Promise<void>(resolve => {
                         this.extension.buildInfo.setResolveToken(resolve)
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -5,7 +5,7 @@ import * as cp from 'child_process'
 import * as cs from 'cross-spawn'
 import * as tmp from 'tmp'
 import {Mutex} from '../lib/await-semaphore'
-import {replaceArgumentPlaceholders} from '../utils/utils'
+import {replaceArgumentPlaceholders, sleep} from '../utils/utils'
 
 import type {Extension} from '../main'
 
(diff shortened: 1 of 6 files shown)


jlelong:
Unfortunately, the PR does not work on Catalina because it only kills `latexmk` but not the `pdflatex` processes. The [documentation of the kill method](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal) says

> On Linux, child processes of child processes will not be terminated when attempting to kill their parent. This is likely to happen when running a new process in a `shell` or with the use of the shell option of `ChildProcess`

This was already discussed in #1680. Hence, the use of `pkill -P`.

jlelong:
Btw, I am not willing to rename `kill` into `stop` for several reasons

1. This method indeed calls the `kill` method from `child_process`
2. Users will keep looking for a `kill` method, which they will not find

alexreg:
Alright, I will look into that tomorrow, but it should be simple enough to resolve.

Regarding the renaming:
1. An implementation detail should not matter (affect naming). The standard terminology in build processes is "Stop" or occasionally "Cancel".
2. You could always keep that command around for a short transitionary period and have it either alias the "Stop" command or simply display an informational message about the renaming.
Indeed, I spent a while looking for a command to stop the build before realising it was weirdly called "Kill". It would seem sensible to adopt the terminology most consistent across build processes, but that's all I'll say on the matter.

alexreg:
@jlelong See my new push. This solves the issue, I believe. The real issue at the end of the day was that one can have more than one `pkill` installed (I had one in `/usr/local/bin` that was getting used). This is a different version of pkill, that seems to be problematic in that it doesn't like being run from within VS Code (maybe within the same process group?). So this explicitly specifies the `PATH` env var to get around that. I also do `spawn` instead of `exec` to a) prevent a shell possibly interfering with environment variables, b) saving time forking a shell process, even though it's not too much. Apart from that, the changes generally make the process of stopping/killing the build more robust, and like I said before, attempts a graceful termination before a hard kill.

You may want to double check that everything still works on Windows with regards to specifying the PATH variable correctly.

src/components/builder.ts
```diff
@@ -48,24 +48,68 @@ export class Builder {
...
+     * Stop the current build process.
      */
-    kill() {
-        const proc = this.currentProcess
-        if (proc) {
-            const pid = proc.pid
+    async stop() {
+        if (this.currentProcess) {
+            const envVarsPosix: ProcessEnv = {PATH: '/bin:/usr/bin'}
+            const envVarsWindows: ProcessEnv = {PATH: '%SystemRoot%\\system32;%SystemRoot%;%SystemRoot%\\System32\\Wbem'}
```
jlelong, line 56: I am against changing the `PATH` on Windows. I do not know how canonical they are on Windows and there is hardly no chance that there is an other `pkill` executable available in the `PATH`. Please, revert these changes.

src/components/builder.ts
```diff
@@ -48,24 +48,68 @@ export class Builder {
...
+
+            const currentProcessExit = new Promise(listener => this.currentProcess?.on('exit', listener))
+            const pid = this.currentProcess.pid
+
+            // Try to terminate the build process gracefully first.
             if (process.platform === 'linux' || process.platform === 'darwin') {
-                cp.execSync(`pkill -P ${pid}`)
+                try {
+                    // Terminate build process and all its child processes.
+                    await this.spawnAndWait('pkill', ['-TERM', '-P', pid.toString()], {env: envVarsPosix})
```
jlelong, line 65: Use [`spawnSync`](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options) instead of defining your own `spawnAndWait`.

src/components/builder.ts
```diff
@@ -48,24 +48,68 @@ export class Builder {
...
+                    await this.spawnAndWait('kill', ['-TERM', pid.toString()], {env: envVarsPosix})
```
jlelong, line 66: This is exactly what `this.currentProcess.kill()` does. Why not use it?

jlelong (changes requested):

jlelong:
As the PR has become quite complicated, I suggest to do one thing at a time and postpone the renaming to an other PR once  #2484 is fixed. 

Concerning the core of this PR,
- I am against changing the `PATH` on Windows. I do not know how canonical they are on Windows and there is hardly no chance that there is an other `pkill` executable available in the `PATH`. Please, revert these changes.
- The default signal used by the `kill` & co functions is `SIGTERM` so I do not think your approach is more graceful.

My general feeling is that the real issue is more on the [proctools](http://proctools.sourceforge.net/)  you have in `/usr/local/bin`. I am not sure how far the extension should go to support that.

src/components/builder.ts
alexreg: The second line, you mean? It is more consistent with the other `pkill` and `taskkill` commands (in that it runs a program), but I do see your point. I can switch to using `kill` if you stll prefer.

alexreg:
> * I am against changing the PATH on Windows. I do not know how canonical they are on Windows and there is hardly no chance that there is an other pkill executable available in the PATH. Please, revert these changes.

100% canonical, since they rely on %systemroot%. You can check this yourself in the Windows documentation.

> * The default signal used by the kill & co functions is SIGTERM so I do not think your approach is more graceful.

I should be more specific: it is more *robust* in that it tries SIGTERM and then SIGKILL if the former fails. It is more *graceful* in that you were previously automatically doing a hard kill (`/f`) on Windows, which is obviously a bad idea. Furthermore, just using synchronous functions/methods like `execSync` and `spawnSync` is not great, as it can cause VS Code to operate less smoothly (I experienced this during testing).

* I am not 100% sure that pkill is always in /bin or /usr/bin on Unix. Could not it be in /sbin or /usr/sbin?

I am almost positive they are always there from experience, but it's no big deal to add those two to the PATH too. What you don't want in the path is `/usr/local/bin` and `/opt/local/bin` and other stuff.

* My general feeling is that the real issue is more on the proctools you have in /usr/local/bin. Note that proctools has not been updated since 2013. Why not just remove that package from your homebrew installation? I am not sure how far the extension should go to support that pkill is not the standard one.

Telling people what to do with the other software they have installed (and what they put in their PATH) is a very bad idea. Not only will people resent being told, but moreover it is a well-known sound principle of software engineering to avoid conflicts with other installed software and libraries wherever possible. There could be many other scenarios of a `kill` or `pkill` program in the user's PATH that operates differently to the standard Posix one.

tamuratak:
I am also against merging this PR. Please uninstall `proctools`.

alexreg:
@tamuratak You have a real habit of insolence, you know. Such ingratitude. Nor do you understand basic software engineering principles, it seems. Who is in charge of this project? I would like to know that.

PR #2497, opened 2021-01-26
Should catch errors of kill commands.

tamuratak:
We should catch errors of kill commands. Otherwise, `proc.kill()` might not be executed.

`execSync` throws an error when the build process does not have a child process.

- https://nodejs.org/docs/latest-v12.x/api/child_process.html#child_process_child_process_execsync_command_options

Version: 1.52.1
Commit: ea3859d4ba2f3e577a159bc91e3074c5d85c0523
Date: 2020-12-16T16:30:02.420Z
Electron: 9.3.5
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Darwin x64 18.7.0

macOS Mojave 10.14.6

### DIFF
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -55,18 +55,19 @@ export class Builder {
         if (proc) {
             const pid = proc.pid
             try {
+                this.extension.logger.addLogMessage(`Kill child processes of the current process. PPID: ${pid}`)
                 if (process.platform === 'linux' || process.platform === 'darwin') {
-                    cp.execSync(`pkill -P ${pid}`)
+                    cp.execSync(`pkill -P ${pid}`, { timeout: 100 })
                 } else if (process.platform === 'win32') {
-                    cp.execSync(`taskkill /F /T /PID ${pid}`)
+                    cp.execSync(`taskkill /F /T /PID ${pid}`, { timeout: 100 })
                 }
             } catch (e) {
                 if (e instanceof Error) {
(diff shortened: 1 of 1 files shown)


jlelong (commented):
I am fine with this but I have two questions

- Do you we need a timeout in `execSync` in case the `pkill` commands hangs?
- What is the best way to make our `kill()` function async?

tamuratak:
I have set `timeout`. 

> What is the best way to make our kill() function async?

What do you want by making `kill` async? 

I think using `child_process.exec` is an appropriate way. However, it makes things a little complicated.

jlelong:
> What do you want by making `kill` async?
> I think using `child_process.exec` is an appropriate way. However, it makes things a little complicated.

I mean declaring 

https://github.com/James-Yu/LaTeX-Workshop/blob/6b594d54c5d84d4eb4f9935c8518c430601cfee0/src/components/builder.ts#L52-L53

async, but yes I agree that if it boilds down to using `child_process.exec` it makes handling the response a little (too?) complicated.

tamuratak:
Using `util.promisify` is one of solutions. See

https://nodejs.org/docs/latest-v12.x/api/child_process.html#child_process_child_process_exec_command_options_callback

jlelong:
That sounds interesting and may be used in 

https://github.com/James-Yu/LaTeX-Workshop/blob/4e9d49a9e22f7c58da49ea973b6c1e56954a67d2/src/components/cleaner.ts#L98-L117

PR #2499, opened 2021-01-28
CompilerLogParser.parse should not throw errors.

### LINKED ISSUE #2498: Compilation hangs if Latexlog exclude regex is illegal
You can tweak this setting by adding some wrong regex expressions, such as a single `*`, and then compile any latex project.

image

tamuratak:
CompilerLogParser.parse should not throw errors. Close #2498.

### DIFF
--- a/src/components/parser/latexlog.ts
+++ b/src/components/parser/latexlog.ts
@@ -68,7 +68,15 @@ export class LatexLogParser {
 
    private parseLine(line: string, state: ParserState, buildLog: LogEntry[]) {
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
-        const excludeRegexp = (configuration.get('message.latexlog.exclude') as string[]).map(regexp => RegExp(regexp))
+        let excludeRegexp: RegExp[]
+        try {
+            excludeRegexp = (configuration.get('message.latexlog.exclude') as string[]).map(regexp => RegExp(regexp))
+        } catch (e) {
+            if (e instanceof Error) {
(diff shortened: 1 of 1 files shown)


jlelong:
I am wondering whether it would not be sufficient to catch the errors raised by

https://github.com/James-Yu/LaTeX-Workshop/blob/c7d12eed41b5bfac8335f21736a348e0d7c80f1b/src/components/parser/latexlog.ts#L99-L103

tamuratak:
`RegExp(regexp)`  in this line throws errors we are seeing:

https://github.com/James-Yu/LaTeX-Workshop/blob/addcc3895a0b8e75995330cddf872c28feddd672/src/components/parser/latexlog.ts#L71

`RegExp` cannot compile `regexp` because it is invalid as a regular expression.

I think any other errors during executing `CompilerLogParser.parse` can lead to the same issue.

jlelong:
Good point for the construction of the regexp `RegExp(regexp)`!

Of course, it does not harm to catch errors in `CompilerLogParser.parse` but I cannot really how it could happen.

tamuratak:
I have edited as suggested. Catch errors thrown by `RegExp(regexp)`.

tamuratak:
> Of course, it does not harm to catch errors in `CompilerLogParser.parse` but I cannot really how it could happen.

Even if it does not happen now, it would happen unintentionally when we add codes to the parser in the future.

tamuratak:
I merge this PR. It is a good practice to catch errors near where they occur.

PR #2508, opened 2021-02-04
Sort bibtex fields

### LINKED ISSUE #2490: Sort bibtex fields
Related to #2481. @alexreg

Sort bibtex fields alphabetically or following an order specified as an array of field names, e.g., ["title", "author", "journal", "volume", ...].

jlelong:
Close #2490

Sorting fields is enabled by `latex-workshop.bibtex-fields.sort.enabled`. This is only taken into account when also aligning fields. The order is defined by `latex-workshop.bibtex-fields.order` and then alphabetically for not listed fields.

### DIFF
--- a/src/providers/bibtexformatter.ts
+++ b/src/providers/bibtexformatter.ts
@@ -73,6 +73,7 @@ export class BibtexFormatter {
             sortFields: config.get('bibtex-fields.sort.enabled') as boolean,
             fieldsOrder: config.get('bibtex-fields.order') as string[]
         }
+        this.extension.logger.addLogMessage(`Bibtex format config: ${JSON.stringify(configuration)}`)
         const lineOffset = range ? range.start.line : 0
         const columnOffset = range ? range.start.character : 0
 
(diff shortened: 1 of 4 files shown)


src/providers/bibtexformatter.ts
```diff
@@ -69,7 +69,9 @@ export class BibtexFormatter {
             right: leftright[1],
             trailingComma: config.get('bibtex-format.trailingComma') as boolean,
             sort: config.get('bibtex-format.sortby') as string[],
-            alignOnEqual: config.get('bibtex-format.align-equal.enabled') as boolean
+            alignOnEqual: config.get('bibtex-format.align-equal.enabled') as boolean,
+            sortFields: config.get('bibtex-fields.sort.enabled') as boolean,
+            fieldsOrder: config.get('bibtex-fields.order') as string[]
         }
```
tamuratak, line 75: Here, we should add the `configuration` variable to log messages with `addLogMessage`.

tamuratak (changes requested):

tamuratak (approved):

Issue #499, opened 2018-03-26
Feature Request: Autofix `$ $` to `\( \)` and `$$ $$` to `\[ \]`


chantisnake:
chktex will mark this as warning, but you cannot auto-fix it.

jlelong:
`chktex` only reports bad `LaTeX` constructions along with a suggestion to fix it, but no auto-fix suggestion. Some linters are able to fix some warnings by themselves, but unfortunately `chktex` cannot. Implementing an auto-fix feature directly in the extension does not really make sense and would really require a lot of work, actually redoing most of the linting process.

Although this is a desirable feature, I think it is not feasible with `chktex`. I am closing the issue. Feel free to reopen it if you have other ideas or suggestions.

jlelong:
It can be done quite easily for when the maths expression `$...$` or `$$...$$` is on a single line as `chktex` detects the whole expression. For multiline constructions like
```
$$
.....
$$
````
only the opening `$$` are highlighted. Auto-fixing would require to parse the document to detect the end pattern `$$`. Do not know if it is worth it.

I suggest to start with single line constructions.

PR #2512, opened 2021-02-07
Enable autofix on multiline maths

jlelong:
Related to #499.

Autofix `$ $` to `\( \)` and `$$ $$` to `\[ \]` works for multiline expressions

### DIFF
--- a/src/providers/codeactions.ts
+++ b/src/providers/codeactions.ts
@@ -1,7 +1,6 @@
 import * as vs from 'vscode'
 
 import type { Extension } from '../main'
-import { TeXMathEnvFinder } from '../providers/preview/mathpreviewlib/texmathenvfinder'
 
 const CODE_TO_ACTION_STRING: {[key: number]: string} = {
     1: 'Terminate command with empty statement',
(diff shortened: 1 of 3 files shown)


src/providers/codeactions.ts
```diff
@@ -1,6 +1,7 @@
 import * as vs from 'vscode'
 
 import type { Extension } from '../main'
+import { TeXMathEnvFinder } from '../providers/preview/mathpreviewlib/texmathenvfinder'
```
tamuratak, line 4: We should not directly import `mathpreviewlib/texmathenvfinder`.

Please make `MathPreview.texMathEnvFinder` public and use `Extension.mathPreview.texMathEnvFinder`.

https://github.com/James-Yu/LaTeX-Workshop/blob/e3c795752b10d0f6dd56df9280002b16bad8b57f/src/providers/preview/mathpreview.ts#L19-L29

tamuratak (changes requested):

jlelong:
I have made the necessary changes. For the sake of coherence, I have move all the `replaceXXX` functions inside the `CodeActions` class.

src/providers/codeactions.ts
jlelong: I have made the necessary changes. For the sake of coherence, I have move all the `replaceXXX` functions inside the `CodeActions` class.

src/providers/codeactions.ts
```diff
@@ -172,16 +132,64 @@ export class CodeActions {
...
-                replaceRangeWithString(document, range, fixString)
+                this.replaceRangeWithString(document, range, fixString)
                 break
             case 45:
-                replaceMathDelimitersInRange(document, range, '$$', '\\[', '\\]')
+                this.replaceMathDelimitersInRange(document, range, '$$', '\\[', '\\]')
                 break
             case 46:
-                replaceMathDelimitersInRange(document, range, '$', '\\(', '\\)')
+                this.replaceMathDelimitersInRange(document, range, '$', '\\(', '\\)')
```
tamuratak, line 141: We should not call `texMathEnvFinder` for `$`. It does not work.

tamuratak (changes requested):

tamuratak:
This PR does not work on some rare cases. For example:

```latex
$ x \$ $
```

```latex
$$
x
$$$y$
```

It would be appropriate to use `Extension.pegParser.parseLatex` instead of `texMathEnvFinder`. I don't strongly push using `parseLatex`.

src/providers/codeactions.ts
jlelong: For standard expressions like

```latex
Some maths $1 + 3 =
4$ within text
```

 it is working. Can you give a failing example?

tamuratak: I mean, it is not intended behavior. The behavior might change in the future refactoring.

jlelong:
Actually, `texMathEnvFinder.findEndPair` should use `Extension.pegParser.parseLatex` and only use the current implementation as a fallback. Let's keep it for an other PR

src/providers/codeactions.ts
jlelong: Fine, we can drop the support for `$...$` split over several lines. This is consistent with the hovering feature.

tamuratak (approved):

src/providers/codeactions.ts
```diff
@@ -172,16 +132,68 @@ export class CodeActions {
...
+        const edit = new vs.WorkspaceEdit()
+        edit.replace(document.uri, range, replacementString)
+        return vs.workspace.applyEdit(edit)
+    }
+
+    private replaceRangeWithRepeatedString(document: vs.TextDocument, range: vs.Range, replacementString: string) {
+        return this.replaceRangeWithString(document, range, replacementString.repeat(range.end.character - range.start.character))
+    }
+
+    private replaceMathDelimitersInRange(document: vs.TextDocument, range: vs.Range, oldDelim: string, startDelim: string, endDelim: string) {
```
tamuratak, line 172: It would be better to change the type of `oldDelim` from `string` to `'$' | '$$'`.

tamuratak (changes requested):

jlelong:
```latex
$$
x
$$$y$
```
works fine now.

tamuratak (approved):

PR #2531, opened 2021-02-16
Check intellisense json files are valid

### LINKED ISSUE #2530: We must take care of cases where user-defined JSON files are invalid
We must take care of cases where user-defined JSON files are invalid. We must check whether objects returned by `JSON.parse` have appropriate properties. Because the TypeScript type of returned objects is `any`, the TypeScript compiler can do nothing on it.

https://github.com/James-Yu/LaTeX-Workshop/blob/c5b6cad0f478a5ee27f9c1480bacc5bdc81cdc62/src/providers/completer/environment.ts#L231

https://github.com/James-Yu/LaTeX-Workshop/blob/c5b6cad0f478a5ee27f9c1480bacc5bdc81cdc62/src/providers/completer/command.ts#L341

jlelong:
Close #2530

### DIFF
--- a/src/providers/completer/command.ts
+++ b/src/providers/completer/command.ts
@@ -349,7 +349,7 @@ export class Command implements IProvider {
                             this.packageCmds[pkg].push(this.entryCmdToCompletion(key, cmds[key]))
                         } else {
                             this.extension.logger.addLogMessage(`Cannot parse intellisense file: ${filePath}`)
-                            this.extension.logger.addLogMessage(`Missing field in entry: ${JSON.stringify(cmds[key])}`)
+                            this.extension.logger.addLogMessage(`Missing field in entry: "${key}": ${JSON.stringify(cmds[key])}`)
                         }
                     })
                 } catch (e) {
--- a/src/providers/completer/command.ts
+++ b/src/providers/completer/command.ts
@@ -18,6 +18,10 @@ interface CmdItemEntry {
     postAction?: string
 }
 
+function isCmdItemEntry(obj: any): obj is CmdItemEntry {
+    return (typeof obj.command === 'string') && (typeof obj.snippet === 'string')
+}
+
 export interface Suggestion extends vscode.CompletionItem {
     package: string
 }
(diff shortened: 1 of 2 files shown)


src/providers/completer/command.ts
```diff
@@ -338,10 +342,19 @@ export class Command implements IProvider {
...
-                    this.packageCmds[pkg].push(this.entryCmdToCompletion(key, cmds[key]))
-                })
+                try {
+                    const cmds: {[key: string]: CmdItemEntry} = JSON.parse(fs.readFileSync(filePath).toString())
+                    Object.keys(cmds).forEach(key => {
+                        if (isCmdItemEntry(cmds[key])) {
+                            this.packageCmds[pkg].push(this.entryCmdToCompletion(key, cmds[key]))
+                        } else {
+                            this.extension.logger.addLogMessage(`Cannot parse intellisense file: ${filePath}`)
+                            this.extension.logger.addLogMessage(`Missing field in entry: ${JSON.stringify(cmds[key])}`)
```
tamuratak, line 352: We should also add the value of `key` to log messages.

src/providers/completer/environment.ts
```diff
@@ -228,8 +232,20 @@ export class Environment implements IProvider {
...
             return {}
         }
-        const envs: {[key: string]: EnvItemEntry} = JSON.parse(fs.readFileSync(filePath).toString())
-        return envs
+        try {
+            const envs: {[key: string]: EnvItemEntry} = JSON.parse(fs.readFileSync(filePath).toString())
+            Object.keys(envs).forEach(key => {
+                if (! isEnvItemEntry(envs[key])) {
+                    this.extension.logger.addLogMessage(`Cannot parse intellisense file: ${filePath}`)
+                    this.extension.logger.addLogMessage(`Missing field in entry: ${JSON.stringify(envs[key])}`)
```
tamuratak, line 240: We should also add the value of `key` to log messages.

tamuratak (changes requested):

tamuratak (approved):

PR #2534, opened 2021-02-21
Enable access to Web-View from remote machines

mzietlow:
Is there a reason that the Web-Viewer is listening to 127.0.0.1?
To enable access from remote machines I found it useful to instead listen to 0.0.0.0


tamuratak:
Out of the scope of the extension.

PR #2536, opened 2021-02-22
Check for duplicate labels

### LINKED ISSUE #2505: labels in LaTeX are unique
A named label, as `\label{eq:MyEquation}`, is unique in a LaTeX file; otherwise, the reference is ambiguous and causes a warning.
**Feature request:** 
If `\label{eq:MyEquation}` appears twice in a .tex-file, then every occurrence of `eq:MyEquation` should be highlighted in red color, say, as a warning.
This useful feature facilitates writing LaTeX as tracking reference errors is superfluous. In addition, the new feature removes a slight advantage of TeXstudio over LaTeX Workshop.
Thank you!

jlelong:
Close #2505. This PR checks for duplicate labels, which are computed based on intellisense data fro references.

### DIFF
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -256,6 +256,7 @@ export class Manager {
                 this.rootFile = rootFile
                 this.rootFileLanguageId = this.inferLanguageId(rootFile)
                 this.extension.logger.addLogMessage(`Root file languageId: ${this.rootFileLanguageId}`)
+                this.extension.duplicateLabels.reset()
                 this.initiateFileWatcher()
                 this.bibWatcher.initiateBibWatcher()
                 this.parseFileAndSubs(this.rootFile) // finish the parsing is required for subsequent refreshes.
@@ -726,7 +727,12 @@ export class Manager {
         this.extension.logger.addLogMessage(`Added to file watcher: ${file}`)
         if (['.tex', '.bib'].concat(this.weaveExt).includes(path.extname(file)) &&
             !file.includes('expl3-code.tex')) {
-            this.updateCompleterOnChange(file)
+            this.updateCompleterOnChange(file).then(() => {
+                const configuration = vscode.workspace.getConfiguration('latex-workshop')
+                if (['onIntellisenseUpdate', 'onSave'].includes(configuration.get('check.duplicatedLabels.run') as string)) {
+                    this.extension.duplicateLabels.run(file)
+                }
+            })
         }
     }
 
@@ -736,7 +742,12 @@ export class Manager {
         if (['.tex', '.bib'].concat(this.weaveExt).includes(path.extname(file)) &&
             !file.includes('expl3-code.tex')) {
             this.parseFileAndSubs(file, true)
-            this.updateCompleterOnChange(file)
+            this.updateCompleterOnChange(file).then(() => {
+                const configuration = vscode.workspace.getConfiguration('latex-workshop')
+                if (configuration.get('check.duplicatedLabels.run') === 'onSave') {
+                    this.extension.duplicateLabels.run(file)
+                }
+            })
         }
         this.buildOnFileChanged(file)
     }
@@ -777,9 +788,9 @@ export class Manager {
     }
 
     // This function updates all completers upon tex-file changes.
-    private updateCompleterOnChange(file: string) {
-        fs.readFile(file).then(buffer => buffer.toString()).then(content => this.updateCompleter(file, content))
+    private async updateCompleterOnChange(file: string) {
         this.extension.completer.input.getGraphicsPath(file)
+        return fs.readFile(file).then(buffer => buffer.toString()).then(content => this.updateCompleter(file, content))
     }
 
     /**
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -75,7 +75,7 @@ export class Manager {
     private fileWatcher?: chokidar.FSWatcher
     private readonly pdfWatcher: PdfWatcher
     private readonly bibWatcher: BibWatcher
-    readonly intellisenseWatcher: IntellisenseWatcher
+    private readonly intellisenseWatcher: IntellisenseWatcher
     private readonly finderUtils: FinderUtils
     private readonly pathUtils: PathUtils
     private filesWatched: string[] = []
@@ -757,6 +757,10 @@ export class Manager {
             this.extension.logger.addLogMessage('Start searching a new root file.')
             this.findRoot()
         }
+    }
+
+    onDidUpdateIntellisense(cb: (file: string) => void) {
+        return this.intellisenseWatcher.onDidUpdateIntellisense(cb)
     }
 
     watchPdfFile(pdfPath: string) {
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -75,7 +75,7 @@ export class Manager {
     private fileWatcher?: chokidar.FSWatcher
     private readonly pdfWatcher: PdfWatcher
     private readonly bibWatcher: BibWatcher
-    private readonly intellisenseWatcher: IntellisenseWatcher
+    readonly intellisenseWatcher: IntellisenseWatcher
     private readonly finderUtils: FinderUtils
     private readonly pathUtils: PathUtils
     private filesWatched: string[] = []
@@ -100,7 +100,7 @@ export class Manager {
         }
         this.pdfWatcher = new PdfWatcher(extension)
         this.bibWatcher = new BibWatcher(extension)
-        this.intellisenseWatcher = new IntellisenseWatcher(extension)
+        this.intellisenseWatcher = new IntellisenseWatcher()
         this.finderUtils = new FinderUtils(extension)
         this.pathUtils = new PathUtils(extension)
     }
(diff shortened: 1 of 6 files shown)


src/providers/duplicatelabels.ts
```diff
@@ -0,0 +1,101 @@
...
+    [key: string]: {
+        file: string,
+        range: vscode.Range
+    }[]
+}
+
+export class DuplicateLabels {
+    private readonly extension: Extension
+    private readonly duplicatedLabelsDiagnostics = vscode.languages.createDiagnosticCollection('Duplicate Labels')
+    private labelEntries: LabelEntries = {}
```
tamuratak, line 16: Why is `labelEntries` a property of `DuplicateLabels` class? How about making `parseFileLabels` return it and `showDiagnostics` called with it as an argument?

tamuratak (changes requested):

src/components/manager.ts
```diff
@@ -736,7 +742,12 @@ export class Manager {
         if (['.tex', '.bib'].concat(this.weaveExt).includes(path.extname(file)) &&
             !file.includes('expl3-code.tex')) {
             this.parseFileAndSubs(file, true)
-            this.updateCompleterOnChange(file)
+            this.updateCompleterOnChange(file).then(() => {
+                const configuration = vscode.workspace.getConfiguration('latex-workshop')
+                if (configuration.get('check.duplicatedLabels.run') === 'onSave') {
+                    this.extension.duplicateLabels.run(file)
+                }
+            })
```
tamuratak, line 750: We should use `async` and `await`.

src/components/manager.ts
```diff
@@ -720,13 +720,19 @@ export class Manager {
...
         this.extension.logger.addLogMessage(`Added to file watcher: ${file}`)
         if (['.tex', '.bib'].concat(this.weaveExt).includes(path.extname(file)) &&
             !file.includes('expl3-code.tex')) {
-            this.updateCompleterOnChange(file)
+            this.updateCompleterOnChange(file).then(() => {
+                const configuration = vscode.workspace.getConfiguration('latex-workshop')
+                if (['onIntellisenseUpdate', 'onSave'].includes(configuration.get('check.duplicatedLabels.run') as string)) {
+                    this.extension.duplicateLabels.run(file)
+                }
+            })
```
tamuratak, line 735: We should use `async` and `await`.

tamuratak (changes requested):

tamuratak (changes requested):

src/main.ts
```diff
@@ -200,7 +201,11 @@ export function activate(context: vscode.ExtensionContext) {
             }
             updateCompleter = setTimeout(() => {
                 const file = e.document.uri.fsPath
-                extension.manager.updateCompleter(file, content)
+                extension.manager.updateCompleter(file, content).then(() => {
+                    if (configuration.get('check.duplicatedLabels.run') === 'onIntellisenseUpdate') {
+                        extension.duplicateLabels.run(file)
+                    }
+                })
```
tamuratak, line 208: We should use `async` and `await`.

src/main.ts
```diff
@@ -200,7 +201,11 @@ export function activate(context: vscode.ExtensionContext) {
             }
             updateCompleter = setTimeout(() => {
```
tamuratak, line 202: We should rename `updateCompleter` to an appropriate one.

tamuratak (changes requested):

src/components/manager.ts
jlelong: I cannot really see the benefit of using `async` and `await` over `.then`. Can you expand a bit on this?

src/providers/duplicatelabels.ts
jlelong: `labelEntries` is a property of `DuplicateLabels` because it holds the labels for all the project files and we only update the labels defined in the file from which the check was triggered.

> I don't think caching them here has benefits. It is enough to use `this.extension.manager.cachedContent[file].element.reference` every time.

I do not really understand your point. Can you expand a little?

src/components/manager.ts
tamuratak: It makes codes more readable.

src/providers/duplicatelabels.ts
tamuratak: We should avoid using caches if possible. Using caches requires us to ensure caches to be consistent with updates. We easily fail to achieve that, which leads to bugs.

I think it can be acceptable even if the code becomes much slower.

jlelong: I do feel concerned by the computational cost of every feature, which led me to adopt a partial update approach. However, to deal with the removal of inclusions as `\include{myfile}`, we need to remove any references to files not in `manager.getIncludedTeX()`

https://github.com/James-Yu/LaTeX-Workshop/blob/292fe7ee0258441f55a7bf7f9e25fbf0096f24e0/src/providers/duplicatelabels.ts#L55-L58

Then, I am not sure it still makes sense to adopt a per file approach rather than recomputing `labelEntries` every time.

If I understood you well, this is also your point, isn't it?

tamuratak: I am saying that we don't have to feel concerned by the computational cost here. We should be concerned about whether the changes lead to bugs here.

At least, we should stop adding cache here and there ad hoc. If you want to add `labelEntries` somewhere, it seems to me appropriate that `labelEntries` becomes a property of `this.extension.manager`. We should update `labelEntries` every time `this.extension.manager.cachedContent[file].element.reference` updated.

tamuratak: Another solution is that extracting a class having the `labelEntries` property from `DuplicateLabels` class into `src/components/managerlib/`. 

I guess that we will have to access `labelEntries` for purposes other than duplicate labels.

jlelong:
When `intellisense.update.aggressive.enabled` is set to `true`, we call `extension.manager.updateCompleter` every time the user stops typing for long enough. Yet, the list of children is not update

https://github.com/James-Yu/LaTeX-Workshop/blob/f333832669f3671f8a95162faecf00b9d221048e/src/main.ts#L202-L210

To be compared with https://github.com/James-Yu/LaTeX-Workshop/blob/f333832669f3671f8a95162faecf00b9d221048e/src/components/manager.ts#L738-L744

Hence, the list returned by `extension.manager.getIncludedTeX()` is wrong every time the change involves a `\include` addition or removal. This prevents from implementing the duplicate labels check when there is no typing for long enough.

Should not we call `manager.parseFileAndSubs(file)` before https://github.com/James-Yu/LaTeX-Workshop/blob/f333832669f3671f8a95162faecf00b9d221048e/src/main.ts#L204

tamuratak:
It suggests that we should implement something like `onDidIntellisenseUpdated` using `EventEmitter`. See https://nodejs.org/docs/latest-v12.x/api/events.html#events_class_eventemitter

jlelong:
That sounds like a smart approach to implement a `onDidUpdatedIntellisense` listener. I will do so.

However, my point above was also that currently the intellisense data is only properly updated when a file is saved even when `intellisense.update.aggressive.enabled` is set to true. Consider the following situation with two `.tex` file: `main.tex` and `.sub.tex`. Add `\include{sub.tex}` to `main.tex`, no intellisense data defined in `sub.tex` will be available before `main.tex` is actually saved even with `intellisense.update.aggressive.enabled` set to true. This is because the list of children of `main.tex` will only be updated when it is saved. I believe this is a bug.

tamuratak:
I see. Thank you.

tamuratak:
In the current implementation, we don't have to use `EventEmitter`. It is enough that `emitUpdate` directly calls `onDidUpdateIntellisense`.


I assumed the following one similar to VS Code API.

```ts
import * as events from 'events'

export class IntellisenseWatcher {
    private readonly intellisenseEmitter: events.EventEmitter
    private readonly cbSet: Set<(file: string) => void> = new Set()

    constructor() {
        this.intellisenseEmitter = new events.EventEmitter()
        this.intellisenseEmitter.on('update', (file: string) => {
            this.cbSet.forEach((cb) => {
                cb(file)
            })
        })
    }

    onDidUpdateIntellisense(cb: (file: string) => void) {
        this.cbSet.add(cb)
        const diposable = {
            dipose: () => { this.cbSet.delete(cb) }
        }
        return diposable
    }

    emitUpdate(file: string) {
        this.intellisenseEmitter.emit('update', file)
    }
}
```

jlelong:
It seems we are almost there. I suggest to tackle in an other PR the bug regarding the intellisense data update when `intellisense.update.aggressive.enabled` is `true`. I would like to investigate if we can avoid reparsing all the included files.

I still need to figure out if there is a real interest in caching `labelEntries` for the sake of computational savings.

jlelong:
@tamuratak Does it better correspond to what you had in mind?

jlelong:
> I still need to figure out if there is a real interest in caching `labelEntries` for the sake of computational savings.

Before doing so, it is better to fix the intellisense update bug to see how much can be gained by a smart update the `children` field. I will come back to this PR when the above is done.

src/providers/duplicatelabels.ts
jlelong: I have simplified the computation of duplicates and do not use cache anymore.

tamuratak:
I am offline for days.

Issue #2047, opened 2020-04-22
\mathrm and other commands are missing in the completion items of the current master


tamuratak:
`\mathrm` and other commands are missing in the completion items of the current master. `data/packages/latex-document_cmd.json` is not loaded. 

![スクリーンショット 2020-04-22 15 20 41](https://user-images.githubusercontent.com/10665499/79948149-44b35200-84ae-11ea-9d27-a702bb71ffdc.png)

jlelong:
You are right. It seems that `data/packages/latex-document_cmd.json` has never been loaded, which created this issue when I sanitized the file `data/commands.json`. Maybe I was a bit too agressive.

jlelong:
Actually, it turns out that 3 major completion files are never loaded:
```
latex-dev_cmd.json
latex-document_cmd.json
latex-document_env.json
```

It is clear that the last two must be loaded by default but I am not so sure about `latex-dev_cmd.json`

tamuratak:
I was thinking that it was intentional not loading `latex-document_cmd.json`, which enables us to control snippets of basic LaTeX commands by hand. I do not know much about snippets of LaTeX Workshop.

There are two things to say.

Now, we see duplicates of some commands.

![スクリーンショット 2020-04-22 19 59 44](https://user-images.githubusercontent.com/10665499/79974226-e305de80-84d3-11ea-8b77-d92f6cd42393.png)

I prefer less used commands, `\left\backslash` and others, to be removed since they push out `\left\{` from the suggestion.

![スクリーンショット 2020-04-22 20 01 23](https://user-images.githubusercontent.com/10665499/79974382-37a95980-84d4-11ea-9616-805dbfab0361.png)

jlelong:
Good point! 
So it may be wiser not to load `latex-document_{cmd,env}.json` but to only add some of their entries into `commands.json` and `environments.json`.

jlelong:
>  I do not know much about snippets of LaTeX Workshop.

There are two different systems working in parallel

1. The standard snippet mechanism of VSCode based on the content of `snippets/latex.json`. Completion based on this file is triggered after two characters are typed when `editor.suggest.showSnippets` is true.

1. The internal LaTeX-Workshop completion provider triggered by `\`. The data used to populated completion come from the directory `data`. The following files are always loaded

    - `bibtex-entries.json`: completion in bibtex files
    - `classnames.json`: list of LaTeX classes, used for completion in `\documentclass`
[...]

PR #2557, opened 2021-03-03
Remove duplicate commands in data/commands.json 

tamuratak:
Remove duplicate commands in ` data/commands.json`. They are also defined in `data/packages/latex-document_cmd.json`.

We can find them with `jq`.

```
jq -s add data/commands.json data/packages/latex-document_cmd.json | jq 'to_entries | map(.value | .snippet) |  sort | group_by(.) | map(select(length>1) | .[0])'
```

### DIFF
--- a/data/commands.json
+++ b/data/commands.json
@@ -108,10 +108,6 @@
     "snippet": "{${1}\\\\}",
     "detail": "curly brackets \\{ ... \\}"
   },
-  "title": {
-    "command": "title",
-    "snippet": "title{${1:text}}"
-  },
   "part": {
     "command": "part",
     "snippet": "part{${1}}"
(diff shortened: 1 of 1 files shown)


jlelong:
The question of duplicate entries has already been discussed in #2047, in which we concluded that it may be useful to keep standard commands in `data/commands.json` aven though they are also defined in `data/packages/latex-document_cmd.json`. We should keep in mind that `data/packages/latex-document_cmd.json` is only loaded when `latex-workshop.intellisense.package.enabled` is true.

If we want to remove duplicates here, then we probably need refactor the way we are loading `data/packages/latex-document_{cmd,env}.json`.

tamuratak:
> If we want to remove duplicates here, then we probably need refactor the way we are loading data/packages/latex-document_{cmd,env}.json.

I agree.

PR #2561, opened 2021-03-05
Fix a regexp which might causes ReDoS

tamuratak:
CodeQL warns the following regular expression might cause the regular expression denial of service. https://en.wikipedia.org/wiki/ReDoS

```regexp
(?:(?:\[[^[\]]*\])*(?:{[^{}]*})?)*
```

This PR fixes it by changing to the following:

```regexp
(?:\[[^[\]]*\]|{[^{}]*})*
```


jlelong (approved):
It looks fine to me

PR #2562, opened 2021-03-05
Fix potential exponential backtracking when removing comments

jlelong:
Related to CodeQL inefficient regular expression warning.

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -27,7 +27,7 @@ export function escapeRegExp(str: string) {
  * Note the number lines of the output matches the input
  */
 export function stripComments(text: string, commentSign: string = '%'): string {
-    const pattern = '([^\\\\]|^)' + commentSign + '.*$'
+    const pattern = '(^|(?:(?<!\\\\)(?:\\\\\\\\)+))' + commentSign + '.*$'
     const reg = RegExp(pattern, 'gm')
     return text.replace(reg, '$1')
 }
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -22,13 +22,11 @@ export function escapeRegExp(str: string) {
  * Remove comments
  *
  * @param text A string in which comments get removed.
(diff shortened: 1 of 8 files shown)


tamuratak:
I think the function has a bug. It does not remove a comment immediately after `\\`.

https://github.com/James-Yu/LaTeX-Workshop/blob/54fa70f22910d0fe50e991d919e4845c63694db7/src/utils/utils.ts#L29-L33

```latex
\begin{align*}
x \\%abc
y
\end{align*}
```

jlelong:
Indeed! This bug has been around for two years and nobody ever complained.

jlelong:
Not yet!

src/utils/utils.ts
```diff
@@ -26,8 +26,8 @@ export function escapeRegExp(str: string) {
  * @return the input text with comments removed.
  * Note the number lines of the output matches the input
  */
-export function stripComments(text: string, commentSign: string): string {
-    const pattern = '([^\\\\]|^)' + commentSign + '.*$'
+export function stripComments(text: string, commentSign: string = '%'): string {
+    const pattern = '(^|[^\\\\]|(?:(?<!\\\\)\\\\\\\\))' + commentSign + '.*$'
```
tamuratak, line 30: I don't like the API. It makes things too complicated unnecessarily. Please delete the `commentSign` argument and use a literal regular expression. When we need to remove comments which begin other than `%`, let's define another function.

tamuratak (changes requested):

tamuratak:
It does not remove the following comment.

```latex
\begin{align*}
x \\\\%abc
y
\end{align*}
```

jlelong:
This was done on purpose because I thought two consecutive newlines `\\` were not valid LaTeX. Although, it is not inside `align`, it is in text mode. I will fix it.

jlelong:
It should be fine now.

tamuratak (approved):

PR #2585, opened 2021-03-17
input.ts: use nul-separated output

Artoria2e5:
Although this shouldn't be exploitable, it is still perfectly possible for someone to have a newline in the filename and get that fed through git. When the extension does a .split('\n') it becomes incorrectly interpreted as two files.

### DIFF
--- a/src/providers/completer/input.ts
+++ b/src/providers/completer/input.ts
@@ -32,8 +32,8 @@ export class Input implements IProvider {
                     cp.execSync('which git')
                 }
                 const command = 'git'
-                const args = ['check-ignore', '-z', '--'].concat(files)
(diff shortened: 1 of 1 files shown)


Artoria2e5:
Well, actually this breaks in a different way:
```console
$ echo '._*' > .gitignore
$ touch $'._\nthats_a_newline'
$ git check-ignore ._*
"._\nthats_a_newline"
```
Since the command is designed for human use, git escapes some unsavory characters. That's bad news for the string matching check.

tamuratak:
@jlelong  I think we had better delete `filterIgnoredFiles`.

jlelong:
@tamuratak I agree

PR #2599, opened 2021-03-28
Fix typo in hat (@^) snippet

r0uv3n:
It seems that this is a typo. At least I can't find a command \Hat in symbols-a4.


jlelong:
You are right `\Hat` is not referenced in `symbols-a4` but it is defined by `amsmath` as a synonymous of `\hat` for backward compatibility. Anyway, it is better to use `\hat`.

PR #2628, opened 2021-04-19
Support <<child='...'>>= for .rnw file inclusion 

### LINKED ISSUE #2607: Inclusion of Rnw files using <<child=...>> does not work
I am working with multiple Rnw files. The structure overview shows the root tex file that gets created after knitting. This is not helpful in navigation as I want to navigate inside the Rnw and not the tex file.

Let the structure overview show the contents of the Rnw. Maybe it could detect the root magic 

`%!Rnw root=index.Rnw`

**Additional context**
I've seen that a content for an Rnw file is only produced when \begin{documtent} is present. But as my Rnw is a child, there is no \begin{document}. 

Screenshot from 2021-04-07 15-11-28

jlelong:
To fully support the use of `<<child='...'>>=` for `.rnw` file inclusion, we need to integrate it to

- [x] the rootFile detection mechanism
- [x] the structure provider
- [x] magic comments : accept files with a non `.tex` extension

Close #2607

### DIFF
--- a/src/components/managerlib/pathutils.ts
+++ b/src/components/managerlib/pathutils.ts
@@ -19,34 +19,27 @@ export interface MatchPath {
     index: number
 }
 
-export class PathUtils {
-    private readonly extension: Extension
-    readonly inputRegex: RegExp
-    readonly childRegex: RegExp
+export class PathRegExp {
+    private readonly inputRegexp: RegExp
+    private readonly childRegexp: RegExp
 
-    constructor(extension: Extension) {
-        this.extension = extension
-        this.inputRegex = /\\(?:input|InputIfFileExists|include|SweaveInput|subfile|(?:(?:sub)?(?:import|inputfrom|includefrom)\*?{([^}]*)}))(?:\[[^[\]{}]*\])?{([^}]*)}/g
-        this.childRegex = /<<(?:[^,]*,)*\s*child='([^']*)'\s*(?:,[^,]*)*>>=/g
+    constructor() {
+        this.inputRegexp = /\\(?:input|InputIfFileExists|include|SweaveInput|subfile|(?:(?:sub)?(?:import|inputfrom|includefrom)\*?{([^}]*)}))(?:\[[^[\]{}]*\])?{([^}]*)}/g
+        this.childRegexp = /<<(?:[^,]*,)*\s*child='([^']*)'\s*(?:,[^,]*)*>>=/g
     }
 
-    private get rootDir() {
-        return this.extension.manager.rootDir
-    }
-
-    private getOutDir(texFile: string) {
-        return this.extension.manager.getOutDir(texFile)
+    resetLastIndex() {
+        this.inputRegexp.lastIndex = 0
+        this.childRegexp.lastIndex = 0
     }
 
     /**
      * Return the matched input or child path. If there is no match, return undefined
      *
-     * @param inputReg a copy of this.inputRegex. We must use a copy to properly handle the .lastIndex property of RegExp.
-     * @param childReg a copy of this.childRegex. We must use a copy to properly handle the .lastIndex property of RegExp.
      * @param content the string to match the regex on
      */
-    exec(inputReg: RegExp, childReg: RegExp, content: string): MatchPath | undefined {
-        let result = inputReg.exec(content)
+    exec(content: string): MatchPath | undefined {
+        let result = this.inputRegexp.exec(content)
         if (result) {
             return {
                 type: MatchType.Input,
@@ -56,7 +49,7 @@ export class PathUtils {
                 index: result.index
             }
         }
-        result = childReg.exec(content)
+        result = this.childRegexp.exec(content)
         if (result) {
             return {
                 type: MatchType.Child,
@@ -93,6 +86,27 @@ export class PathUtils {
             }
         }
         return undefined
+    }
+
+}
+
+export class PathUtils {
+    private readonly extension: Extension
+    readonly inputRegex: RegExp
+    readonly childRegex: RegExp
+
+    constructor(extension: Extension) {
+        this.extension = extension
+        this.inputRegex = /\\(?:input|InputIfFileExists|include|SweaveInput|subfile|(?:(?:sub)?(?:import|inputfrom|includefrom)\*?{([^}]*)}))(?:\[[^[\]{}]*\])?{([^}]*)}/g
+        this.childRegex = /<<(?:[^,]*,)*\s*child='([^']*)'\s*(?:,[^,]*)*>>=/g
+    }
+
+    private get rootDir() {
+        return this.extension.manager.rootDir
+    }
+
+    private getOutDir(texFile: string) {
+        return this.extension.manager.getOutDir(texFile)
     }
 
     /**
--- a/src/components/managerlib/pathutils.ts
+++ b/src/components/managerlib/pathutils.ts
@@ -23,12 +23,18 @@ export class PathUtils {
 
     parseInputFilePath(regResult: RegExpExecArray, currentFile: string, rootFile: string): string | undefined {
         const texDirs = vscode.workspace.getConfiguration('latex-workshop').get('latex.texDirs') as string[]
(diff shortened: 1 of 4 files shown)


src/components/managerlib/pathutils.ts
```diff
@@ -8,9 +8,11 @@ import type {Extension} from '../../main'
 
 export class PathUtils {
     private readonly extension: Extension
+    readonly inputRegex: RegExp
 
     constructor(extension: Extension) {
         this.extension = extension
+        this.inputRegex = /(?:(?:\\(?:input|InputIfFileExists|include|SweaveInput|subfile|(?:(?:sub)?(?:import|inputfrom|includefrom)\*?{([^}]*)}))(?:\[[^[\]{}]*\])?){([^}]*)})|(?:<<(?:[^,]*,)*\s*child='([^']*)'\s*(?:,[^,]*)*>>=)/g
```
tamuratak, line 15: Making a regexp longer increases the maintenance burden. We should make it shorter if possible.

For example, with the following prototype, we can keep each regexp, `regA` and `regB`, short.
```ts
class A {
    private regA: RegExp
    private regB: RegExp

    match(s: string) {
        let match = this.regA.exec(s)
        if (match) {
            ...
            return ...
        } 
        match = this.regB.exec(s)
        if (match) {
            ...
            return ...
        }
    }
}
```

jlelong: I do agree. Indeed, the real maintenance burden is 

https://github.com/James-Yu/LaTeX-Workshop/blob/bfd87dadd0d01d1ff4a003b0e2257a7c45d67899/src/providers/structure.ts#L94-L101

especially because of the double escaping. It would be better to split the regex into a sectioning regex and an input regex. I will think of it

src/components/managerlib/pathutils.ts
```diff
@@ -21,8 +25,21 @@ export class PathUtils {
...
     }
 
+    /**
+     * Compute the resolved file path from matches of this.inputReg or this.childReg
+     *
+     * @param regResult is the the result of this.inputReg.exec() or this.childReg.exec()
+     * @param currentFile is the name of file in which the match has been obtained
+     * @param rootFile
+     */
     parseInputFilePath(regResult: RegExpExecArray, currentFile: string, rootFile: string): string | undefined {
```
tamuratak, line 35: - `regResult[0], regResult[1], regResult[2]` are less readable. We had better introduce a type representing the result. For example,
```
type A = {
   matchType: 'input' | 'child',
   path: string,
   ...
}
```

- We had better make the method `parseInputFilePath` private. `regResult` is too specific. I think we can make a new method in which we call `inputRegex.exec`, `childRegex.exec` and `parseInputFilePath`.

jlelong: I see your second point but this is not so obvious because of the `lastIndex` property of `RegExp` objects. We need to call `.exec` on copies of `inputReg` and `childReg` to properly use the `lastIndex` property. Hence, encapsulating the `.exec` and `parseInputFilePath ` calls  into a new method is not so easy all the more so as an `undefined` return of `.exec` is handled differently in `manager.ts` and `structure.ts`.

tamuratak: > We need to call .exec on copies of inputReg and childReg

Would you like to explain where we need copies? Are we already copying them?

jlelong: Yes, we are already copying them 

https://github.com/James-Yu/LaTeX-Workshop/blob/20e379c2fbdc6bf60a72df1c3d817bccb0a77b75/src/components/manager.ts#L531-L532

See [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex#description) for an explanation of how the `lastIndex` property works.

jlelong: I have refactored the way to apply child/input regex and how we process their matches

tamuratak: An assignment is not copying in JavaScript. `inputReg` and `this.pathUtils.inputRegex` refers to the same object.

jlelong: Good point! I was mislead by 

> If lastIndex is greater than the length of the string, `test()` and `exec()` fail, then lastIndex is set to 0. [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex#description)

and thought that it was making a shallow copy (as the type of `.latexIndex` is elementary, everything would have been fine).

We need to explicitly create new `RegExp`

https://github.com/James-Yu/LaTeX-Workshop/blob/5b46a70867a0c8ab5c2dc5d7cf2f0ef4bca4f39f/src/components/manager.ts#L533-L534

tamuratak: I don't understand why we always have to clone regexps yet.

However, if we always have to clone regexps, how about defining the following new class and extracts some methods in PathUtils to the class?
```ts
class A {
    private readonly inputRegex: RegExp
    private readonly childRegex: RegExp

        // initialize clone regexps
    }

    resetLastIndex() {

    }

    exec() {

    }

    parseInputFilePath() {

    }
}
```

jlelong: > I don't understand why we always have to clone regexps yet.

The whole thing is related to the use of the `g` flag  (see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex#description))

Consider the code

```js
regex = /myRegExp/g
while (true) {
    result = regex.exec(content)
    if (!result) {
        break
    }
}
```

Every time a match is found, the `lastIndex` property of is set to the next position following the most recent match. Note the value of `lastIndex`  is kept when there is no match and is not reset to `0`.

Let us go back to our code 

https://github.com/James-Yu/LaTeX-Workshop/blob/5b46a70867a0c8ab5c2dc5d7cf2f0ef4bca4f39f/src/components/manager.ts#L533-L539

We might get away with it by just setting 

```js
this.pathUtils.inputRegex.lastIndex = 0
this.pathUtils.childRegex.lastIndex = 0
while(true) {
    const result: MatchPath | undefined = this.pathUtils.exec(this.pathUtils.inputRegex, this.pathUtils.childRegex, content)
```

Yet, if the code is called inside two `async` functions, this may lead to concurrency issues on the `lastIndex` property. Moreover, as the regexp is modified by the `exec` call, I find it much cleaner to use a dedicated instance of the regexp for every `while` loop.

I hope this is clearer now. I agree that we could define a specific class to encapsulate all this and create a new instance of the class every time the regexps are cloned.

tamuratak: >  Note the value of lastIndex is kept when there is no match and is not reset to 0.

See `RegExp.prototype.exec()` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#return_value
> If the match fails, the exec() method returns null, and sets lastIndex to 0.

It says `lastIndex` is reset to `0`.

tamuratak: When I execute the following code:
```js
const regex = /a/g
console.log(regex.lastIndex)
while (true) {
    result = regex.exec('bbabbab')
    console.log(regex.lastIndex)
    if (!result) {
        break
    }
}
console.log(regex.lastIndex)
```

I obtain

```
$  node -v
v12.20.1
$  node t.js
0
3
6
0
0
```

jlelong: I get it! The reason why it does not work if we do not clone regexps is that the function `parseInputFiles` is recursive.

https://github.com/James-Yu/LaTeX-Workshop/blob/f23e77e59c1862043a1b8ece4b70ba706471ac3a/src/components/manager.ts#L575-L601

It calls `parseFileAndSubs`, which in turn calls `parseInputFiles`.  We need to make sure every call to `parseInputFiles` uses its own copy of the regex otherwise `lastIndex` is wrong when entering a recursive call or going up to an ancestor call. 

When looking at all this quickly, I first had the impression that `lastIndex` was not reset properly but it was indeed more subtle.

tamuratak: I see. I have understood the reason.

tamuratak (approved):

PR #2462, opened 2021-01-04
Update default settings

XuehaiPan:
1. Add biber to latex.tools, and "pdflatex ➞ biber ➞ pdflatex × 2" to latex.recipes.
2. CHANGE default latex.outDir to "%DIR%/out".

PR #2633, opened 2021-04-22
Add biber as tool and include outdir for pdflatex

mhSangar:
This PR includes:
- Adding `biber` as a default tool.
- Fixing `pdflatex` command, including support for output directory.

Biber, unlike `bibtext`, supports utf-8 characters and includes more entry types on the `.bib` file. On the other hand, the output directory configuration was not being taken into consideration for the pdflatex command.


tamuratak:
Duplicate of #2462

Issue #2637, opened 2021-04-24
Lag when pressing "Enter" key


Javinator9889:
### Disable all the other extensions except for LaTeX Workshop, restart VS Code, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

## Describe the bug [Required]

With the latest version published on Wednesday 21st, typing text is OK but when pressing Enter a *huge delay* happens. I visited the Wiki and noticed that some extensions can cause Enter delay. I'm currently using [`Code Spell Checker`](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker) and [`Bracket Pair Colorizer (v1)`](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker). I tried disabling (and uninstalling both) but the issue does not dissapear.

Until Wednesday I was able to just write the documents as always with all the extensions I have but now it's very difficult to just type as always.

### To Reproduce
1. Create a new LaTeX document.
2. Start typing as usual and, when perssing Enter, a delay should happen (not always but often).
3. In a blank document, just type as always and the Enter key delay will dissapear.

### Expected behavior

Pressing enter produces almost no delay.

## Logs [Required]

Please paste the whole log messages here, not parts of ones. The log should start with `Initializing LaTeX Workshop`. It is very important to identify problems.

### LaTeX Workshop Output [Required]

```
[10:32:08] Initializing LaTeX Workshop.
[10:32:08] Extension root: /home/javinator9889/.vscode/extensions/james-yu.latex-workshop-8.17.0
[...]

PR #2639, opened 2021-04-26
Fix cached packages list update

jlelong:
When updating the cached packages list, there is no check that a package is already in the list. So every time the  `updatePkg` function is called, the list gets doubled. This is probably related to the lag issue reported in #2637

### DIFF
--- a/src/providers/completer/command.ts
+++ b/src/providers/completer/command.ts
@@ -235,11 +235,6 @@ export class Command implements IProvider {
             this.updatePkgWithNodeArray(file, nodes)
         } else if (content !== undefined) {
             const pkgReg = /\\usepackage(?:\[[^[\]{}]*\])?{(.*)}/g
-            const pkgs: string[] = []
-
-            if (this.extension.manager.cachedContent[file].element.package === undefined) {
-                this.extension.manager.cachedContent[file].element.package = []
-            }
 
             while (true) {
                 const result = pkgReg.exec(content)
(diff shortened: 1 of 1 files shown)


src/providers/completer/command.ts
```diff
@@ -281,11 +281,9 @@ export class Command implements IProvider {
                             if (node.name === 'documentclass') {
                                 pkg = 'class-' + pkg
                             }
-                            const pkgs = this.extension.manager.cachedContent[file].element.package
-                            if (pkgs) {
+                            const pkgs = this.extension.manager.cachedContent[file].element.package as string[]
```
tamuratak, line 284: We should not use `as string[]`.  We should check the value even if we know it is not `undefined` now. In the future refactoring, it would be unintentionally `undefined`. And, type checks exist for such unintentional bugs.

tamuratak (changes requested):

src/providers/completer/command.ts
jlelong: I was almost sure you were going to say so...

tamuratak (approved):

src/providers/completer/command.ts
jlelong: The funny thing is that DeepScan detects the new test as an always true condition. Do you know if there is a way to tell DeepScan to ignore it?

src/providers/completer/command.ts
```diff
@@ -248,14 +243,16 @@ export class Command implements IProvider {
...
                         return
                     }
-                    const filePkgs = this.extension.manager.cachedContent[file].element.package
-                    if (filePkgs) {
+                    let filePkgs = this.extension.manager.cachedContent[file].element.package
+                    if (!filePkgs) {
+                        filePkgs = []
+                        this.extension.manager.cachedContent[file].element.package = filePkgs
+                    }
+                    if (filePkgs && !filePkgs.includes(pkg)) {
```
tamuratak, line 254: It is correct what DeepScan is saying. We can change `filePkgs && !filePkgs.includes(pkg)` to `!filePkgs.includes(pkg)`.

jlelong: Great!

tamuratak (approved):

PR #2641, opened 2021-04-27
sets: use whenever possible sets instead of arrays

Javinator9889:
Due to the lag problem commented at #2637 (and the simple solution proposed) new future bugs and performance issues can be approached using `Set`s.

Using those data structures allows having unique items inside them so when trying to add a duplicated one won't be an issue. In addition, most of the operations are order) such as insertions, deletions and searches. The main "problem" is that sets are unordered so then any kind of displaying won't have always the same distribution (an approach, if needed, would be converting the `Set` to an `Array` and then sortering, if necessary).

With this PR, some operations can be simplified by just using sets, allowing cleaner code alongside faster and efficient operations

### DIFF
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -38,7 +38,7 @@ interface Content {
             environment?: EnvEntry[],
             bibitem?: CiteEntry[],
             command?: CmdEntry[],
-            package?: string[]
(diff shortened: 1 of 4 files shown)


jlelong:
Merged in 7ac1fc39

PR #2654, opened 2021-05-01
Remove src/config.ts

tamuratak:
Remove `src/config.ts`.

We can use `deprecationMessage` and `markdownDeprecationMessage` for deprecated configurations. See https://code.visualstudio.com/api/references/contribution-points

https://github.com/microsoft/vscode/issues/91618#issuecomment-621361916


jlelong:
I will have a look at it but I need a bit more time to study the use cases of `src/config.ts`.

jlelong:
I have a mixed opinion on removing `src/config.ts`. 

I agree that `deprecationMessage` and `markdownDeprecationMessage` provide a smart way of dealing with old configuration variables but only while editing your settings. You do not get a warning message at startup when a deprecated configuration variable has the non standard value in your settings.

Yet, it seems that vscode' folks can live with it. The real benefit I can see from removing `src/config.ts` is that we would stop playing with the user's settings in his back if he has disabled warning popups.

tamuratak:
How about stopping updating users' configurations? I think updating is too much. It is enough to display warnings for deprecated configurations.
 
It is also not easy to use functions defined in `src/config.ts`. So, we had better define simpler methods in `src/components/configuration.ts`. It enables us to display more information in the warnings.

jlelong:
I agree that we should stop updating the user's configurations and only display warnings for deprecated settings.

I am fine with defining a simpler approach in `src/components/configuration.ts`. We could use `markdownDeprecationMessage` to display more information

tamuratak:
I have pushed a prototype. I have moved deprecated settings from `src/config.ts` to `src/components/configurationlib/deprecated.ts`. Should we still treat these settings? Or, should we start with an empty `DeprecatedConfiguration.deprecatedConfigurations`?

src/components/configurationlib/deprecated.ts
```diff
@@ -0,0 +1,121 @@
+import * as vscode from 'vscode'
+import type {Extension} from '../../main'
+
+type DeprecatedConfig = {
+    oldConfigKey: string,
+    newConfigKey?: string,
+    message?: string
```
jlelong, line 7: The `message` field is not used in the function `check`.

jlelong:
The last change to the list of deprecated settings goes back to July 2020. I think it is fair to start with a fresh empty list 
in `DeprecatedConfiguration.deprecatedConfigurations`

src/components/configurationlib/deprecated.ts
```diff
@@ -0,0 +1,121 @@
...
+    check() {
+        const configuration = vscode.workspace.getConfiguration()
+        for (const conf of this.deprecatedConfigurations) {
+            const hasConf = configuration.has(conf.oldConfigKey)
+            if (hasConf) {
+                let msg: string
+                if (conf.newConfigKey) {
+                    msg = `"${conf.oldConfigKey}" has been replaced by "${conf.newConfigKey}". Please manually remove the deprecated config from your settings.`
+                } else {
+                    msg = `"${conf.oldConfigKey}" has been deprecated. Please manually remove the deprecated config from your settings.`
```
jlelong, line 113: Use `conf.message` if it exists. It would allow us to display a bespoke message. This may be useful to warn the user against a change of accepted values for a configuration variable (see 49a55b4 for instance).

tamuratak:
Notice that `vscode.window.showInformationMessage` does not support the markdown format. See microsoft/vscode/issues/69214.

If we want to show users more information,  we can set `mode: true` in `MessageOptions`.
See https://code.visualstudio.com/api/references/vscode-api#MessageOptions. Then, when users click `OK`,  we call `workbench.action.openSettings` to open a new setting in the setting editor. See microsoft/vscode/issues/72352.

jlelong (approved):

PR #2656, opened 2021-05-04
Stop using fs-extra

tamuratak:
Stop using `fs-extra`.  We can use `fs.promises.readFile` instead of `fs-extra.readFile`.

See 

- https://nodejs.org/docs/latest-v12.x/api/fs.html#fs_fs_promises_api
- https://nodejs.org/docs/latest-v12.x/api/fs.html#fs_promise_example

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -487,15 +487,6 @@
         "@types/node": "*"
       }
     },
-    "@types/fs-extra": {
-      "version": "9.0.11",
-      "resolved": "https://registry.npmjs.org/@types/fs-extra/-/fs-extra-9.0.11.tgz",
-      "integrity": "sha512-mZsifGG4QeQ7hlkhO56u7zt/ycBgGxSVsFI/6lGTU34VtwkiqrrSDgw0+ygs8kFGWcXnFQWMrzF2h7TtDFNixA==",
-      "dev": true,
-      "requires": {
-        "@types/node": "*"
-      }
-    },
     "@types/glob": {
       "version": "7.1.3",
       "resolved": "https://registry.npmjs.org/@types/glob/-/glob-7.1.3.tgz",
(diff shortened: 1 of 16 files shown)


tamuratak:
@jlelong  I have a question. Why do we call `fs.statSync(fullOutDir).isDirectory()`?

https://github.com/James-Yu/LaTeX-Workshop/blob/1ad79ba072271f17134a99099ec55e9a710b8e9e/src/components/builder.ts#L223

jlelong:
If `fullOutDir` exists as a path but is not a directory, the build will fail but this is not handled properly. By using

https://github.com/James-Yu/LaTeX-Workshop/blob/1858bfea42d1811b88dc6399255b5e9f4fa0e3e8/src/components/builder.ts#L223-L225

we make the error clearer and happen earlier. I agree that this should be refactored into a more easily understandable code.

jlelong:
This is not directly related to the PR, but yet I think we should refactor and clarify the function `cleanGlob`, which is very sensitive

https://github.com/James-Yu/LaTeX-Workshop/blob/4e9d49a9e22f7c58da49ea973b6c1e56954a67d2/src/components/cleaner.ts#L40-L74

- Before calling `unlink` or `rmdir`, we should check the type of `file`(true file or directory). Indeed, why should we remove directories here? Removing directories could create a major security issue.
- We should make clear in the description of `latex.clean.subfolder.enabled` that intermediate sub-folders are not removed

    https://github.com/James-Yu/LaTeX-Workshop/blob/91cdbf04ffa42ef75ff01c8b70c6ba37e3f2f5fe/package.json#L1148-L1152

tamuratak:
> If fullOutDir exists as a path but is not a directory, the build will fail but this is not handled properly.

I see. Can I merge this PR?

jlelong (approved):

jlelong:
> Can I merge this PR?

Yes. Once done, we should take care of https://github.com/James-Yu/LaTeX-Workshop/pull/2656#issuecomment-831756053

PR #2667, opened 2021-05-07
Refactor cleaner

jlelong:
Only remove files, not directory. Related to
https://github.com/James-Yu/LaTeX-Workshop/pull/2656#issuecomment-831756053


src/components/cleaner.ts
```diff
@@ -55,14 +55,19 @@ export class Cleaner {
...
-                // If unlinking fails, replace it with an rmdir Promise
-            }, () => fs.promises.rmdir(file).then(() => {
-                this.extension.logger.addLogMessage(`Folder removed: ${file}`)
-            }, () => {
-                this.extension.logger.addLogMessage(`Error removing file: ${file}`)
-            })))
+            files.map(file => {
+                fs.promises.stat(file).then((stats: fs.Stats) => {
+                    if (stats.isFile()) {
+                        fs.promises.unlink(file).then(() => {
```
tamuratak, line 61: We had better use `await` and `async` instead of `then` for readability.

tamuratak (changes requested):

src/components/cleaner.ts
jlelong: I fully agree that chaining promises makes the function `cleanGlob` very difficult to understand and therefore maintain.

src/components/cleaner.ts
```diff
@@ -55,15 +55,20 @@ export class Cleaner {
...
+                const stats: fs.Stats = fs.statSync(file)
+                if (stats.isFile()) {
+                        fs.promises.unlink(file).then(() => {
+                            this.extension.logger.addLogMessage(`File cleaned: ${file}`)
+                        }, () => {
+                            this.extension.logger.addLogMessage(`Error removing file: ${file}`)
+                        })
+                    } else {
+                       this.extension.logger.addLogMessage(`Not removing non-file: ${file}`)
+                    }
```
tamuratak, line 68: Indentation is incorrect.

tamuratak (changes requested):

src/components/cleaner.ts
```diff
@@ -55,16 +55,20 @@ export class Cleaner {
...
+                if (stats.isFile()) {
+                    fs.promises.unlink(file).then(() => {
+                        this.extension.logger.addLogMessage(`File cleaned: ${file}`)
+                    }, () => {
+                        this.extension.logger.addLogMessage(`Error removing file: ${file}`)
+                    })
+                } else {
+                    this.extension.logger.addLogMessage(`Not removing non-file: ${file}`)
+                }
+            })
```
tamuratak, line 69: The current code is not correct. In the original code, `files.map(file => fs.promises.unlink(file).then(...))` returns `Promise<void>[]`. Then, `Promise.all(promises: Promise<void>[])` is called.

https://github.com/James-Yu/LaTeX-Workshop/blob/7467c884f116fffa11e2e60d427607dee46217ab/src/components/cleaner.ts#L56-L65

However, in the current code,  `files.map(file => {...})` returns `void[]`. 

I think we should rewrite the whole code of `cleanGlob`.

tamuratak (changes requested):

src/components/cleaner.ts
```diff
@@ -45,48 +45,32 @@ export class Cleaner {
...
+                            this.extension.logger.addLogMessage(err.message)
+                        }
+                    })
                 } else {
-                    resolve(files)
+                    this.extension.logger.addLogMessage(`Not removing non-file: ${file}`)
                 }
-            })
+            }
+            resolve()
```
tamuratak, line 69: The code is not correct. `fs.unlink` asynchronously removes a file. When `resolve()` called, the removal might not finish yet. 

- https://nodejs.org/docs/latest-v14.x/api/fs.html#fs_fs_unlink_path_callback

We can use `await`.

```ts
    private async cleanGlob(rootFile: string): Promise<void> {
        let globs = configuration.get('latex.clean.fileTypes') as string[]
        const outdir = path.resolve(path.dirname(rootFile), this.extension.manager.getOutDir(rootFile))
        if (configuration.get('latex.clean.subfolder.enabled') as boolean) {
            globs = globs.map(globType => './**/' + globType)
        }
        
        const files = globs.map(g => glob.sync(g, {cwd: outdir}))
                           // Reduce the array of arrays to a single array containing all the files that should be deleted
                           .reduce((all, curr) => all.concat(curr), [])
                           // Resolve the absolute filepath for every file
                           .map(file => path.resolve(outdir, file))
        for (const file of files) {
            const stats: fs.Stats = fs.statSync(file)
            if (stats.isFile()) {
                try {
                    await fs.promises.unlink(file)
                } catch (err) {
                    this.extension.logger.addLogMessage(`Error while cleaning: ${file}`)
                    this.extension.logger.addLogMessage(err.message)
                    continue
                }
                this.extension.logger.addLogMessage(`Cleaning file: ${file}`)
            } else {
                this.extension.logger.addLogMessage(`Not removing non-file: ${file}`)
            }
        }
    }
```

tamuratak (changes requested):

tamuratak (changes requested):

tamuratak (approved):

PR #2673, opened 2021-05-10
2671-strip-verbatim

### LINKED ISSUE #2671: LaTeX Workshop doesn't ignore verbatim in external files
I have a main file that `\include`s some chapters from another folder. When I save these chapter files with just some text, everything works well and `main.tex` is recompiled. However, if I have a verbatim in these chapters with the `\begin{document}` content, LaTeX Workshop thinks that it's a real file and tries to compile, generating many aux files and an annoying warning.

## Minimum, Complete and Verifiable Example: 

Suppose the following directory structure:
```plain
.
├─ chapters
│   └─ introduction.tex
└─ main.tex
```

`main.tex` has:
```latex
\documentclass{article}

\begin{document}
    \include{./chapters/introduction.tex}
\end{document}
```

And `chapters/introduction.tex` has:
```latex
\section{Introduction}

This is a minimum \LaTeX\ document:

\begin{verbatim}
    \documentclass{article}

    \begin{document}
        Hello World!
    \end{document}
\end{verbatim}
```

If you save `chapters/introduction.tex` in VSCode, LaTeX Workshop will trigger the compilation recipe, generate a bunch of files and error out because, obviously, it's not a valid LaTeX file.

jlelong:
Close #2671

- Make sure to remove any verbatim parts before scanning the `.tex` content. For the moment, it breaks if there are nested verbatim environments.
- Make the list of verbatim-like environments configurable.

Once this PR is merged, I will investigate how the `content` property can be removed from `manager. cachedContent`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1000,6 +1000,14 @@
           "default": [],
           "markdownDescription": "List of directories where to look for extra input `.tex` files. \nAbsolute paths are required. You may also need to set the environment variable `TEXINPUTS` properly for the LaTeX compiler to find the `.tex` files, see the `env` parameter of [recipes](https://github.com/James-Yu/LaTeX-Workshop/wiki/Compile#latex-recipes)."
         },
+        "latex-workshop.latex.verbatimEnvs": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "default": ["verbatim", "lstlisting", "minted"],
+          "markdownDescription": "List environments with verbatim-like content. These environments are stripped off the `.tex` files before any parsing occurs. Note that this variable has no effect on syntax highlighting."
+        },
         "latex-workshop.kpsewhich.path": {
           "type": "string",
           "default": "kpsewhich",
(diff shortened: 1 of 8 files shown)


tamuratak:
We should stop adding regular expressions for ad hoc fixes. There are lots of corner cases. We had better use `Extension.pegParser.parseLatex` to find `\begin{document}`.

jlelong:
> We should stop adding regular expressions for ad hoc fixes. There are lots of corner cases. We had better use `Extension.pegParser.parseLatex` to find `\begin{document}`.

Ideally yes, but a lot of features rely on parsing the `.tex` content: hove preview, structure provider, find root. For these to work, we must make sure, all verbatim content is removed. It is not only about finding `\begin{document}`. The [`getTeXChildren`](https://github.com/James-Yu/LaTeX-Workshop/blob/9d1ca32bf2f852de3b30a0514e21cd3ec4a406ac/src/components/manager.ts#L597) uses the `.tex` content to find included files. 

Moreover, regex are used as a fallback in case an error occurs in `Extension.pegParser.parseLatex`. A regex fallback is indeed important to handle weave files as the `Extension.pegParser.parseLatex` cannot handle them.

I suggest to merge this PR as it will fix #2671 and to handle improvements in a new PR.

PR #2680, opened 2021-05-12
Make Manager.getCachedContent strict.

tamuratak:
Make `Manager.getCachedContent` strict.


jlelong:
Indeed it seems to me that this PR is mandatory as prior to it, calling `this.getCachedContent(file).element` may result in an _Uncaught TypeError: Cannot read property 'element' of undefined_. 

I suggest to make the `if` clauses more homogeneous (even if they are semantically equivalent) between

```
const cache = this.extension.manager.getCachedContent(file)
if (!cache)
```

```
const cache = this.extension.manager.getCachedContent(file)
if (cache === undefined)
```

jlelong (approved):

PR #2682, opened 2021-05-12
Adopt a more lazy approach to read tex content

jlelong:
Following https://github.com/James-Yu/LaTeX-Workshop/issues/2671#issuecomment-835562749, I have had a closer look at how `cachedContent[file].content` is set/used. 

- If `file` is opened in a vscode `editor`, the content is read using `TextDocument.getText`, which uses an internal cache mechanism https://github.com/Microsoft/vscode/blob/166efae36645bc787afc8bbfb12d8ac5adda8ead/src/vs/editor/common/model/mirrorTextModel.ts#L58-L80

     The cached gets invalidated when a change event is fired. Hence, calling `getText` repeatedly is harmless as long as the editor (buffer) is not modified.
- If the file is not opened inside vscode, the file content is read via `fs.readFile`. I doubt that this function uses an internal cache and we call it several times in a row with no reason. For instance, https://github.com/James-Yu/LaTeX-Workshop/blob/de2bd02bc45d4b2c7615fcc8e13a12bac57632bc/src/components/manager.ts#L776-L784

    - In `this.parseFileAndSubs`, we call `this.getDirtyContent(file, true)`, which reads the content from disk
    - The following line,`this.updateCompleterOnChange(file)` calls `fs.promises.readFile(file)` again. https://github.com/James-Yu/LaTeX-Workshop/blob/de2bd02bc45d4b2c7615fcc8e13a12bac57632bc/src/components/manager.ts#L841-L844 and `this.extension.completer.input.getGraphicsPath` reads the file from disk once more https://github.com/James-Yu/LaTeX-Workshop/blob/71f66741ce3bcd9c52a0a44e8d69d0146951bf2d/src/providers/completer/input.ts#L28-L29
 

This PR  implements a lazy update of `cachedContent[file].content` (or at least lazier than it used to be) and uses `getDirtyContent` to access a `tex` buffer/file content instead of directly calling `fs.readFile`.

### DIFF
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -846,7 +846,7 @@ export class Manager {
 
     // This function updates all completers upon tex-file changes.
     private updateCompleterOnChange(file: string) {
-        fs.promises.readFile(file).then(buffer => buffer.toString()).then(content => this.updateCompleter(file, content))
+        this.updateCompleter(file, this.getDirtyContent(file))
         this.extension.completer.input.getGraphicsPath(file)
     }
 
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -495,12 +495,16 @@ export class Manager {
     /**
      * Get the buffer content of a file if it is opened in vscode. Otherwise, read the file from disk
      */
-    getDirtyContent(file: string): string {
+    getDirtyContent(file: string): string | undefined {
         const cache = this.cachedContent[file]
         if (cache !== undefined) {
             if (cache.content) {
                 return cache.content
             }
+        }
+        if (!fs.existsSync(file)) {
+            this.extension.logger.addLogMessage(`Cannot read dirty content of unknown ${file}`)
+            return undefined
         }
         const fileContent = fs.readFileSync(file).toString()
         this.setCachedContent(file, {content: fileContent, element: {}, children: [], bibs: []})
@@ -541,7 +545,11 @@ export class Manager {
             this.fileWatcher.add(file)
             this.filesWatched.push(file)
         }
-        const content = utils.stripCommentsAndVerbatim(this.getDirtyContent(file))
+        let content = this.getDirtyContent(file)
+        if (!content) {
+            return
+        }
+        content = utils.stripCommentsAndVerbatim(content)
         this.cachedContent[file].children = []
         this.cachedContent[file].bibs = []
         this.parseInputFiles(content, file, baseFile)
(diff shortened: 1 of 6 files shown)


src/main.ts
```diff
@@ -227,20 +230,18 @@ export function activate(context: vscode.ExtensionContext): ReturnType<typeof ge
...
-        if (cache !== undefined) {
-            cache.content = content
-        }
         if (configuration.get('intellisense.update.aggressive.enabled')) {
             if (updateCompleter) {
                 clearTimeout(updateCompleter)
             }
             updateCompleter = setTimeout(() => {
+                const content = e.document.getText()
+                cache.content = content
```
tamuratak, line 244: We had better make a method to update the `content` property in the `Manager` class and call it here. The method could take a `TextDocument` object as the argument.

```ts
updateAaaaaa(document: vscode.TextDocument) {
...
}
```

tamuratak (changes requested):

src/components/manager.ts
```diff
@@ -478,17 +492,21 @@ export class Manager {
...
+        const cache = this.cachedContent[file]
+        if (cache !== undefined) {
+            if (cache.content) {
+                return cache.content
             }
-            return this.cachedContent[cachedFile].content
         }
-        const fileContent = utils.stripCommentsAndVerbatim(fs.readFileSync(file).toString())
+        if (!fs.existsSync(file)) {
+            this.extension.logger.showErrorMessage(`Cannot read dirty content of ${file}`)
```
tamuratak, line 506: We should call `this.extension.logger.addLogMessage` to log errors too.

tamuratak (changes requested):

src/components/manager.ts
```diff
@@ -478,17 +485,17 @@ export class Manager {
...
+    getDirtyContent(file: string): string {
+        const cache = this.cachedContent[file]
+        if (cache !== undefined) {
+            if (cache.content) {
+                return cache.content
             }
-            return this.cachedContent[cachedFile].content
         }
-        const fileContent = utils.stripCommentsAndVerbatim(fs.readFileSync(file).toString())
+        const fileContent = fs.readFileSync(file).toString()
```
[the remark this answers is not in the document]
jlelong: Done for checking if file exists on disk before trying to read it. I am not sure we should immediately return  if  the file is one of the keys of `Manager.cachedContent`. I prefer to just add a new entry in that case.

src/components/manager.ts
jlelong: Do we really need to show a popup error with `showErrorMessage`? Maybe logging the error is enough.

tamuratak: I agree that logging the error is enough.

tamuratak (approved):

PR #2694, opened 2021-05-19
Use Map to store envs.

tamuratak:
Use Map to store envs.

### DIFF
--- a/src/providers/completer/environment.ts
+++ b/src/providers/completer/environment.ts
@@ -160,11 +160,11 @@ export class Environment implements IProvider {
         // Load environments from the package if not already done
         if (!this.packageEnvsAsCommand.has(pkg)) {
             const entry: Suggestion[] = []
-            this.packageEnvsAsCommand.set(pkg, entry)
             const envs: {[key: string]: EnvItemEntry} = this.getEnvItemsFromPkg(pkg)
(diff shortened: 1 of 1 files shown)


jlelong:
Why not changing the return type of `getEnvItemsFromPkg` to `Map<string, Suggestion[]>`?

src/providers/completer/environment.ts
```diff
@@ -158,21 +158,23 @@ export class Environment implements IProvider {
         }
 
         // Load environments from the package if not already done
-        if (!(pkg in this.packageEnvsAsCommand)) {
-            this.packageEnvsAsCommand[pkg] = []
+        if (!this.packageEnvsAsCommand.has(pkg)) {
+            const entry: Suggestion[] = []
+            this.packageEnvsAsCommand.set(pkg, entry)
```
jlelong, line 163: Same comment as in the other PR. I would move this instruction to line 168. It is more a matter of taste than anything else actually.

tamuratak:
> Why not changing the return type of getEnvItemsFromPkg to Map<string, Suggestion[]>?

We don't have to do that. Using an object returned by `JSON.parse` is safe. Set the value of the object to another object can be not safe. 

The `getEnvItemsFromPkg` method is also private.

jlelong (approved):

PR #2695, opened 2021-05-19
Use Map to store citations.

tamuratak:
Use Map to store citations.

### DIFF
--- a/src/providers/completer/citation.ts
+++ b/src/providers/completer/citation.ts
@@ -5,9 +5,34 @@ import {bibtexParser} from 'latex-utensils'
 import type {Extension} from '../../main'
 import type {IProvider} from './interface'
 
+
+class Fields extends Map<string, string> {
+
+    get author() {
+        return this.get('aurthor')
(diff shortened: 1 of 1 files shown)


src/providers/completer/citation.ts
```diff
@@ -184,12 +212,12 @@ export class Citation implements IProvider {
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
         if (fs.statSync(file).size >= (configuration.get('intellisense.citation.maxfilesizeMB') as number) * 1024 * 1024) {
             this.extension.logger.addLogMessage(`Bib file is too large, ignoring it: ${file}`)
-            if (file in this.bibEntries) {
-                delete this.bibEntries[file]
+            if (this.bibEntries.has(file)) {
+                this.bibEntries.delete(file)
```
jlelong, line 216: The `.delete` method can be safely called without checking if the key is in the `Map`.

jlelong (approved):

PR #2697, opened 2021-05-20
Use Map for Reference.prevIndexObj

tamuratak:
Use Map for `Reference.prevIndexObj`

### DIFF
--- a/src/providers/completer/reference.ts
+++ b/src/providers/completer/reference.ts
@@ -234,7 +234,7 @@ export class Reference implements IProvider {
             if (result === null) {
                 break
             }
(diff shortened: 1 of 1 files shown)


jlelong (changes requested):
In https://github.com/James-Yu/LaTeX-Workshop/blob/4c6c9589c0ee80b45832c47c6ba36b58d7f5c4ca/src/providers/completer/reference.ts#L237 we have to use

```js
this.prevIndexObj.has(result[1].replace('@cref', ''))
```

jlelong (approved):

PR #2703, opened 2021-05-20
Change Citation.getEntryDict to Citation.getEntry

tamuratak:
Change `Citation.getEntryDict` to `Citation.getEntry`. Avoid using object to store citations.

### DIFF
--- a/src/providers/completer/citation.ts
+++ b/src/providers/completer/citation.ts
@@ -120,11 +120,10 @@ export class Citation implements IProvider {
         })
     }
 
-    getEntryDict(): {[key: string]: Suggestion} {
+    getEntry(key: string): Suggestion | undefined {
         const suggestions = this.updateAll()
-        const entries: {[key: string]: Suggestion} = {}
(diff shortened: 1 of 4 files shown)


src/providers/completer/citation.ts
```diff
@@ -120,11 +120,10 @@ export class Citation implements IProvider {
         })
     }
 
-    getEntryDict(): {[key: string]: Suggestion} {
+    getEntry(key: string): Suggestion | undefined {
         const suggestions = this.updateAll()
-        const entries: {[key: string]: Suggestion} = {}
-        suggestions.forEach(entry => entries[entry.key] = entry)
-        return entries
+        const entry = suggestions.find((elm) => elm.key === key)
```
jlelong, line 125: The function `find` returns the first element matching the condition whereas in the previous implementation we actually returned the last one if there were multiple matches. I am not if it can happen or even it matters.

tamuratak: I think the change is acceptable.

jlelong (approved):

PR #2705, opened 2021-05-21
Stop using Object.assign.

tamuratak:
Stop using Object.assign.

### DIFF
--- a/src/providers/completer/glossary.ts
+++ b/src/providers/completer/glossary.ts
@@ -39,7 +39,7 @@ export class Glossary implements IProvider {
         if (result[1] && result[1].match(/^ac/i)) {
             suggestions = this.acronyms
         } else {
-            suggestions = new Map( [...this.acronyms.entries(), ...this.glossaries.entries()] )
+            suggestions = new Map( [...this.acronyms, ...this.glossaries] )
         }
 
         // Compile the suggestion object to array
--- a/src/providers/completer/glossary.ts
+++ b/src/providers/completer/glossary.ts
@@ -43,17 +43,7 @@ export class Glossary implements IProvider {
         }
 
         // Compile the suggestion object to array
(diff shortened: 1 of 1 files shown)


jlelong:
We should also replace https://github.com/James-Yu/LaTeX-Workshop/blob/dfaba4137c555aa785ab6f2644f25dae80cafdbe/src/providers/completer/glossary.ts#L46-L47 by

```js
let keys = Array.from(new Set(suggestions.keys())
```

as `Object.keys` does not return the keys of the` Map`.

src/providers/completer/glossary.ts
```diff
@@ -39,7 +39,7 @@ export class Glossary implements IProvider {
         if (result[1] && result[1].match(/^ac/i)) {
             suggestions = this.acronyms
         } else {
-            suggestions = Object.assign({}, this.acronyms, this.glossaries)
+            suggestions = new Map( [...this.acronyms.entries(), ...this.glossaries.entries()] )
```
jlelong, line 42: According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#cloning_and_merging_maps, we can directly write

```js
suggestions = new Map( [...this.acronyms, ...this.glossaries] )
```

tamuratak:
It seems that we can change the whole part

https://github.com/James-Yu/LaTeX-Workshop/blob/365e728b68f050e967c78c4425e1dd974fc49759/src/providers/completer/glossary.ts#L46-L57

to 

```ts
  const items = Array.from(suggestions.values())
  return items
```

Is this correct?

jlelong:
Yes, I agree

tamuratak:
Can I merge this PR?

jlelong (approved):

jlelong:
Yes

PR #2728, opened 2021-05-29
Use await in Manager.parseFlsFile

tamuratak:
We should await operations for each file with `for` loops. We cannot await them with `forEach`.


tamuratak:
This PR needs some tweaks.

PR #2741, opened 2021-06-07
Put fs.readFile in try/catch

jlelong:
Calling fs.exists prior to fs.readFile is a bad idea, see
https://nodejs.org/api/fs.html#fs_fs_exists_path_callback


tamuratak (commented):
I think it would be better to define a method at `src/components/lwfs.ts` as follows and use it.

```ts
export class LwFileSystem {
    readFileSyncGracefully(filepath: string) {
        try {
            const ret = fs.readFileSync(filepath)
            return ret
        } catch (err) {
            ...
            return undefined
        }
    }
}
```

tamuratak (approved):

PR #2756, opened 2021-06-12
Use Custom Editor as a hook to start the internal PDF viewer.

tamuratak:
Use Custom Editor as a hook to start the internal PDF viewer.

Since we cannot open muptiple cutom editors for the same PDF file in the same editor group,
we don't directly use Custom Editor for the viewer.

- https://code.visualstudio.com/api/extension-guides/custom-editors
- https://github.com/microsoft/vscode/issues/41289

https://user-images.githubusercontent.com/10665499/121760157-459cbc80-cb64-11eb-9a15-5e99ec343eb4.mp4

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     "onLanguage:doctex",
     "onLanguage:jlweave",
     "onLanguage:rsweave",
-    "onLanguage:pdf",
     "onLanguage:bibtex",
     "onCommand:latex-workshop.build",
     "onCommand:latex-workshop.recipes",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -2068,7 +2058,19 @@
...
       ]
-    }
+    },
+    "customEditors": [
+      {
+        "viewType": "latex-workshop-pdf-hook",
+        "displayName": "LaTeX Workshop Internal PDF Viewer",
+        "selector": [
+          {
+            "filenamePattern": "*.pdf"
```
jlelong, line 2068: Maybe we could add `*.PDF` too.

jlelong (approved):

PR #2759, opened 2021-06-13
Drop support for versions prior to VS Code 1.53.2

tamuratak:
Drop support for versions prior to VS Code 1.53.2


jlelong:
I think that is fair. VS Code `1.52` goes back to November 2020.

Issue #2763, opened 2021-06-14
How to change shortcut-bindings?


1900011604:
I am used to [TexStudio](https://github.com/texstudio-org/texstudio)'s shortcut `Ctrl+Alt+F` for `\frac{}{}`, is there any way of editing the shortcut list?

jlelong:
The `\frac{}{}` snippet is defined in https://github.com/James-Yu/LaTeX-Workshop/blob/c1900b2dde52cf6e7c7eeb3987fa8e1af10084a9/snippets/latex.json#L37-L41

You can associate a keybinding to it as explained at https://code.visualstudio.com/updates/v1_9#_insert-snippets

PR #2773, opened 2021-06-18
Add keyboard shortcuts for frac and partial

1900011604:
Two small changes, maybe you could change them to `ctrl+m ctrl+f` and `ctrl+m ctrl+p`

### DIFF
--- a/package.json
+++ b/package.json
@@ -683,6 +683,18 @@
         "command": "latex-workshop.shortcut.mathcal"
       },
       {
+        "key": "shift+alt+f",
+        "mac": "shift+alt+f",
+        "when": "editorTextFocus && !editorReadonly && editorLangId =~ /latex|rsweave|jlweave/",
+        "command": "latex-workshop.shortcut.frac"
+      },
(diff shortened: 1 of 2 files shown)


1900011604:
I will give a better pull-request in July, adding support for `\sqrt` and maybe some other snippets.

jlelong:
Keybindings are very personal. We do not want to define new ones inside the extensions. So please, define them in your own settings as explained #2763.

1900011604:
`vscode.commands.registerCommand('latex-workshop.shortcut.frac', () => extension.commander.toggleSelectedKeyword('frac'))`

I cannot define them as you did not provide the command like `latex-workshop.shortcut.frac`

jlelong:
You do not need the `latex-workshop.shortcut.frac` function. In your `keybindings.json` file, add

```json
{
  "key": "cmd+m ctrl+f'",
  "command": "editor.action.insertSnippet",
  "args": { "name": "fraction" }
}
```

PR #2783, opened 2021-06-29
Add support for glossary external file

### LINKED ISSUE #2780: Intellisense for glossary in external file
Hello everyone!
I'm using Latex-Workshop for an extensive writing project and make heavy use of acronym entries in a glossary. I would like to use the completion functionality of IntelliSense for this. While going through the issues I found #516 and pull request #2298 which adds glossary completion. However, it did not work for me. After reducing the files to an MWE I figured out that it is due to putting the glossary entries in a seperate file.

`main.tex`:
```tex
\documentclass{article}
\usepackage{lipsum}  
\usepackage{glossaries}
\makeglossaries
\loadglsentries{glossary}

\begin{document}
\lipsum[2-4]
\gls{rf} \gls{rf} \gls{rf}
\end{document}
```
`glossary.tex`:
```tex
\newacronym{rf}{RF}{radio-frequency}
\newacronym{te}{TM}{Transverse Magnetic}
```
When entering `\gls{` no suggestion is showing up in the main file.

As it is probably similar to the completion for bib entries shouldn't that be possible somehow? I couldn't find any info in the wiki. Compiling is fine and it works as desired.

I'm sorry that I don't stick to the bug report template for issues. But I have the impression most questions are not suitable for the problem. Maybe it's more like a feature request? Sorry if that's the case. As pointed out at the end of #2298 user feedback is welcome, so I thought it would be useful.

 - OS: Windows 10
 - VS Code version: 1.57.1
 - Extension version: 8.19.2
 - TeX distribution version: MikTex 21.6

VSCodium? No
Snap or Flatpack versions of VS Code? No
Are you using LaTeX Workshop with VS Code Remote? No

Thanks a lot for this great extension!

jlelong:
This PR implements 

- the scanning of glossary entries from an external file
- filename auto-completion for the command `\loadglsentries`.

It also fixes discovery of the tex children using the `.fls` approach.

Close #2780


tamuratak (commented):
It is unclear we can avoid infinite loops in `parseFlsFile`. We had better write in comments how we avoid infinite loops in `parseFlsFile`. 

https://github.com/James-Yu/LaTeX-Workshop/blob/6c5c3e590723cdeae378eccf727a4d4b20f28d57/src/components/manager.ts#L694

jlelong:
I have looked into this further and found that we can use the same condition as in `parseInputFiles`.
It is enough to ignore files which are already watched;

tamuratak (commented):
I am not sure whether we can avoid infinite loops.  We call `this.parseFileAndSubs(inputFile, texFile)` without `this.addToFileWatcher(inputFile)`. Can we guarantee that we have already called `this.addToFileWatcher(inputFile)`?

https://github.com/James-Yu/LaTeX-Workshop/blob/2b28d48e5b60a59f8cd845a2717578dfb4a7fbea/src/components/manager.ts#L711-L725

jlelong:
An infinite loop cannot happen in `parseFlsFile` because there is a single `.fls` file per LaTeX project and hence `parseFlsFile` is called only once per project parsing. The `.fls` file is generated by the compiler itself and named after the root file. Moreover, `this.pathUtils.parseFlsContent` returns sets to avoid parsing a file twice. Hence, the condition 

```ts
if (this.filesWatched.has(inputFile)) {
    /* We already watch this file, no need to enforce a new parsing */
    continue
}
```

is not really needed actually. We just need to take care of `texFile`, which is listed as an `INPUT` file.

> We call `this.parseFileAndSubs(inputFile, texFile)` without calling `this.addToFileWatcher(inputFile)`

This is `this.parseFileAndSubs(inputFile, texFile)` which calls `this.addToFileWatcher(inputFile)`.

https://github.com/James-Yu/LaTeX-Workshop/blob/2b28d48e5b60a59f8cd845a2717578dfb4a7fbea/src/components/manager.ts#L551-L563

If a file is not already watched, we add it to the file watcher.

Infinite loops may only happen in `parseInputFiles`. We make sure it does not happen with 

https://github.com/James-Yu/LaTeX-Workshop/blob/2b28d48e5b60a59f8cd845a2717578dfb4a7fbea/src/components/manager.ts#L657-L661

In the end, I do not think we need to introduce a set of parsed files and certainly not in `parseFlsFile`. It could have been an interesting approach for `parseInputFiles` but relying on the set on watched files has worked fine so far and this PR does not deal with `parseInputFiles`.

tamuratak:
> An infinite loop cannot happen in parseFlsFile because there is a single .fls file per LaTeX project 

We should not rely on the behavior of external programs to avoid infinite loops. Even if there are multiple `.fls` files, e.g. multiple documents are in the same directory, we should avoid infinite loops.  Even if the `.fls` file is corrupted or malicious, we should be able to avoid infinite loops. If we can not avoid them, it means that LaTeX Workshop is vulnerable.

>  parseFlsFile is called only once per project parsing. 

`parseFlsFile` calls `parseFileAndSubs`. And `parseFileAndSubs` calls `parseFlsFile`. Can `parseFlsFile` be called twice?

jlelong:
> `parseFlsFile` calls `parseFileAndSubs`. And `parseFileAndSubs` calls `parseFlsFile`. Can `parseFlsFile` be called twice?

Yes it can. Here is the calling sequence

1. `parseFlsFile` call `parseFileAndSubs` for every file that is not already watched.
2. In `parseFileAndSubs`,
    - if the file is not already watched, watch it. 
    - Call `parseFlsFile`
3. Back in `parseFlsFile`, we drop every file that is already watched, which takes care of the infinite loop issue.

jlelong:
Looking at the whole logic for parsing files once the root file has been found, I think we could simplify it and avoid calling `parseFlsFile` from `parseFileAndSubs`. Note that `parseFileAndSubs` is called once the root file has been found. Hence, we are only interested in its associated `.fls` file. So, it is sufficient to add

```ts
await this.parseFlsFile(this.rootFile)
```

after https://github.com/James-Yu/LaTeX-Workshop/blob/15109bde7a41151d8b861a2310cc09fa1b209192/src/components/manager.ts#L344 

and 

https://github.com/James-Yu/LaTeX-Workshop/blob/ead725e184971a5b58f13a06fc998aa012c70170/src/main.ts#L236

src/components/manager.ts
```diff
@@ -708,8 +708,8 @@ export class Manager {
                 !fs.existsSync(inputFile)) {
                 continue
             }
-            // Drop the current rootFile often listed as INPUT and drop any file that is already in the texFileTree
-            if (texFile === inputFile || inputFile in this.cachedContent) {
+            if (inputFile === texFile && this.filesWatched.has(inputFile)) {
```
tamuratak, line 711: Why is `&&` used instead of `||` in the condition?

jlelong: Because it is an error. Thanks a lot for catching this one!

tamuratak:
Can we immediately return here if `this.filesWatched.has(file)` is true?

https://github.com/James-Yu/LaTeX-Workshop/blob/1c43480d0a1e55460317a6a4398a7dd57e637ee9/src/components/manager.ts#L563-L566

jlelong:
No, because the function `parseFileAndSubs` is also called when a watched file changes

https://github.com/James-Yu/LaTeX-Workshop/blob/1c43480d0a1e55460317a6a4398a7dd57e637ee9/src/components/manager.ts#L825-L833

It is responsible for updating the cached content and the TeX children.

tamuratak:
Things are too complicated. In some case, we use `filesWatched` to stop parsing. In another place, we don't. Please write it in comments explicitly.

tamuratak:
I don't like using `filesWatched` to avoid infinite loops.

jlelong:
Here is the parsing logic (a new file is a file that is not already watched)

manager-logic

I will push this scheme to the repo.

- Infinite loops can only occur in `ParseFileAndSubs` but not in `ParseFls` anymore.
- You can see that indeed testing if a file is already watched is the way to prevent infinite loops.
- Once a file is watched, it is handled by `OnWatchedFileChanged`, so we should not take care of it in the cases _after successful build_ and _after a new root has been found_.

I hope things are clearer now.

tamuratak:
What I am talking about is the maintenance burden in the future.

The diagram does not ensure the actual codes are correct in the future. Because we keep adding changes, they can differ from the diagram. When we add a new change, another loop could appear. It increases the maintenance burden to checking whether the actual code matches the diagram every time.

We cannot use the type system of TypeScript to avoid this kind of infinite loop. We had better add tests to check whether we can prevent them. But we don't.

At least, we must write in comments how we avoid infinite loops. Otherwise, because we easily forget about everything, we will add a new loop in the future.

jlelong:
I tried to add tests for detecting infinite loop issues, but I have difficulties catching the error `rejected promise not handled within 1 second: RangeError: Maximum call stack size exceeded` without making a lot of changes in the extension activation code. Any suggestion?

The test is `test/fixtures/rootfile/fixture002`.
To trigger the error comment https://github.com/James-Yu/LaTeX-Workshop/blob/d048124625be33a015d92e4f18a3cb2557119ba1/src/components/manager.ts#L657-L660

The error can be caught during activation by using something like
```ts
    void extension.manager.findRoot().then(() => extension.linter.lintRootFileIfEnabled(), err => {
        extension.logger.addLogMessage('An error occurred in findRoot')
        if (err instanceof Error) {
           extension.logger.logError(err)
        }
    })
```
but then it is not clear to me how I can get inside the test framework.

tamuratak:
How about exposing the real `Extension` object on CI tests? We can directly access what we want to test.

```diff
diff --git a/src/main.ts b/src/main.ts
index 2a3eb030..d55024c9 100644
--- a/src/main.ts
+++ b/src/main.ts
@@ -135,6 +135,7 @@ function registerLatexWorkshopCommands(extension: Extension) {
 
 function generateLatexWorkshopApi(extension: Extension) {
     return {
+        realExtension:  process.env['LATEXWORKSHOP_CI'] ? extension : undefined,
         getGraphicsPath: () => extension.completer.input.graphicsPath,
         builder: {
             isBuildFinished: process.env['LATEXWORKSHOP_CI'] ? ( () => extension.builder.isBuildFinished() ) : undefined
```

We are running tests setting those environment variables:

https://github.com/James-Yu/LaTeX-Workshop/blob/2ca8d13b80f723ff04b7a80c67c2569e740cabb6/test/runTest.ts#L58-L61

jlelong:
I am afraid the three steps you mention are not enough, because in `findRoot`, `parseFileAndSubs` is only called once a `rootFile` has been found

https://github.com/James-Yu/LaTeX-Workshop/blob/ead725e184971a5b58f13a06fc998aa012c70170/src/components/manager.ts#L337-L344

Finding the `rootFile` is based on https://github.com/James-Yu/LaTeX-Workshop/blob/ead725e184971a5b58f13a06fc998aa012c70170/src/components/manager.ts#L586

The functions `parseFileAndSubs` and `getTeXChildren` are completely independent.

tamuratak:
I see. Are you saying that we also have to test whether a value returned by `Manager.getCachedContent` includes all the appropriate file paths?

jlelong:
Yes, this may be an alternative approach to catching the error. Probably easier to implement. I will give it a try.

jlelong:
If you are fine with the test, I think it is ready to be merged.

tamuratak (approved):

AguirreNicolas:
I would like to inform (**sorry if this is not the right place**) that the "auto-completion" does not work correctly when the glossaries label has a " _ " symbol. 
What follows " _ " simply does not appear. 

Here is an example:

gloss_def
vscode


It seems that it does not show what follows " _ ", but nevertheless shows the description "E _ P", although later in the construction, the acronym of {E} (Elastic Modulus) appears.

PR #2803, opened 2021-07-21
Added support for LaTeX Songs in syntax highlighting

DPDmancul:
This commit adds support for [LaTeX Songs environment](http://songs.sourceforge.net/) in syntax highlighting.

As can be seen in the following image (on left) currently the syntax to write a chord is mistaken by a math display which never closes; with this patch (on right) there is no more this problem.
Screenshot_20210721_101950-1


jlelong (changes requested):
Can you remove the sample `.tex` files from your PR? I do not want to merge them.
You can leave them in the conversion associated to the PR for test purposes.

DPDmancul:
Done

PR #2815, opened 2021-07-30
Set the host's UID and GID for the executed command within the docker container

123456abcdef:
The default user within a container is root (id=0). The written files in the mounted folder ($pwd) will end up with wrong/different file ownership. Starting the container with the host's UID and GID will result in files written with the "correct" ownership.


tamuratak:
Rejected.

Please use [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) if you need some tweaks.

We will not add tweaks to our docker support on our side.

PR #2821, opened 2021-08-03
fix: properly remove temporary files from latexindent

gakada:
There are multiple ways in which [`src/providers/latexformatter.ts`](https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/providers/latexformatter.ts) can fail to remove its temporary files `__latexindent_temp.tex` and `indent.log`:

- on OS errors (they aren't deleted in the `'error'` handler)
- on formatting errors (not deleted under `code !== 0`, this is relevant for https://bugs.archlinux.org/task/60210)
- when attempting to format empty files (not deleted before the final`return`).

This is an attempt to handle those cases: `removeTemporaryFiles` is always called in the `'error'` and `'close'` handlers.

### DIFF
--- a/src/providers/latexformatter.ts
+++ b/src/providers/latexformatter.ts
@@ -164,6 +164,14 @@ export class LaTexFormatter {
             const temporaryFile = documentDirectory + path.sep + '__latexindent_temp.tex'
             fs.writeFileSync(temporaryFile, textToFormat)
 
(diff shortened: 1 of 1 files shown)


James-Yu:
This PR looks good to me. If no objection, I will merge in few days.

jlelong:
I agree that this PR deserves to be merged.

PR #2823, opened 2021-08-04
Add config.latex-workshop.bind.enter.key to when clauses.

tamuratak:
To disable the onEnterKey command when configured, add config.latex-workshop.bind.enter.key to when clauses.

### DIFF
--- a/package.json
+++ b/package.json
@@ -704,32 +704,32 @@
       {
         "command": "latex-workshop.onEnterKey",
         "key": "enter",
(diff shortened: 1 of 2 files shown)


jlelong:
With this PR, I think that the following code is not needed anymore 

https://github.com/James-Yu/LaTeX-Workshop/blob/bec6a92b94b7e46581d2324dbb8009f509f0b0b3/src/commander.ts#L421-L426

tamuratak:
I agree.

PR #2837, opened 2021-08-16
Strip .tex extension in \input completion

innerlee:
Currently the auto-completion for `\input` will insert the filename with extension. This pr provides option to strip certain extensions.
E.g., before: `\input{introduction}`, after: `\input{introduction}`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1646,6 +1646,16 @@
           ],
           "markdownDescription": "Patterns to ignore in file completion"
         },
+        "latex-workshop.intellisense.file.stripExtension": {
+          "type": "array",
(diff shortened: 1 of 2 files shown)


James-Yu:
As per discussions in https://tex.stackexchange.com/questions/319506/why-do-include-and-input-handle-extension-name-differently and https://tex.stackexchange.com/questions/508691/different-file-extension-for-input-files, I think stripping the extension is a complicated issue that is quite personal choice. To avoid burdening the maintenance effort, I'd close this PR for now. Discussions welcomed.

PR #2841, opened 2021-08-19
Add BibLaTeX backend for intellisense  (Fix #2813)

### LINKED ISSUE #2813: Add support for biblatex entries in Intellisense

You visited the wiki?: Yes

While this great extension supports me in writing latex I have the problem that Intellisense does not have support for the biblatex entry types (i.e. online). It would be much easier to write the documents if Intellisense would support the Biblatex-entries.

It would be nice to create the following two files for biblatex so that it will get supported by Intellisense:

- https://github.com/James-Yu/LaTeX-Workshop/blob/master/data/bibtex-entries.json
- https://github.com/James-Yu/LaTeX-Workshop/blob/master/data/bibtex-optional-entries.json

As some entries differ in their required and optional arguments (i.e. book) it would be the best way to create additional files. I think there should then also be a config option so that the user can decide if he'd like to use BIBTeX or biblatex.
For example there could be the following config option:
```
    "latex-workshop.intellisense.citation.backend": "",
```
The values could be `bibtex` and `biblatex` with `bibtex` as default to not break the workflow of some users.

## Desktop [Required]

 - OS: Fedora 34
 - VS Code version: 1.58.2
 - Extension version: v8.19.2
 - TeX distribution version: texlive

### Are you using VSCodium?

Yes

No

No

caminsha:
In this PR there are the entries to support IntelliSense for BibLaTeX. This will close this feature request: https://github.com/James-Yu/LaTeX-Workshop/issues/2813

At the moment the following entries are not inside the json files:
- set
- xdata
- custom[a-f]

These entries are not included as these are special things and not normal entries.


jlelong:
Close #2813

PR #2848, opened 2021-08-28
Update mathtools_cmd.json

jiajia2u:
add "coloneqq"

### DIFF
--- a/data/packages/mathtools_cmd.json
+++ b/data/packages/mathtools_cmd.json
@@ -9,6 +9,16 @@
     "package": "mathtools",
     "snippet": "clap{${1:text}}"
   },
+   "coloneq": {
+    "command": "coloneq",
+    "package": "mathtools",
+    "snippet": "coloneq"
+  },
+  "coloneqq": {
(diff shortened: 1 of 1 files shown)


jlelong:
The file ending with `_cmd.json` are generated from the `.cwl` files of https://github.com/LaTeXing/LaTeX-cwl. This repository does not seem to be maintained so I maintain a fork at https://github.com/jlelong/LaTeX-cwl. If you intend to make a PR on the `.cwl` files, consider to make it at https://github.com/jlelong/LaTeX-cwl

So, there is no point making a PR on `_cmd.json` files.

As far as the commands `coloneqq` & co are concerned, they are already provided by http://milde.users.sourceforge.net/LUCR/Math/data/unimathsymbols.txt , which is used to generate https://github.com/James-Yu/LaTeX-Workshop/blob/master/data/unimathsymbols.json. See https://github.com/James-Yu/LaTeX-Workshop/wiki/Intellisense#latex-workshopintellisenseunimathsymbolsenabled

PR #2852, opened 2021-08-31
Highlight markdown environment

### LINKED ISSUE #2850: Syntax highlighting for {markdown} environment

I'm using `markdown` package to write markdown file in LaTeX (maybe it is faster and more familiar to me?). It is obvious that syntax highlighting of markdown disappears. 

## To Reproduce  [Required]
test file

```tex
\documentclass{ctexart}
\usepackage{amsmath}
\usepackage[hybrid, fencedCode]{markdown}
\markdownSetup{underscores = false}

\begin{document}
\begin{markdown}

# Heading

1. list
2. list

## Another Test

- test list
    - test list

> quotation

$e^{i\theta }=\cos\theta +i\sin \theta$

\begin{equation}
    (a+b)^2=a^2+2ab+b^2
\end{equation}

Here is text **Here is Test** *Here is test* `coding`


\end{markdown}
\end{document}
```
No need to build. (If you want, you can build it with XeLaTeX * 2)

It looks like:
image

I'm not so familiar with `json` language ... as well as "Pull Request". Here is my test: Add the following code to `%userprofile%\.vscode\extensions\james-yu.latex-workshop-8.20.2\syntax\LaTeX.tmLanguage.json`.
```json
  {
    "begin":"(^\\s*)?(?=\\\\begin\\{(markdown)\\})",
    "end":"(\\\\end\\{markdown\\})",
    "contentName" : "meta.function.environment.markdown.latex",
    "patterns": [
      {
        "include": "$base"
      },
      {
        "include": "text.html.markdown"
      }
    ]
  },
```
Add them to 
image

Restart VSCode, then the workspace will be:

image

## Known issues

- inline math and command will lose highlighting
- As well as `\begin{markdown}` and `\end{markdown}`......

image

```
no logs
```

## Desktop [Required]

 - OS: Windows 10
 - VS Code version: 1.59.1
 - Extension version: v8.20.2
 - TeX distribution version: TexLive 2021

THANKS!! 😀

jlelong:
This PR is an attempt to hightlight the content of the `markdown` environment as `md`.  

Current status

- The content of the `markdown` environment must not be indented by more than 2 or 3 spaces  for the syntax highlighting to work.
- Inline LaTeX math is not highlighted inside the `markdown` environment unless it starts at column 0. This is because the `md` syntax matches anything starting with a `[0-9a-zA-Z ]` as `meta.paragraph.markdown` and there is no way to include the LaTeX syntax at this point.

Close #2850.


jlelong:
To overcome the difficulty of combining LaTeX and markdown syntaxes, a possible approach is for LW to define a new language id with its own syntax obtained by including the LaTeX syntax into the markdown one.

jlelong:
Now, this PR defines a new `markdown-latex` language and we embed it into the LaTeX grammar for the content of the `markdown` environment. This approach enables us to completely solve all the above issues and many more.

It remains one issue, `` ` `` is automatically closed by `'`. I believe this is an issue in VSCode, let us see what they answer to my bug report microsoft/vscode/issues/133397

tamuratak:
- I prefer `markdown_embedded_latex` or other longer identifiers.  `markdown_latex` can conflict with identifiers provided by other VS Code extensions.



- We had better make the copyright and the license of `markdown-latex.tmLanguage.json` clear.
 1. Copy https://github.com/microsoft/vscode/blob/main/LICENSE.txt to `syntax/markdown-latex-license.txt`. 
 1. Create `syntax/README.md` and add that the license of  `markdown-latex.tmLanguage.json` is `syntax/markdown-latex-license.txt`.

jlelong:
I agree with using longer identifiers, which are more likely to be unique.

Concerning the licence, I assume we should do the same for `cpp-grammar-bailout.tmLanguage.json`, which is generated from https://github.com/jeff-hykin/better-cpp-syntax/

syntax/markdown-latex-combined-license.txt
```diff
@@ -0,0 +1 @@
+<html><body>You are being <a href="https://raw.githubusercontent.com/microsoft/vscode-markdown-tm-grammar/main/LICENSE.txt">redirected</a>.</body></html>
```
tamuratak, line 1: It would be better to copy the file itself instead of writing a link to it.

tamuratak (changes requested):

syntax/markdown-latex-combined-license.txt
jlelong: My bad! I messed up the `curl` command when downloading the license file.

jlelong:
OK, I have added a section on the LaTeX grammar in README.md.

tamuratak (approved):

PR #2862, opened 2021-09-08
Do not sort @xdata entries in .bib file

### LINKED ISSUE #2861: Avoid sorting @xdata in .bib file
When `Latex-workshop › Bibtex-fields › Sort: Enabled` & `Latex-workshop › Bibtex-format › Align-equal: Enabled` is also sorting `@xdata`

@String is not sorted as it should be,

### MWE

```
@string{ucp = {University of California Press}}

@xdata{oup,
  publisher = {Oxford University Press},
  location  = {Oxford}
}

@book{annas:plato,
  author    = {Annas, Julia},
  title     = {Plato},
  subtitle  = {A very short introduction},
  year      = {2003},
  xdata     = {oup},
  isbn      = {019280216X},
  series    = {Very short introductions},
}
@book{genette:architext,
  author       = {Genette, G{\'e}rard},
  title        = {The architext},
  subtitle     = {An introduction},
  year         = {1992},
  foreword     = {Scholes, Robert},
  translator   = {Lewin, Jan E.},
  publisher    = ucp,
  isbn         = {0-520-07661-3},
  series       = {Quantum books},
  location     = {Berkeley and Oxford},
  origlanguage = {french}
}
```
After `sorting` you will get `@xdata` at the end of the file, but not `@string` which remains as it should be in the top of the file.

jlelong:
Fix #2861.

Do not move `@xdata` entries when sorting the file.


tamuratak:
I don't know nothing about `@xdata`. However, it might make things simpler to move `@xdata` and `@string` before bibtex entries instead of leaving them at the original positions.

I am not sure what we should do if `@xdata` exists in the middle of a `.bib` file.

jlelong:
I agree with you. Actually, I am not really happy with my PR. The question about the position of `@xdata` is more or less the same as where to declare variables: at the beginning of the function or right before using them.

I am considering adding two new settings: the list entries not to be moved and the list of entries to put first. This should make everyone happy.

jlelong:
Humm. Thinking about it twice, _entries not to be moved_ does not make much sense as all the other entries will be moved. Hence, the original position becomes meaningless. Hence, I will only define a list of entries to be put first with the default value `["string", "xdata"]`.

jlelong:
I have tempted to handle `@string` and other entries together but it is not so obvious as they are parsed as different `bibtexParser` types. Not sure this is the proper way to proceed. @tamuratak any opinion?

tamuratak:
I think the way of treating entries is appropriate. However, we had better move `src/utils/bibtexutils.ts` to `src/components` and define `BibtexUtils` class since the code is enough large. Moving the file to `src/providers/bibtexformatterlib` is another choice.

tamuratak (approved):

PR #2886, opened 2021-09-28
Prettier bug report template.

real-yfprojects:
Rewrite template to make it prettier and improve readability.

I used html comments for explainers and collapsible sections (using `<details>`) for log messages.
I also changed wording a little bit. All in all you should be able to get the same information from this new bug report.


tamuratak:
Not convinced.

- We want to keep the source code readable for users. 
- It is OK even if the output is not pretty. We are used to the current one.

PR #2900, opened 2021-10-10
Add more placeholders

### LINKED ISSUE #2884: More flexibility for placeholders in "latex-workshop.latex.outDir" and "latex-workshop.latex.external.build.args"

You visited the wiki?: Yes

I use an alternative build system (Bazel, which uses LuaLatex provided by [these rules](https://github.com/ProdriveTechnologies/bazel-latex) under the hood), which makes different assumptions for project structure than what LatexWorkshop seems to. Specifically, Bazel defines a workspace in some root folder (the toplevel directory I would open in VSCode), and then describes build targets that can be referenced *relative to that root folder*. For instance, with the following directory structure:

```
root/
    WORKSPACE  ------ [the toplevel file that Bazel looks for to define the root]
    report/
        BUILD.bazel  ---- [file that describes the build target for the latex pdf (inc. dependencies, files, etc)]
        my_report.tex  ---- [the main tex file]
    bazel-bin/  ----- [this is a generated directory which is managed by Bazel, similar to the build/ convention in C++ projects]
        report/
            my_report.pdf   ---- [ Bazel generates this file as the output of build target 'my_report']
```

Say report/BUILD.bazel defines a "build target" called `my_report`, which compiles my_report.tex into my_report.pdf. To build this from the command line, you would run:

```
bazel build //report:my_report
``` 

In general, the command line syntax for building is `bazel build //<path_to_target_relative_to_root>:<target_name>`, which produces an output at:
`root/bazel-bin/<path_to_target_relative_to_root>/<target_name>.pdf`

This relative directory structure produces two issues I can't figure out how to resolve elegantly with existing placeholders:

1. BUILDING: In `latex-workshop.latex.external.build.args`, I need to get the path relative to the current build target from the root folder. Currently, my workaround is to put all build targets in a BUILD.bazel file in the root directory that have the same name as the .tex file, and use `%DOCFILE%` to get the file name. This is pretty restrictive and error-prone. 

2. VIEWING: In `latex-workshop.latex.outDir`, the correct Bazel output directory (`bazel-bin`) is defined relative to the root workspace as well. Currently, i'm just hardcoding this on a per-project level to get things to work, but it's frustrating if I need to work across computers.

## To Reproduce  [Required]

If necessary, I can provide a sample project with this issue.

I propose a couple new placeholder variables:

1. %WORKSPACE%: this would be the toplevel directory currently open in a VSCode project. As far as I understand, this is a well-defined concept in VSCode already (and not specific to Bazel), since VSCode places project-related configuration at root/.vscode/settings.json already. 
2. %DOC_RELATIVE%: this would be the path of the document relative to the toplevel root directory.

With these settings, I could set:
 ```"latex-workshop.latex.outDir": %WORKSPACE%/bazel-bin/%DOC_RELATIVE%```
and
```
"latex-workshop.latex.external.build.args": [
        "build",
        "//%DOC_RELATIVE%:%DOCFILE%"
    ]
```

A small note: currently, I'm making the assumption that %DOCFILE% (the main latex source file) is the same name as the build target I define in BAZEL.build, which isn't too cumbersome. To relax that assumption would require a much more involved solution using some introspection with the Bazel build system, so I don't think that's really in scope of this project.

If it were possible to implement these two placeholders, I'd be happy to contribute a HOW-TO section or something regarding using LatexWorkshop w/ Bazel.

## Logs [Required]

N/A

N/A

N/A

## Screenshots

N/A

## Desktop [Required]

N/A

### Are you using VSCodium?

N/A

N/A

N/A

If the answer is Yes, please write which one you are using. Write the versions of the remote extension.

N/A

### Additional context

N/A

jlelong:
This PR adds three new placeholders

- `WORKSPACE_FOLDER`: current workspace path
- `RELATIVE_DIR`: file directory relative to the workspace folder
- `RELATIVE_DOC`: file path relative to the workspace folder

Close #2884

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -144,7 +144,7 @@ export function replaceArgumentPlaceholders(rootFile: string, tmpDir: string): (
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
         const docker = configuration.get('docker.enabled')
 
-        const workspaceDir = vscode.workspace.workspaceFolders ? vscode.workspace.workspaceFolders[0].uri.fsPath : ''
+        const workspaceDir = vscode.workspace.workspaceFolders ? vscode.workspace.workspaceFolders[0].uri.fsPath.split(path.sep).join('/') : ''
         const rootFileParsed = path.parse(rootFile)
         const docfile = rootFileParsed.name
         const docfileExt = rootFileParsed.base


tamuratak:
Notice that `Uri.fsPath` returns a path with the separator `\`, not `/`, on Windows.

- https://code.visualstudio.com/api/references/vscode-api#Uri

jlelong:
Good point. I always forget that some systems use `\` as path separator 🤪.

To be coherent with the other placeholders, we should use `/` even on Windows. I am not convinced `_W32` counterparts are really useful, especially as LaTeX commands expect `/` as path separator.

Issue #2898, opened 2021-10-08
Custom latexmk (lualatex) does not compile 


Neolize:
### Disable all the other extensions except for LaTeX Workshop, restart VS Code, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: Yes

## Describe the bug [Required]

If I use my custom recipes with latexmk (configured to use lualatex), I get an error message (paths have manually been shortened here):

Rc files read:
  .latexmkrc
Latexmk: This is Latexmk, John Collins, 29 May 2021, version: 4.74b.
Rule 'lualatex': File changes, etc:
   Changed files, or newly in use since previous run(s):
      'c:/.../test2.tex'
------------
Run number 1 of rule 'lualatex'
------------
------------
Running 'lualatex  --max-print-line=10000 --shell-escape --synctex=1 --interaction=nonstopmode -file-line-error -recorder  "c:\...\test2.tex"'
------------
Latexmk: applying rule 'lualatex'...
lualatex: unrecognized option '--max-print-line=10000'
Latexmk: fls file doesn't appear to have been made.
Collected error summary (may duplicate other messages):
  lualatex: Command for 'lualatex' gave return code 1
      Refer to 'test2.log' for details
----------------------
Latexmk: Errors, so I did not complete making targets
This message may duplicate earlier message.
Latexmk: Failure in processing file 'c:/.../test2.tex':
   *LaTeX didn't generate the expected log file 'test2.log'
----------------------
Latexmk: Use the -f option to force complete processing,
 unless error was exceeding maximum runs, or warnings treated as errors.

I have looked into the "LaTeX Workshop" (NOT the "LaTeX Complier") and ONLY when using the lualatex version I get the error:

The environment variable $PATH: undefined
The environment variable $SHELL: undefined

I have added vscode and the extension manually to my system path (using Windows 10), but it does not seem to be working.

Also my latexmk with xelatex does not have a problem at all and the built-in console in VS code works fine aswell with the latexmk -lualatex option. 

### To Reproduce

(see above)

### Expected behavior

I want the document to compile with my custom recipe, but it does not work. 

## Logs [Required]

```
[...]

PR #2917, opened 2021-10-19
Treat latexmk --luatex properly.

tamuratak:
Treat `latexmk --luatex` properly.

Related to #2898

### DIFF
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -490,7 +490,8 @@ export class Builder {
                 if (!step.args) {
                     step.args = []
                 }
-                if (this.isMiktex && ((step.command === 'latexmk' && !step.args.includes('-lualatex') && !step.args.includes('-pdflua')) || step.command === 'pdflatex')) {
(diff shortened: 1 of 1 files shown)


tamuratak:
I am not sure it is appropriate to add `--max-print-line=` when the option `-pdflualatex="COMMAND"` given.

Since I don't know much about latexmk, I would like you to tweak for such cases after this PR merged.

src/components/builder.ts
```diff
@@ -490,7 +490,8 @@ export class Builder {
                 if (!step.args) {
                     step.args = []
                 }
-                if (this.isMiktex && ((step.command === 'latexmk' && !step.args.includes('-lualatex') && !step.args.includes('-pdflua')) || step.command === 'pdflatex')) {
+                const isLuaLatex = step.args.includes('-lualatex') || step.args.includes('-pdflua') || step.args.includes('--lualatex') || step.args.includes('--pdflua')
```
jlelong, line 493: It is not appropriate to add `--max-print-line=` when the option `-pdflualatex="COMMAND"` is given.  Just add `step.args.includes('-pdflualatex') || step.args.includes('--pdflualatex')` to the alternation.

PR #2948, opened 2021-10-31
Provide @snippets as completion items

### LINKED ISSUE #2920: Provide `latex/snippets.json` as completion items
Related to #2919.

@jlelong  the behavior of snippets is fragile and can be annoying. To avoid the drawbacks, we can explore to provide snippets with the `@` prefix defined in `latex/snippets.json` as completion items.

How do you think?

jlelong:
Close #2920


jlelong:
There are two main approaches to provide former `@snippets` as completion items

1. Use the current `Completer` class as the unique `CompletionItemProvider` and have `@`trigger it. This PR currently relies on this approach https://github.com/James-Yu/LaTeX-Workshop/blob/de2edc73874bf23bee73ec1b731f22ea46a34690/src/main.ts#L277-L281
2. Define a new `CompletionItemProvider` only triggered by `@`.

Before going any further on refactoring all this, I would like to choose one approach or the other. I think the second approach is a bit cleaner. @tamuratak what do you think?

tamuratak:
I prefer the second approach too.

jlelong:
Close in favour of #2950.

PR #2950, opened 2021-11-01
Provide `@snippets` as completion items using a separate `CompletionItemProvider`

### LINKED ISSUE #2920: Provide `latex/snippets.json` as completion items

jlelong:
This PR is a preferred alternative approach to #2948.
Close #2920


jlelong:
@tamuratak Can I merge this PR?
I will also implement a replacement mechanism such as the one for [commands](https://github.com/James-Yu/LaTeX-Workshop/wiki/Intellisense#latex-workshopintellisensecommandsJSONreplace)

src/providers/completer/snippet.ts
```diff
@@ -0,0 +1,68 @@
...
+                return filteredSuggestions.map(item => {
+                    item.range = new vscode.Range(args.position.translate(undefined, -item.label.toString().length), args.position)
+                    return item
+                })
+            }
+        }
+        return suggestions
+    }
+
+    private provide(document?: vscode.TextDocument, position?: vscode.Position): vscode.CompletionItem[] {
```
tamuratak, line 57: Do arguments have to be optional? Could we remove the `?` in the argument?

tamuratak (changes requested):

src/providers/completer/snippet.ts
jlelong: Clumsy copy paste from `command.ts`.

tamuratak (approved):

PR #2959, opened 2021-11-08
Remove Snippet Panel

tamuratak:
Remove Snippet Panel to reduce the maintenance burden.

We don't have to have both Snippet Panel and Snippet View at the same. We can move Snippet View to the panel if we want to make the view larger. We can also move the panel to the right or the left. I think it's enough.

@jlelong  how do you think?

スクリーンショット 2021-11-08 12 26 46


jlelong:
I agree that we do not need the Snippet Panel anymore.

PR #2967, opened 2021-11-10
Fix ranges for (sub)*section entries in the structure provider

### LINKED ISSUE #2961: Breadcrumbs for higher level symbols
## Preliminary questions [Required]

<details> <summary>Preliminary questions</summary>

  You visited the wiki?: Yes
  
</details>

It seems like there's a feature missing: LatexWorkshop provides a full outline of a `.tex` file, and if I launch the command `breadcrumbs.focusAndSelect` it does reach all sections and labels.

However, the breadcrumbs topbar only shows three dots for any symbol not root level:
`file > ...` 
where for root level symbols we get a full path
`file > symbol`

The more limiting aspect of it is that, when I launch the command `breadcrumbs.focusAndSelect` (through the default keybind `Ctrl`+`Shift`+`.`) the breadcrumbs menu shown starts from the first symbol instead of the current symbol.

## To Reproduce  [Required]
In this MWE you can see it by placing your cursor in the scope of Subsection1.1

Minimal working example:
```latex
\documentclass[12pt]{article}

\begin{document}

\section{Section1}
Lorem ipsum

  \subsection{Subsection1.1}
  Lorem ipsum

\section{Section2}
Lorem ipsum

\end{document}
```

I'd expect a full path to labels and subsections to be possible. But I'm primarily interested in support of the `breadcrumbs.focusAndSelect` starting at the current symbol. It would make navigation within `.tex` files much more efficient.

<details><summary>Logs aren't necessary</summary>

## Logs [Required]

Please paste the whole log messages here, not parts of ones. The log should start with `Initializing LaTeX Workshop`. It is very important to identify problems.

```

[Paste the log here]
```

```
To access the log, click 'help' -> 'Toggle Developer Tools' -> 'Console'. Paste anything suspicious.

[Paste the log here]
```
</details>

## Screenshots

image

image

## Desktop [Required]

 - OS: Ubuntu 20.04
 - VS Code version: 1.62 x86
 - Extension version: 8.21.1
 - TeX distribution version: TeX Live 2019.20200218-1

<details><summary>Additional questions</summary>

### Are you using VSCodium?

No

Yes

No

</details>

Thanks for the great extension, by the way :+1:

jlelong:
Fix #2961

### DIFF
--- a/src/providers/structure.ts
+++ b/src/providers/structure.ts
@@ -10,6 +10,7 @@ export class SectionNodeProvider implements vscode.TreeDataProvider<Section> {
 
     private readonly _onDidChangeTreeData: vscode.EventEmitter<Section | undefined> = new vscode.EventEmitter<Section | undefined>()
     readonly onDidChangeTreeData: vscode.Event<Section | undefined>
+    private readonly hierarchy: string[]
     private readonly showLabels: boolean
     private readonly showFloats: boolean
     private readonly showNumbers: boolean
@@ -22,6 +23,7 @@ export class SectionNodeProvider implements vscode.TreeDataProvider<Section> {
         this.onDidChangeTreeData = this._onDidChangeTreeData.event
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
 
+        this.hierarchy = configuration.get('view.outline.sections') as string[]
         this.showLabels = configuration.get('view.outline.labels.enabled') as boolean
         this.showFloats = configuration.get('view.outline.floats.enabled') as boolean
         this.showNumbers = configuration.get('view.outline.numbers.enabled') as boolean
@@ -57,6 +59,10 @@ export class SectionNodeProvider implements vscode.TreeDataProvider<Section> {
             newFileStack = fileStack
         }
         newFileStack.push(filePath)
+
+        if (!sectionNumber) {
+            sectionNumber = Array<number>(this.hierarchy.length).fill(0)
+        }
 
         let content = this.extension.manager.getDirtyContent(filePath)
         if (!content) {
@@ -184,16 +190,15 @@ class StructureModel {
     private readonly hierarchy: string[]
     private readonly headerPattern: string
     private readonly sectionDepths = Object.create(null) as { [key: string]: number }
-    private sectionNumber: number[]
     private prevSection: Section | undefined = undefined
 
     constructor(
         private readonly extension: Extension,
-        public filePath: string,
-        public rootStack: Section[],
-        public children: Section[],
-        public fileStack: string[],
-        sectionNumber: number[] | undefined
+        private readonly filePath: string,
+        private readonly rootStack: Section[],
+        private readonly children: Section[],
+        private readonly fileStack: string[],
+        private sectionNumber: number[]
     ) {
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
         this.hierarchy = configuration.get('view.outline.sections') as string[]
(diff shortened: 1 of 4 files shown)


jlelong:
@tamuratak Could you test this PR? Let me know if you some TOC examples which break the range of any `Section` item.

tamuratak:
`t.tex`:

```latex
\documentclass{article}
\begin{document}

\section{Section1}
Lorem ipsum

  \subsection{Subsection1.1}
  Lorem ipsum

\section{Section2}
Lorem ipsum

\input{t1.tex}

abc

\section{Section4}
Lorem ipsum

\end{document}
```

`t1.tex`:

```latex
\section{Section3}
Lorem ipsum
```

The structure view is correct:

スクリーンショット 2021-11-14 10 11 59

The breadcrumbs is not correct:

スクリーンショット 2021-11-14 10 12 11

I am not sure the breadcrumbs should work in this case.

jlelong:
From [VSCode documentation](https://code.visualstudio.com/docs/editor/editingevolved#_breadcrumbs)

> Breadcrumbs always show the file path and, with the help of language extensions, the symbol path up to the cursor position. The symbols shown are the same as in Outline view and Go to Symbol.

The symbols used by breadcrumbs are limited to the current file and come from the [`DocumentSymbolProvider`](https://code.visualstudio.com/api/references/vscode-api#DocumentSymbolProvider) . A [`DocumentSymbol`](https://code.visualstudio.com/api/references/vscode-api#DocumentSymbol) must be contained in a single file (the `range` field cannot spread across files).

In you example, the line with `abc` should be enclosed in the range of `Section3`, which is defined in an other file. This is impossible. The issue is that LaTeX accepts much more versatile project splittings across several files than any other languages (a scope name or a class cannot spread across multiple files)

Unfortunately, I do not think that breadcrumbs can work in this case as it relies on the

tamuratak:
> This is impossible. The issue is that LaTeX accepts much more versatile project splittings across several files than any other languages (a scope name or a class cannot spread across multiple files)

I see.

src/providers/structure.ts
```diff
@@ -90,152 +74,62 @@ export class SectionNodeProvider implements vscode.TreeDataProvider<Section> {
...
-                    // Avoid circular inclusion
-                    if (inputFilePath === filePath || newFileStack.includes(inputFilePath)) {
-                        continue
-                    }
-                    if (prevSection) {
-                        prevSection.subfiles.push(inputFilePath)
-                    }
+               const inputFilePath = structureModel.buildImportModel(line)
+               if (inputFilePath) {
                     this.buildModel(inputFilePath, newFileStack, rootStack, children, sectionNumber)
```
tamuratak, line 91: Is there possibility that this line results in an infinite loop?

jlelong: It could happen and in this case the promise returned by `provideDocumentSymbols` will fail. I am not sure how far we should go to properly handle the infinite loop error as it basically means that either there is an error in this file or it is a false circular inclusion because we de not interpret LaTeX code.

tamuratak: The following is an example I have in mind to avoid infinite loops.

- https://github.com/tamuratak/LaTeX-Workshop/tree/use_memo
- https://github.com/James-Yu/LaTeX-Workshop/commit/6a36a50b38cdd932a007b95230d3e2f4cf6303d7

jlelong: That is indeed the simplest way of doing it. We already use this pattern quite a lot in LW.

tamuratak (approved):

PR #2986, opened 2021-11-27
Disable commands and features on virtual workspaces

tamuratak:
We had better disable commands and features on virtualworkspaces.  Extensions can be activated on virtualworkspaces even if `"virtualWorkspaces": false`.

See https://github.com/microsoft/vscode/issues/137152#issuecomment-974045485

> Extensions should always check and listen for workspace/folder changes and disable functionality when there's a workspace/folder they cannot work with.

### DIFF
--- a/package.json
+++ b/package.json
@@ -318,18 +318,21 @@
         "command": "latex-workshop.build",
         "title": "Build LaTeX project",
         "category": "LaTeX Workshop",
-        "icon": "$(debug-start)"
+        "icon": "$(debug-start)",
+        "enablement": "!virtualWorkspace"
       },
       {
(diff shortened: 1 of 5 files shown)


jlelong:
Issue microsoft/vscode#137152 is a good catch!

jlelong (commented):
All the changes look fine to me. 

I am just wondering whether we should check if the workspace is not virtual in https://github.com/James-Yu/LaTeX-Workshop/blob/eeb4c1be4f47d92b233e58d88c0094084e67e2ee/src/providers/projectsymbol.ts#L13

jlelong (approved):

Issue #2731, opened 2021-05-30
 Basic language extension as a built-in extension in VS Code


tamuratak:
https://github.com/James-Yu/LaTeX-Workshop/commit/8dfaeb2cc314872293546fa453de352940b88bea#commitcomment-51335775

> I wanted to bring to your attention that setting "virtualWorkspaces": false will disable the whole extension, inlcuding grammars and language configuration contributions.
> To avoid this, we recommend to split off a basic language extension (grammars, language configuration, snippets) form the rich language support and have two extensions. Given the popularity of LateX, we would be willing to host such a basic language extension as a built-in extension in VS Code. Let me know if you are interested.
> See also https://github.com/microsoft/vscode/wiki/Virtual-Workspaces#disabling-a-language-extension

aeschli:
Thanks for the issue. That's better.
We were discussing about where the grammars would be hosted, You currently have them in the same repo as the extension.
That's possible, but might be confusing as the extension, at some point, won't contain them any more (but rely on the built-in ones.

Do you prefer hosting the grammars where you have them now, or do you want to look into extracting them in a separate repo? The separate repo has the advantage that other editors/ides might want to consume that repo as well and you can easier invite other contributors to share the maintenance.

jlelong:
Sorry for my silence, I am quite overworked these days and will not really be available before next week.

In the long term, I guess it would be clearer to host the grammars into a separate repo. 

I have a few questions

- Just to be sure, we would need to extract https://github.com/James-Yu/LaTeX-Workshop/tree/master/syntax for sure but is there anything else that should be moved too? I am thinking of https://github.com/James-Yu/LaTeX-Workshop/tree/master/snippets
- The language `id`'s are defined by the basic language extension, aren't they?
-  it is not clear how to handle the transition: new versions of the extensions (without the grammars) will not work with older VSCode versions with no LaTeX built-in grammars. We should probably keep a copy of the grammars in the extension repo for a while, shouldn't we?

aeschli:
No worries, there's no rush. We are now in May endgame, so I won't have time myself in the next 1.5 weeks anyways.

The basic language extension would not contain any snippets. It's better they come from extensions. Users often disagree what a good snippet is and we don't want to be the judge on that. Also we don't have a mechanism to disable specific snippets that code from built-in extensions.
So I suggest to keep the snippets in your extension.

The language ids with file associations all go in the basic language extension

[...]

PR #2993, opened 2021-11-29
Remove intellisense tricks from autoClosingPairs

jlelong:
This PR is related to https://github.com/James-Yu/LaTeX-Workshop/issues/2731#issuecomment-981348890

It seems that intellisense for everything involving `(`, `{`, `[` just works fine even without explicitly disabling the auto closing using https://github.com/jlelong/vscode-latex-basics/blob/898c5155fc8a657df293ec6abc19890e9c0b3247/latex-language-configuration.json#L22-L48

@tamuratak Can you test this PR? Am I missing something? That would be good news.


tamuratak:
> Am I missing something? 

Without these tweaks, we see the following behavior. I am not sure it is acceptable or not:

https://user-images.githubusercontent.com/10665499/143957215-cd284ea7-7b5d-40fb-aa59-6a17610c4f4e.mov


We can also remove the same parts in other `language-configuration.json` files.

- https://github.com/James-Yu/LaTeX-Workshop/blob/02dcacc2638962c1feb6206f7b8b277c95e6de25/syntax/latex-weave-language-configuration.json#L24-L43

jlelong:
The only broken behaviour occurs when intellisense is triggered inside `\left()`. Actually, I think the real drawback of removing these tweaks occurs when inserting `\left(   \right.` Users will have to first delete the closing `)`.

Nov-30-2021 06-17-55

The most efficient way of dealing with this use case is to keep the tweaks. As this is not actually related to intellisense, I think that these auto closing definitions should not be considered as tweaks but as real requirements for the LaTeX language. Therefore, I am in favour of keeping these definitions in the language definitions provided by the basic LaTeX language extension.

tamuratak:
> Therefore, I am in favour of keeping these definitions in the language definitions provided by the basic LaTeX language extension.

`vscode-latex-basics` should work with other LaTeX extensions, e.g. vscode-texlab.  I am not sure these tweaks can work with the extensions.  I don't want to be bothered to install other extensions to see whether they work. Moreover, since the behavior of snippets is fragile, we might have to check every VS Code update. At least, I don't want to do that.

tamuratak:
I think enabling LaTeX Workshop and vscode-latex-basics to have different language-configuration.json files makes things simple.

PR #3008, opened 2021-12-11
Support the internal PDF viewer on virtual workspaces

tamuratak:
Support the internal PDF viewer on virtual workspaces. A virtual workspace provided by Live Share v1.0.5196 supports binary files.  This PR enables us to preview PDF files through Live Share sessions, although we have to add more tweaks to make the feature work with Live Share after this PR merged.

- We use `vscode.Uri` instead of `string` to treat the path of PDF files.
- We always add PDF files to the `manager.pdfWatcher` even when we preview them through `extension.commander.pdf`. To prevent the internal PDF viewer from reloading twice, we make `pdfWatcher` ignore PDF files built from root files.

https://github.com/James-Yu/LaTeX-Workshop/blob/ca07039e058f7d5841387d9a2f1bcb5d2af0c589/src/components/builder.ts#L196-L197


jlelong:
This is a hell of a PR 👍
I may need a day or two to walk through it.

tamuratak:
If you don't have enough time, I would be pleased for you to review only `src/components/managerlib/pdfwatcher.ts` and its usage.

src/components/builder.ts
```diff
@@ -112,6 +112,9 @@ export class Builder {
      * @param rootFile The root file to be compiled.
      */
     async buildWithExternalCommand(command: string, args: string[], pwd: string, rootFile: string | undefined = undefined) {
+        if (rootFile) {
+            this.extension.manager.ignorePdfFile(rootFile)
```
jlelong, line 116: External build command generally do not honor the `%OUTDIR%` placeholder and therefore `tex2pdf` is likely to return the wrong path. I am not sure we can really do anything about it.

src/components/managerlib/pdfwatcher.ts
```diff
@@ -7,10 +7,18 @@ export class PdfWatcher {
     private readonly extension: Extension
     private readonly pdfsWatched = new Set<string>()
     private readonly pdfWatcher: chokidar.FSWatcher
+    readonly vscodeFsWatcher: vscode.FileSystemWatcher
+    private readonly pdfUrisWatched = new Set<string>()
```
jlelong, line 11: It seems that this property is never read. We just add elements to it in `watchPdfFile`.
Am I missing something?

jlelong:
> We always add PDF files to the `manager.pdfWatcher` even when we preview them through `extension.commander.pdf`. To prevent the internal PDF viewer from reloading twice, we make `pdfWatcher` ignore PDF files built from root files.

I am not sure to udnerstand why we need this new logic. I fear that reverse engineering if a PDF file is the result of a build is not so robust. With this PR,
- the `pdfFileWatcher` watches any PDF file that has ever been opened either from the `open-preview` command or from the file explorer;
- the `vscodeFsWatcher` watches all PDF files in the current workspace folders.
Why do we need that `pdfFileWatcher` watches any PDF file that has ever been opened?

tamuratak:
> Why do we need that pdfFileWatcher watches any PDF file that has ever been opened?

The current `pdfFileWatcher` has two issues:

- When the internal PDF viewer  is restored at the start of VS Code, the watcher doesn't watch a PDF file the viewer previews.
- When opening a PDF file associated with a root LaTeX file by clicking on the explorer on VS Code, the internal PDF viewer reloads twice on the refresh of the PDF file.

This PR solves the issues.

src/components/builder.ts
```diff
@@ -112,6 +112,9 @@ export class Builder {
...
+        }
```
jlelong, line 117: External build command generally do not honour the `%OUTDIR%` placeholder and therefore `tex2pdf` is likely to return the wrong path. I am not sure we can really do anything about it.

src/components/managerlib/pdfwatcher.ts
jlelong, line 11: It seems that this property is never read. We just add elements to it in `watchPdfFile`.
Am I missing something?

jlelong:
> > Why do we need that pdfFileWatcher watches any PDF file that has ever been opened?
> 
> The current `pdfFileWatcher` at `master` has two issues:
> 
> * When the internal PDF viewer  is restored at the start of VS Code, the watcher doesn't watch a PDF file the viewer previews.
> * When opening a PDF file associated with a root LaTeX file by clicking on the explorer on VS Code, the internal PDF viewer reloads twice on the refresh of the PDF file.
> 
> This PR solves the issues.

Thanks for the explanations. This is clear.

jlelong:
The function `build` calls `this.extension.manager.ignorePdfFile(rootFile)`. I think we should do the same in `Viewer.OpenBrowser` and `Viewer.OpenTab` to handle the case of opening the PDF file associated to the root file with triggering the build.  It typically happens when reopening a workspace in which the PDF has already been built.

tamuratak:
> the case of opening the PDF file associated to the root file with triggering the build. It typically happens when reopening a workspace in which the PDF has already been built.

Could you elaborate the sequence more?

src/components/managerlib/pdfwatcher.ts
tamuratak: Thank you. It is a bug. I will fix it.

jlelong:
My bad. I meant "without" instead of "with". 

Consider a workspace with an already built LaTeX project. Open VS Code and view the PDF file by calling `Viewer.OpenBrowser` or `Viewer.OpenTab`.

tamuratak:
> The function build calls this.extension.manager.ignorePdfFile(rootFile). I think we should do the same in Viewer.OpenBrowser and Viewer.OpenTab

> Consider a workspace with an already built LaTeX project. Open VS Code and view the PDF file by calling Viewer.OpenBrowser or Viewer.OpenTab.

I think we don't have to call `ignorePdfFile` for those cases. Until called the `build` command, the internal PDF viewer will reload if the files updated. I think it is fine.

tamuratak:
@jlelong  I re-request review by this comment.

The built-in feature to re-request review on GitHub doesn't work.

src/components/managerlib/pdfwatcher.ts
```diff
@@ -5,12 +5,19 @@ import type {Extension} from '../../main'
...
-    private readonly pdfsWatched = new Set<string>()
+    private readonly watchedPdfLocalPaths = new Set<string>()
     private readonly pdfWatcher: chokidar.FSWatcher
+    private readonly watchedPdfVirtualUris = new Set<string>()
+    private readonly ignoredPdfUris = new Set<string>()
 
     constructor(extension: Extension) {
         this.extension = extension
         this.pdfWatcher = this.initiatePdfWatcher()
+        this.initiateVirtualUriWatcher()
```
jlelong, line 16: This function returns a file system watcher but the returned values is discarded. Not sure this is an issue but it looks wieird.

jlelong:
Everything looks fine to me. Yet, I am wondering if we could go one step further

- Drop https://github.com/James-Yu/LaTeX-Workshop/blob/5d84c0f37db3814a31a047926b52ff5ff488409f/src/components/managerlib/pdfwatcher.ts#L9
- Handle all the `.pdf` in https://github.com/James-Yu/LaTeX-Workshop/blob/5d84c0f37db3814a31a047926b52ff5ff488409f/src/components/managerlib/pdfwatcher.ts#L53
- Stop calling https://github.com/James-Yu/LaTeX-Workshop/blob/c2d51873d2313daf227276fe9de53baea655ee2c/src/components/builder.ts#L366

This would make the watcher logic simpler. What do you think?

src/components/managerlib/pdfwatcher.ts
tamuratak: It is intended. At the moment, we don't have to store the returned value as a property of `PdfWatcher`. If it becomes necessary in the future, we will store the returned value.

tamuratak:
Are you proposing to stop using `chokidar` for watching changes on PDF files? `vscode.FileSystemWatcher` cannot watch files out of the workspace. PDF files might be out of it. We had better keep using `chokidar` to watch local PDF files.

jlelong:
Thanks for the explanations and the link to the corresponding issue at VS Code. If they ever implement it, we will be able to stop using `chokidar`.

jlelong (approved):

jlelong:
@tamuratak microsoft/vscode#3025 has been implemented and will be part of the January 2022 release. This should enable us to drop the `chokidar` PDF file watcher and watch all the PDF files using vscode file watcher.

tamuratak:
I will make a PR when I have time. 

- https://github.com/tamuratak/LaTeX-Workshop/blob/use_vscode_watcher/src/components/managerlib/pdfwatcher.ts

tamuratak:
I would like to postpone making the PR because 

1. The file watcher provided by VS Code does not work on WSL 
   - microsoft/vscode-remote-release/issues/5000
   - https://github.com/microsoft/vscode/issues/136894#issuecomment-966926708
2. The VS Code team plans to upgrade the version of Electron to 16. It is a big change. We had better wait to see whether the upgrade causes issues.
   - microsoft/vscode/pull/137241

PR #3038, opened 2022-01-04
Log commands and their arguments separately

### LINKED ISSUE #3037: Improve the logging of commands and theirs arguments
From https://github.com/James-Yu/LaTeX-Workshop/issues/3032#issuecomment-1003920519

We should clearly log the commands and their argument separately, see https://github.com/tamuratak/LaTeX-Workshop/commit/cf8ba0e1894f437974aadd332d53ed6722038e77

jlelong:
Close #3037

### DIFF
--- a/src/components/logger.ts
+++ b/src/components/logger.ts
@@ -23,11 +23,9 @@ export class Logger {
         }
     }
 
-    logCommand(message: string, command: string, args: string[] | undefined) {
+    logCommand(message: string, command: string, args: string[] = []) {
         this.addLogMessage(message + ': ' + command)
-        if (args) {
-            this.addLogMessage(message + ' args: ' + JSON.stringify(args))
-        }
+        this.addLogMessage(message + ' args: ' + JSON.stringify(args))
     }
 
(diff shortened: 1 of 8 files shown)


src/components/logger.ts
```diff
@@ -23,6 +23,13 @@ export class Logger {
         }
     }
 
+    logCommand(message: string, command: string, args: string[] | undefined) {
+        this.addLogMessage(message + ': ' + command)
+        if (args) {
+            this.addLogMessage(message + ' args: ' + JSON.stringify(args))
+        }
+    }
```
tamuratak, line 31: We had better always add arguments to logs as follows:

```ts
    logCommand(message: string, command: string, args: string[] = []) {
        this.addLogMessage(message + ': ' + command)
        this.addLogMessage(message + ' args: ' + JSON.stringify(args))
    }
```

It makes clear that there is no argument if it is empty, and it makes reading log messages much easier.

tamuratak (changes requested):

tamuratak (approved):

PR #3040, opened 2022-01-05
Add configuration latex.toolsEnv to allow environment customization without having to alter all tools

### LINKED ISSUE #3039: Case sensitivity in latex.tools environment variables on Windows
## Preliminary questions [Required]

yes all

The `env` object in an entry of the configuration `latex-workshop.latex.tools` will not allow case insensitivity on windows, and leads to confusion on the end user.

Set the configuration
```js
    {
        "command": "latexmk",
        ...
        "env": {
            "PATH": "foo"
        }
    }, ...
```
and build with `latexmk`.

- Expected behavior: `"foo"` is used to run `foo/latexmk`
- Actual behavior: `process.env.Path` is used instead

## Logs [Required]

Irrelevant. It's a problem with `process.env` and `cs.spawn()` on Windows.

## Desktop [Required]

 - OS: Windows 10
 - VS Code version: 1.63.2
 - Extension version: 8.23

wise0704:
Currently, if you wanted to modify, for example, the `PATH` variable for your tools, you have to copy paste the default list of tools, and then edit the `env` property of each tool.

Instead, you will be able to just set
```js
"latex-workshop.latex.toolsEnv": { "PATH": "..." },
```

This PR also:
- fixes #3039
- Allows access to environment variables in the `env` property value using the vscode style variables `${env:variable}`.


tamuratak:
> Allows access to environment variables in the env property value using the vscode style variables ${env:variable}.

Reject due to the maintenance burden.

wise0704:
> > Allows access to environment variables in the env property value using the vscode style variables ${env:variable}.
> 
> 
> 
> Reject due to the maintenance burden.
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 

How is it a maintenance burden? It's just a simple string replacement without depending on or affecting anything else in the extension.

PR #3046, opened 2022-01-08
Count words on save

### LINKED ISSUE #3041: word count in the status bar

You visited the wiki?: Yes

No

## To Reproduce  [Required]

It exists inside the menu of the LaTeX Workshop.

It would be great to include the word count of the compiled Latex Document in the status bar. 
An extension doing the same exists at https://marketplace.visualstudio.com/items?itemName=geoffkaile.latex-count

## Logs [Required]

Please paste the whole log messages here, not parts of ones. The log should start with `Initializing LaTeX Workshop`. It is very important to identify problems.

```

[Paste the log here]
```

```
To access the log, click 'help' -> 'Toggle Developer Tools' -> 'Console'. Paste anything suspicious.

[Paste the log here]
```

## Screenshots

Screenshot 2022-01-05 at 5 01 35 PM

## Desktop [Required]

 - OS: Mac OS
 - VS Code version: 1.61
 - Extension version: 8.23
 - TeX distribution version: MacTex 2021

### Are you using VSCodium?

No

No

No

If the answer is Yes, please write which one you are using. Write the versions of the remote extension.

- Remote Containers/Remote SSH/Remote WSL
- Remote Extension Version: [e.g. 0.48.0]

### Additional context

It would be easier to keep track of word limits.

jlelong:
This PR adds a configuration variable `latex-workshop.texcount.autorun` to enable `texcount`to be automatically called on the entire project on fie save.

Close #3041

### DIFF
--- a/src/components/counter.ts
+++ b/src/components/counter.ts
@@ -13,15 +13,31 @@ export class Counter {
     private autoRunInterval: number = 0
     private commandArgs: string[] = []
     private commandPath: string = ''
+    private texCountMessage: string = ''
+    private wordCount: string = ''
+    private status: vscode.StatusBarItem
 
     constructor(extension: Extension) {
         this.extension = extension
+        // gotoLine status item has priority 100.5
+        this.status = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Right, 100.6)
         this.loadConfiguration()
         vscode.workspace.onDidChangeConfiguration(e => {
             if (e.affectsConfiguration('latex-workshop.texcount') || e.affectsConfiguration('latex-workshop.docker.enabled')) {
                 this.loadConfiguration()
+                this.updateStatusVisibility()
+
             }
         })
+        this.updateStatusVisibility()
+        vscode.window.onDidChangeActiveTextEditor((e: vscode.TextEditor | undefined) => {
+            if (e && extension.manager.hasTexId(e.document.languageId)) {
+                this.updateStatusVisibility()
+            } else {
+                this.status.hide()
+            }
+        })
+
     }
 
     private loadConfiguration() {
@@ -31,6 +47,25 @@ export class Counter {
         this.commandArgs = configuration.get('texcount.args') as string[]
         this.commandPath = configuration.get('texcount.path') as string
         this.useDocker = configuration.get('docker.enabled') as boolean
+    }
+
+    private updateStatusVisibility() {
+        if (this.autoRunEnabled) {
+            this.updateWordCount()
+            this.status.show()
+        } else {
+            this.status.hide()
+        }
+    }
+
+    private updateWordCount() {
+        if (this.wordCount === '') {
+            this.status.text = ''
+            this.status.tooltip = ''
+        } else {
+            this.status.text = this.wordCount + ' words'
+            this.status.tooltip = this.texCountMessage
+        }
     }
 
     async countOnSaveIfEnabled(file: string) {
@@ -51,10 +86,19 @@ export class Counter {
             this.extension.logger.addLogMessage('Cannot find root file')
             return
         }
-        this.count(this.extension.manager.rootFile)
+        void this.runTeXCount(this.extension.manager.rootFile).then(() => {
+            this.updateWordCount()
+        })
     }
 
+
     count(file: string, merge: boolean = true) {
+        void this.runTeXCount(file, merge).then( () => {
+            void vscode.window.showInformationMessage(this.texCountMessage)
+        })
+    }
+
+    runTeXCount(file: string, merge: boolean = true): Promise<boolean> {
         let command = this.commandPath
         if (this.useDocker) {
             this.extension.logger.addLogMessage('Use Docker to invoke the command.')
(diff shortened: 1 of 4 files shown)


src/components/counter.ts
```diff
@@ -7,15 +7,28 @@ import type {Extension} from '../main'
...
 export class Counter {
     private readonly extension: Extension
+    disableCountAfterSave: boolean = false
 
     constructor(extension: Extension) {
         this.extension = extension
     }
 
-    async count(file: string, merge: boolean = true) {
-        if (this.extension.manager.rootFile !== undefined) {
```
jlelong, line 16: This condition was most likely wrong. Nevertheless, we do not need the `rootFile` inside `count`. Hence no need to mark it as `async` anymore.

src/main.ts
```diff
@@ -156,10 +156,19 @@ export function activate(context: vscode.ExtensionContext): ReturnType<typeof ge
...
-                void extension.manager.buildOnSave(e.fileName)
+            }
+            if (configuration.get('texcount.autorun') as string === 'onSave') {
+                if (extension.counter.disableCountAfterSave) {
+                    extension.logger.addLogMessage('Auto texcount is temporarily disabled during a second.')
+                } else {
+                    extension.logger.addLogMessage(`Auto texcount started on saving file: ${e.fileName}`)
+                    void extension.counter.countOnSave()
+                }
+
```
tamuratak, line 171: We had not better put everything into `src/main.ts`.

How about calling `vscode.workspace.onDidSaveTextDocument` in the constructor of `Counter`.

tamuratak (changes requested):

src/main.ts
jlelong: Thanks for pointing out `vscode.workspace.onDidChangeConfiguration`. It can be quite handy.

I prefer to have a single call to `vscode.workspace.onDidSaveTextDocument` in the whole extension to ease maintenance. In particular, a single call to `vscode.workspace.onDidSaveTextDocument` enables us to finely monitor the execution of the different actions. Yet, we could move all the tests and configuration retrieval from

https://github.com/James-Yu/LaTeX-Workshop/blob/bd2454690afd1e346ace2db71e2815105c7ce35a/src/main.ts#L164-L172

to `extension.counter.countOnSave` in order to keep the callback concise. Would that be fine with you?

tamuratak: Yes, it is fine.

jlelong: Great. I guess we should do with the same with https://github.com/James-Yu/LaTeX-Workshop/blob/7a4d03de17b0ead6784f346630c3b8bebaa0c7b0/src/main.ts#L156-L163

I will handle this later in a separate PR.

jlelong: The whole triggering logic is now implemented in `Counter` and we rely on `onDidChangeConfiguration` to read the related configuration variables.

jlelong:
https://github.com/James-Yu/LaTeX-Workshop/blob/b43b0badb3b1e8392fb70c8918c9facc7dac06dc/src/components/counter.ts#L113

produces a lot of logs. However, it gives access to the complete output of `texcount`, which contains far more than the number of words in the document. I am wondering whether we should keep these logs or not. @tamuratak Do you have any opinion on this?

tamuratak:
We don't need the log.

tamuratak:
When `latex-workshop.texcount.autorun` is `onSave`, a word counter notification pops up every time I save a document.  The behavior is very annoying. The counter should be displayed on the status bar. Otherwise, we had better remove the autorun feature.

jlelong:
When `latex-workshop.texcount.autorun` is `onSave`, the word count is now displayed in the status bar right on the left of the `Line / Column` item. It is updated every time the file is saved and only showed for files with `latex` id.

Calling wordcount from the TeX badge still produces a pop-up notification. Yet, if the word count status bar item is visible, it is updated too.

src/components/counter.ts
```diff
@@ -7,22 +7,100 @@ import type {Extension} from '../main'
...
+    private commandArgs: string[] = []
+    private commandPath: string = ''
+    private texCountMessage: string = ''
+    private wordCount: string = ''
+    private status: vscode.StatusBarItem
 
     constructor(extension: Extension) {
         this.extension = extension
+        // gotoLine status item has priority 100.5
+        this.status = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Right, 100.6)
```
tamuratak, line 23: 100.6 should be less than 100.5. It would be better the counter is at the right of gotoLine status item. With the current implementation, the counter moves unsteadily.


https://user-images.githubusercontent.com/10665499/149036419-172ecc5d-c667-4fb3-a862-1ba1ab16f6a8.mov

tamuratak (changes requested):

jlelong:
I have one last hesitation. Calling `tecount` from the TeX badge only counts words in the current text editor if it is a `latex` file. It would be natural to do the same when calling `tecount` on save. Yet, this is not really coherent with what we do for `chtex` for instance. @tamuratak what do you think?

src/components/counter.ts
jlelong: You are right. I will move the word count item.

tamuratak:
> Yet, this is not really coherent with what we do for chtex for instance. @tamuratak what do you think?

Could you elaborate more? What kind of counter's behavior are you proposing?

jlelong:
I suggest that the word counter has the same behaviour when called on save or through the TeX badge. Either count words in the whole LaTeX project, or only in the current file. For the moment, calling `texcount` from the TeX badge only counts words in the current document.

When `chktex` is called on save, it runs on the whole LaTeX project whereas when called on typing stops, it only lints the current file. 

Then, it would be better to have the word counter take into account the whole LaTeX project. We could also argue that counting words only matters in short documents, which are unlikely to be split in multiple files.

tamuratak:
Thank you for your explanation.

It is up to you. If you like tweaking the feature, I am not against additional features, although I am not their fan.

jlelong:
I have finally decided that calling `texcount` on save will only count the number of words in the current document as it does when running it from the TeX badge.

PR #3057, opened 2022-01-17
Re-add support for LaTeX-Utilities.

leoleoasd:
Hi! I'm Leo, the new maintainer for [tecosaur/LaTeX-Utilities](https://github.com/tecosaur/LaTeX-Utilities). It seems that the support for LaTeX-Utilities is removed due to your concerns about maintenance burdens and compatibility guarantees. Since I'm going to continue maintaining LaTeX-Utilities, I think it's time to re-add the support.

I've made changes to make sure that everything works fine with the latest workshop with support readded. See tecosaur/LaTeX-Utilities#348 for details.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -132,7 +132,34 @@ function registerLatexWorkshopCommands(extension: Extension) {
 
 function generateLatexWorkshopApi(extension: Extension) {
     return {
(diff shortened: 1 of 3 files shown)


tamuratak:
We are not interested in features provided by LaTeX-Utilities. We will not expose our APIs due to the maintenance burden. If you are interested in those features, please implement them independently, extracting our codes from our repository if necessary.

PR #3060, opened 2022-01-18
add texlive as default container image

praenubilus:
With empty default container image, the following error will prompt:
```bash
Unable to find image 'latexmk:latest' locally
```
This issue has been mentioned somewhere else: https://www.reddit.com/r/LaTeX/comments/r9smni/why_is_latex_in_vscode_requiring_docker_to_build/

Add `texlive/texlive` as default container image to eliminate error prompt for first time use.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1825,7 +1825,7 @@
         },
         "latex-workshop.docker.image.latex": {
           "type": "string",
-          "default": "",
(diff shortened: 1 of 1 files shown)


tamuratak:
Rejected. 

We don't set any default value to `latex-workshop.docker.image.latex` due to the maintenance burden.

PR #3069, opened 2022-01-30
Use customEditor api

ManuelHentschel:
This PR modifies the methods `openPdfInTab` etc. to open pdfs using vscode's CustomEditor api. Previously, this api was only used to show `'LaTeX Workshop PDF Viewer is opening a PDF file...'` and then opening a separately managed webview. Advantages of showing pdfs as custom editors include:
* Avoids opening the same pdf multiple times (though this is still possible if intended)
* Does not steal focus from the explorer view after opening a pdf. This can get a bit annoying when trying to e.g. click on a pdf file and then `del` to delete it.
* Pdf icon in the file tab is supported

I've tested this on windows 10 and ubuntu 18.04 using WSL. The only disadvantage I'm aware of is
* It is no longer possible to specify 'left'/'top'/'bottom' as locations for the panel. However, when opening a file from the explorer view this can be done using e.g. `open to the side`.

### DIFF
--- a/src/commander.ts
+++ b/src/commander.ts
@@ -206,12 +206,12 @@ export class Commander {
         this.extension.builder.kill()
     }
 
-    pdf(uri: vscode.Uri | undefined) {
+    pdf(uri: vscode.Uri | undefined, panelIn?: vscode.WebviewPanel) {
         this.extension.logger.addLogMessage('PDF command invoked.')
         if (uri === undefined || !uri.fsPath.endsWith('.pdf')) {
             return
         }
-        return this.extension.viewer.openPdfInTab(uri, 'current', false)
+        return this.extension.viewer.openPdfInTab(uri, panelIn)
     }
 
     synctex() {
--- a/src/components/viewer.ts
+++ b/src/components/viewer.ts
@@ -5,7 +5,6 @@ import * as cs from 'cross-spawn'
 
 import type {Extension} from '../main'
 import type {SyncTeXRecordForward} from './locator'
-import {openWebviewPanel} from '../utils/webview'
 import {getCurrentThemeLightness} from '../utils/theme'
 
 import type {ClientRequest, ServerResponse, PdfViewerState} from '../../viewer/components/protocol'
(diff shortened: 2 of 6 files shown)


tamuratak:
Rejected. Not convinced.

> Avoids opening the same pdf multiple times

The feature of opening the same pdf multiple times is what we want.

tamuratak:
See

> Controls how to open items in trees and lists using the mouse (if supported). Note that some trees and lists might choose to ignore this setting if it is not applicable.
>
> workbench.list.openMode

- https://code.visualstudio.com/docs/getstarted/settings

James-Yu:
> It is no longer possible to specify 'left'/'top'/'bottom' as locations for the panel. However, when opening a file from the explorer view this can be done using e.g. `open to the side`.

This degeneration is unfortunately not acceptable w.r.t. LW in my opinion. A neat feature to LaTeX users is to single click on the `view PDF` icon to the top right (when focusing a `.tex` file) and see the pdf opened at the correct location, defined in the configuration. Is there a workaround to retain this feature while using the customEditor API?

ManuelHentschel:
> > It is no longer possible to specify 'left'/'top'/'bottom' as locations for the panel. However, when opening a file from the explorer view this can be done using e.g. `open to the side`.
> 
> This degeneration is unfortunately not acceptable w.r.t. LW in my opinion. A neat feature to LaTeX users is to single click on the `view PDF` icon to the top right (when focusing a `.tex` file) and see the pdf opened at the correct location, defined in the configuration. Is there a workaround to retain this feature while using the customEditor API? @ManuelHentschel

Agree. I re-implemented the different opening locations on the master branch of my fork https://github.com/ManuelHentschel/LaTeX-Workshop. Unfortunately, the vscode api does not fully support opening (non-text?) documents in most of the locations that can be specified in the LW settings (and also not without taking focus), but the procedure currently used here also works when using the custom editor api.

Since I originally opened this PR, lots of things have changed, leading to lots of merge conflicts, so I did not update this branch. However, let me know if you want me to open a new PR with this modification. Also, feel free to point out/fix any problems due to my unfamiliarity with the code base.

James-Yu:
> Since I originally opened this PR, lots of things have changed, leading to lots of merge conflicts, so I did not update this branch. However, let me know if you want me to open a new PR with this modification. Also, feel free to point out/fix any problems due to my unfamiliarity with the code base.

Welcome a new PR/PRs! Thanks 😃

PR #3075, opened 2022-02-07
Add indentation to display math

JaMeZ-B:
I propose to treat display math like an environment in terms of code indentation:
```
\[
    \int_M d\omega = \int_{\partial M} \omega
\]
```

- This increases code readability, as it provides additional visual distinction between `\(...\)` and `\[...\]`
- It also resembles the output of display math to some extend
- It is the old TextMate behaviour (might help converts like me)

Personally, I would also appreciate to have this snippet trigger on `$$` (as in TextMate…), but I did not manage achieving that.

Thanks for the awesome work on this extension!

### DIFF
--- a/data/commands.json
+++ b/data/commands.json
@@ -100,7 +100,7 @@
   },
   "[": {
     "command": "[",
-    "snippet": "[${1}\\]",
+    "snippet": "[\n\t${1}\n\\]",
     "detail": "display math \\[ ... \\]"
   },
   "{": {


JaMeZ-B:
Thanks a lot for your comments – I actually had tried the `commandsJSON.replace` earlier but apparently got confused on how to modify the command using the settings ui. Next time I'll head straight into the settings file…

After adding `$` to the intellisense triggers, my `$$`-snippet also works as indented 👍

Issue #3082, opened 2022-02-14
Change settings scope from "window" (default) to "resource" for a number of settings


bersbersbers:
### Disable all the other extensions except for LaTeX Workshop, restart VS Code, and check that you can not see the requested feature. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before requesting a feature.

You visited the wiki?: Yes

### If your requested feature is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: Yes

## Is your requested feature  related to a problem? Please describe. [Required]

I have two docs in a structure like this:
```
D:\BUG
│   bug.code-workspace
│
├───doc1
│   └───.vscode
│           settings.json
│
└───doc2
```

I open both docs in one VS Code window. This is what `bug.code-workspace` is for.

In `doc1`, I would like to set a few settings that I do not want to apply to `doc2`. This is what `doc1/.vscode/settings.json` is for, and it looks like this:

```json
{
    "editor.wordWrapColumn": 88,
    "latex-workshop.latexindent.args": [
        "-c",
        "%DIR%/",
        "%TMPFILE%",
        "-y=defaultIndent: '%INDENT%'",
        "-l",
    ],
}
```

So far, so good. However, this is not working because `latex-workshop.latexindent.args` is window-scoped, while `editor.wordWrapColumn` is resource-scoped. In line with that, the setting is greyed out and not applied:

![image](https://user-images.githubusercontent.com/12128514/153915001-83c233fe-c102-4f57-b00d-16e5d49028fb.png)

The reason is that the setting is not properly scoped - it has `window` scope:
https://github.com/microsoft/vscode/blob/6136c815bc6e7b99ec2dac56dccb3869574d2dd8/src/vs/workbench/contrib/preferences/browser/preferencesRenderers.ts#L596-L603

And the reason for that is that it does not have an explicit scope at all, so it falls back to `window` - see https://code.visualstudio.com/api/references/contribution-points:
https://github.com/James-Yu/LaTeX-Workshop/blob/04f73108e439661faef2c8f85eac7baacbdb736b/package.json#L1810-L1822

This is certainly not intentional - I see no reason for this setting to (necessarily) be the same for all resources in a window. It should be able to differ by resource (folder).

## Describe the solution you'd like  [Required]

[...]

PR #3122, opened 2022-02-20
Use resource scoped properties

jlelong:
For now, this is just a POC to figure out the best way of accessing the current workspace configuration.

I think it is best to rely on the `Manager` to use the correct workspace but an alternative approach would be to use 

```ts
    private createSteps(rootFile: string, languageId: string, recipeName: string | undefined): StepCommand[] | undefined {
        let steps: StepCommand[] = []
        const configuration = vscode.workspace.getConfiguration('latex-workshop', vscode.Uri.file(rootFile))

```

 @tamuratak What do you think?


tamuratak:
The alternative approach would be better if we have `rootFile` as an argument.

jlelong:
Based on the list discussed in #3082, the following properties now have the scope `resource`

```
latex-workshop.latex.recipe.default
latex-workshop.latex.tools
latex-workshop.latex.magic.args
latex-workshop.latex.magic.bib.args
latex-workshop.latex.external.build.command
latex-workshop.latex.external.build.args
latex-workshop.latex.build.forceRecipeUsage
latex-workshop.latex.search.rootFiles.include
latex-workshop.latex.search.rootFiles.exclude
latex-workshop.latex.autoBuild.run
latex-workshop.latex.autoBuild.interval
latex-workshop.latex.autoBuild.cleanAndRetry.enabled
latex-workshop.latex.build.clearLog.everyRecipeStep.enabled
latex-workshop.latex.autoClean.run
latex-workshop.latex.rootFile.useSubFile
latex-workshop.latex.rootFile.doNotPrompt
latex-workshop.latex.clean.subfolder.enabled
latex-workshop.latex.clean.fileTypes
latex-workshop.latex.clean.command
latex-workshop.latex.clean.args
latex-workshop.latex.clean.method
latex-workshop.latex.option.maxPrintLine.enabled
latex-workshop.latex.outDir
latex-workshop.chktex.enabled
latex-workshop.chktex.run
latex-workshop.chktex.path
latex-workshop.chktex.args.active
latex-workshop.chktex.delay
latex-workshop.chktex.convertOutput.column.enabled
latex-workshop.chktex.convertOutput.column.chktexrcTabSize
latex-workshop.intellisense.file.exclude
latex-workshop.intellisense.file.base
latex-workshop.latexindent.args
```
Now, I discuss other properties that were in the list but for which the implementation is not straightforward and might not be possible

1. These values are read from the viewer

   ```
    latex-workshop.synctex.afterBuild.enabled: used by the viewer
    latex-workshop.synctex.synctexjs.enabled: used by the viewer
   ```
2. These values are obtained by subscribing to `ConfigurationChangeEvent`

   ```
   latex-workshop.texcount.path
   latex-workshop.texcount.args
   latex-workshop.texcount.autorun
   latex-workshop.texcount.interval
   ```
3. It does not make sense to change `path` properties of executables and `latex-workshop.synctex.path` is called from the viewer so it is not possible to tight it to specific workspace root. To keep things coherent, I suggest to keep `.path` properties `window`-scoped. I put in this list `latex-workshop.texdoc.args` because I see no valid reason to want to attach it to a workspace root.

   ```
   latex-workshop.latexindent.path
   latex-workshop.texdoc.path
   latex-workshop.texdoc.args
   latex-workshop.synctex.path
   ``` 
4. Finally, I discovered that `atex-workshop.chktex.args.root` is never used. We always use `latex-workshop.chktex.args.active`. Do we really need two separate settings?

In the end, I think it would great to handle `2.` but it requires some refactoring.

tamuratak:
We had better set scopes to all the configurations even if they have `"scope": "window"`.

- https://github.com/microsoft/vscode/blob/9e3d77b1e4048e6301fe709bf6fda839dbad8f2f/extensions/typescript-language-features/package.json#L141-L170

jlelong:
> We had better set scopes to all the configurations even if they have `"scope": "window"`.

Done.

jlelong:
A little summary of the current state. I will try to handle the following two cases

 - [ ] `latex-workshop.latex.recipes` but it is used to build the TeX badge during LW initialisation.  We need an update mechanism in https://github.com/James-Yu/LaTeX-Workshop/blob/766e98531671276eaacbf9854d6029e12747f58c/src/components/commander.ts#L18
 - [ ]  These values are obtained by subscribing to a `ConfigurationChangeEvent`
    ```
    latex-workshop.texcount.args
    latex-workshop.texcount.autorun
    latex-workshop.texcount.interval
    ```

tamuratak:
I am against setting the scope of `latex-workshop.latex.recipes` to `resource` due to the maintenance burden.

jlelong:
> I am against setting the scope of `latex-workshop.latex.recipes` to `resource` due to the maintenance burden.

It is actually fairly easy to implement. We can use the same pattern as in https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/providers/structure.ts. It is basically sufficient to write

```ts
class LaTeXCommander implements vscode.TreeDataProvider<LaTeXCommand> {

    private readonly _onDidChangeTreeData: vscode.EventEmitter<LaTeXCommand | undefined> = new vscode.EventEmitter<LaTeXCommand | undefined>()
    readonly onDidChangeTreeData: vscode.Event<LaTeXCommand | undefined>
    private readonly commands: LaTeXCommand[] = []

        this.extension = extension
        this.onDidChangeTreeData = this._onDidChangeTreeData.event
        this.buildCommanderTree()
    }
```

and some `refresh` handler.

Indeed I noticed #3162 while trying to understand why the `refresh` handler was not called. The maintenance burden is very reasonable and so I am in favour of implementing it. However, it can only be implemented once #3162 is fixed.

So in order to merge this PR soon, I suggest to handle the two cases listed at https://github.com/James-Yu/LaTeX-Workshop/pull/3122#issuecomment-1049247686 in other PRs.

tamuratak:
The fact that it is easy to implement doesn't imply it is easy to maintain. It can have bugs. The recipe is a core feature. It must be stable and robust.

At least, please consider adding comprehensive tests seriously.

James-Yu:
> So in order to merge this PR soon, I suggest to handle the two cases listed at [#3122 (comment)](https://github.com/James-Yu/LaTeX-Workshop/pull/3122#issuecomment-1049247686) in other PRs.

I vote for yes for the current PR commit. For the argument on whether `recipe` shall be `resource`, it might be better to open a new PR for discussion? I may need some extra time to look into TeX badge initialization, and is a bit concerned if the menu need to be updated upon folder change if `recipe` is `resource`. Maybe I'm wrong.

jlelong:
Adding tests for multi-root workspaces was indeed on my todo list. Indeed we should have done this a while ago, see  #3163. I will write some tests before merging this PR.

> For the argument on whether `recipe` shall be `resource`, it might be better to open a new PR for discussion? I may need some extra time to look into TeX badge initialization, and is a bit concerned if the menu need to be updated upon folder change if `recipe` is `resource`. Maybe I'm wrong.

Yes. I will make a new PR, once #3163 is fixed. Actually, I already made a POC of how to update the TeX badge menu when the `rootFile` changes. See https://code.visualstudio.com/api/extension-guides/tree-view#updating-tree-view-content for explanation on how to update the content of a tree view.

Issue #3162, opened 2022-02-23
File watcher and structure are wrong when changing rootFile in a multi-root workspace


jlelong:
Consider the multi-root workspace

```
event.code-workspace

A/
  .vscode/settings.json
  A.tex

B/
  B.tex
```

1. Visit the two files `A.text` and `B.tex` once. This is important to make sure the extension knows about these two files.
2. Changing from `A.tex` to `B.tex` and vice versa does not update the outline or structure view from the TeX badge. Moreover, the list of watched files remains unchanged.

We do not enter in https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/components/manager.ts#L343-L355 because  `A.tex` is the rootFile associated to workspace `A` and `B.tex` is the rootFile associated to workspace `B`. Note that `Manager.rootFile` returns the rootFile for the current workspace.

Here are the logs

```
[22:54:15] Initializing LaTeX Workshop.
[22:54:15] Extension root: /Users/jl/devel/LaTeX-Workshop
[22:54:15] $PATH: /Users/jl/.nvm/versions/node/v17.3.0/bin:/usr/local/opt/python/libexec/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/TeX/texbin:/Users/jl/bin:/Users/jl/live/teaching/ensimag/tools/chamilotools
[22:54:15] $SHELL: /bin/zsh
[22:54:15] $LANG: en_US.UTF-8
[22:54:15] $LC_ALL: undefined
[22:54:15] process.platform: darwin
[22:54:15] process.arch: x64
[22:54:15] vscode.env.appName: Visual Studio Code
[22:54:15] vscode.env.remoteName: undefined
[22:54:15] vscode.env.uiKind: 1
[22:54:15] Workspace for configuration: file:///Users/jl/devel/LaTeX-Workshop/mytests/3082/A
[22:54:15] editor.acceptSuggestionOnEnter: "off"
[22:54:15] latex-workshop.bind.enter.key: true
[22:54:15] latex-workshop.docker.enabled: false
[22:54:15] latex-workshop.docker.image.latex: ""
[22:54:15] latex-workshop.hover.preview.mathjax.extensions: []
[22:54:15] latex-workshop.intellisense.package.enabled: true
[22:54:15] latex-workshop.intellisense.update.aggressive.enabled: false
[22:54:15] latex-workshop.intellisense.update.delay: 1000
[...]

PR #3178, opened 2022-03-05
Dynamically update latex badge menus

jlelong:
This is a follow-up on #3122.

This PR updates the TeX badge when the rootFile changes.

### DIFF
--- a/src/components/commander.ts
+++ b/src/components/commander.ts
@@ -2,19 +2,21 @@ import {Extension} from 'src/main'
 import * as vscode from 'vscode'
 
 
-export class LaTeXCommandTreeView {
-    private readonly extension: Extension
-    private readonly latexCommander: LaTeXCommanderProvider
+export class LaTeXCommanderTreeView {
+    private readonly latexCommanderProvider: LaTeXCommanderProvider
 
     constructor(extension: Extension) {
-        this.extension = extension
-        this.latexCommander = this.extension.latexCommanderProvider
+        this.latexCommanderProvider = new LaTeXCommanderProvider(extension)
         vscode.window.createTreeView(
             'latex-workshop-commands',
             {
-              treeDataProvider: this.latexCommander,
+              treeDataProvider: this.latexCommanderProvider,
               showCollapseAll: true
             })
+    }
+
+    update() {
+        this.latexCommanderProvider.update()
     }
 }
 
@@ -32,11 +34,11 @@ export class LaTeXCommanderProvider implements vscode.TreeDataProvider<LaTeXComm
                 this.update()
             }
         })
-        this.commands = this.buildCommanderTree()
+        this.commands = this.buildCommandTree()
     }
 
     update() {
-        this.commands = this.buildCommanderTree()
+        this.commands = this.buildCommandTree()
         this._onDidChangeTreeData.fire(undefined)
     }
 
@@ -49,7 +51,7 @@ export class LaTeXCommanderProvider implements vscode.TreeDataProvider<LaTeXComm
         return parent
     }
 
-    private buildCommanderTree(): LaTeXCommand[] {
+    private buildCommandTree(): LaTeXCommand[] {
         const commands: LaTeXCommand[] = []
         const configuration = vscode.workspace.getConfiguration('latex-workshop', this.extension.manager.getWorkspaceFolderRootDir())
 
--- a/src/components/commander.ts
+++ b/src/components/commander.ts
@@ -22,7 +22,7 @@ export class LaTeXCommanderProvider implements vscode.TreeDataProvider<LaTeXComm
 
     private readonly _onDidChangeTreeData: vscode.EventEmitter<LaTeXCommand | undefined> = new vscode.EventEmitter<LaTeXCommand | undefined>()
     readonly onDidChangeTreeData: vscode.Event<LaTeXCommand | undefined>
-    private readonly commands: LaTeXCommand[] = []
+    private commands: LaTeXCommand[] = []
     private readonly extension: Extension
 
     constructor(extension: Extension) {
@@ -33,18 +33,15 @@ export class LaTeXCommanderProvider implements vscode.TreeDataProvider<LaTeXComm
                 this.update()
             }
         })
-        this.buildCommanderTree()
+        this.commands = this.buildCommanderTree()
     }
 
     update() {
-        // Clear commands and rebuild
-        this.commands.length = 0
-        this.buildCommanderTree()
+        this.commands = this.buildCommanderTree()
         this._onDidChangeTreeData.fire(undefined)
     }
 
     private buildNode(parent: LaTeXCommand, children: LaTeXCommand[]) {
-        this.commands.push(parent)
         if (children.length > 0) {
             parent.collapsibleState = vscode.TreeItemCollapsibleState.Collapsed
             parent.children = children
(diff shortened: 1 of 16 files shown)


src/main.ts
```diff
@@ -327,7 +327,8 @@ export class Extension {
         this.texMagician = new TeXMagician(this)
         this.envPair = new EnvPair(this)
         this.section = new Section(this)
-        this.latexCommandTreeView = new LaTeXCommandTreeView()
+        this.latexCommanderProvider = new LaTeXCommanderProvider(this)
```
tamuratak, line 330: `class Extention` should not have providers. Remove `Extension.latexCommanderProvider`.

src/components/commander.ts
```diff
@@ -15,16 +18,29 @@ export class LaTeXCommandTreeView {
     }
 }
 
-class LaTeXCommander implements vscode.TreeDataProvider<LaTeXCommand> {
-
-    private readonly commands: LaTeXCommand[] = []
+export class LaTeXCommanderProvider implements vscode.TreeDataProvider<LaTeXCommand> {
+    private readonly extension: Extension
+    private readonly _onDidChangeTreeData: vscode.EventEmitter<LaTeXCommand | undefined> = new vscode.EventEmitter<LaTeXCommand | undefined>()
+    readonly onDidChangeTreeData: vscode.Event<LaTeXCommand | undefined>
```
tamuratak, line 24: Move  `_onDidChangeTreeData` and `onDidChangeTreeData` to `class LaTeXCommandTreeView`

src/components/manager.ts
```diff
@@ -356,8 +356,8 @@ export class Manager {
                 // We need to parse the fls to discover file dependencies when defined by TeX macro
                 // It happens a lot with subfiles, https://tex.stackexchange.com/questions/289450/path-of-figures-in-different-directories-with-subfile-latex
                 await this.parseFlsFile(this.rootFile)
-                this.extension.structureProvider.refresh()
                 this.extension.structureProvider.update()
+                this.extension.latexCommanderProvider.update()
```
tamuratak, line 360: Define `LaTeXCommandTreeView.update` and call it instead of calling `LatexCommanderProvider.update` directly.

src/main.ts
```diff
@@ -292,6 +291,7 @@ export class Extension {
     readonly texMagician: TeXMagician
     readonly envPair: EnvPair
     readonly section: Section
+    readonly latexCommanderProvider: LaTeXCommanderProvider
```
tamuratak, line 294: `class Extention` should not have providers. Remove `Extension.latexCommanderProvider`.

tamuratak (changes requested):

src/components/commander.ts
jlelong: According to https://code.visualstudio.com/api/references/vscode-api#TreeDataProvider%3CT%3E, the event `onDidChangeTreeData` must be defined in the `TreeDataProvider` not in the `TreeView` and the documentation says

>  This will trigger the view to update the changed element/root and its children recursively (if shown).

I could define the `LatexCommanderProvider` directly in the `LaTeXCommandTreeView` and chain the two updates but it would not be coherent with what we do in `structure.ts` and would not really follow the logic of VS Code's API. I think it is better to follow as much as possible the same patterns in all the components.

Even though the `LatexCommanderProvider` is not really needed at the `Extension` level per se, could you expand a bit on the motivations behind 

> `class Extention` should not have providers directly

tamuratak: > it would not be coherent with what we do in structure.ts

I don't think what we do in `structure.ts` follows good practice. We had better avoid exposing implementation details. And, `SectionNodeProvider` is an implementation detail. Other components should access it through `StructureTreeView`.

In other words, it is good we have two or more classes, `SectionNodeProvider`, `StructureTreeView`, and others, to implement the structure tree feature. However, we don't have to expose both classes. Hiding implementation details increases the maintainability of the codebase. Splitting a thing into small components is good. Exposing all of them as API is not good.

We can say the same thing with `LaTeXCommandTreeView` and `LaTeXCommanderProvider`. It is good to have multiple classes to implement something. But, we don't have to expose both.

tamuratak: > According to https://code.visualstudio.com/api/references/vscode-api#TreeDataProvider%3CT%3E, the event onDidChangeTreeData must be defined in the TreeDataProvider not in the TreeView and the documentation says

Sorry. We don't have to move `onDidChangeTreeData`.

jlelong: I have refactored the implementation to stop exposing `LaTeXCommanderProvider`. There was a mix between `LaTeXCommand` and `LaTeXCommander`. Now, we use `LaTeXCommander` everywhere.

tamuratak:
I am not positive about this PR. It increases the maintenance burden. I will never investigate issues related to this feature. Only you have to deal with the related issues. If you are fine with that, I am not particularly against it.

tamuratak:
When recipes dynamically changes, what does `"latex-workshop.latex.recipe.default": "lastUsed"` imply?

jlelong:
> When recipes dynamically changes, what does `"latex-workshop.latex.recipe.default": "lastUsed"` imply?

Good catch! The value of the last used recipe is reset when the root file changes.

jlelong:
> I am not positive about this PR. It increases the maintenance burden. I will never investigate issues related to this feature. Only you have to deal with the related issues. If you are fine with that, I am not particularly against this PR.

I think it is worth giving it a try. If it turns out to create issues we did not think about, I will revert it.
@James-Yu do you have a specific opinion?

src/components/manager.ts
```diff
@@ -358,6 +358,8 @@ export class Manager {
                 await this.parseFlsFile(this.rootFile)
                 this.extension.structureProvider.refresh()
                 this.extension.structureProvider.update()
+                this.extension.builder.reset()
```
tamuratak, line 361: I don't like resetting `previouslyUsedRecipe` every time rootFile changes. It make `"latex-workshop.latex.recipe.default": "lastUsed"` useless.

tamuratak:
> I think it is worth giving it a try. If it turns out to create issues we did not think about, I will revert it.

I don't think "giving a try and reverting" is an appropriate approach because the recipe is an essential feature. If the structure view doesn't work, users can just ignore it. However, if the recipe doesn't work, it is just a disaster for users. As I said before, tests should be added to this PR at least.

src/components/manager.ts
jlelong: Right, I have remove my last commit. I am not a big fan of this `lastUsed` recipe feature. 

https://github.com/James-Yu/LaTeX-Workshop/blob/0ce09bd80e65ffd41658b7b96c917fc007c3feb5/src/components/builder.ts#L429-L446

When `"latex-workshop.latex.recipe.default": "lastUsed"` and the last used recipe name is no more in the recipe list, we peek the first recipe. I think it is a good solution and at least it will not change the current behaviour for single root workspace, which is very important.

jlelong:
> > I think it is worth giving it a try. If it turns out to create issues we did not think about, I will revert it.
> 
> I don't think "giving a try and reverting" is an appropriate approach because the recipe is an essential feature. 

Let me rephrase. If it turns out to create issues requiring too deep changes to be solved, I will revert the PR. I am very confident, that my PR does not change a thing for single rooted workspace, which must represent 99.9% of our users usage otherwise #3162 would have been reported long ago. I have added some tests.

jlelong:
Now, we have to decide if this PR deserves to be merged or not. @tamuratak, @James-Yu what do you think?

tamuratak (approved):

PR #3182, opened 2022-03-06
Test suggestion duplicates

### LINKED ISSUE #3174: `\bm` is duplicated in completion items
`\bm` is duplicated in completion items. Notice that the two `\bm` behaviors are slightly different since `\bm{math expression}` has a placeholder and the selected `\bm` doesn't.

```latex
\documentclass[12pt]{article}
\usepackage{amsmath}
\usepackage{bm}

\begin{document}
\begin{align*}
  \bm{A}
\end{align*}
\begin{align*}
  \bm
\end{align*}
\end{document}
```

jlelong:
@tamuratak Could you test this PR?

This idea behind this PR is to always work with command signatures, ie `\commandname{}{}` for a command taking two arguments. So far, arguments are not always taken into account depending on where the command comes from (a package, user defined, default commands).

Close #3174

### DIFF
--- a/src/providers/completer/command.ts
+++ b/src/providers/completer/command.ts
@@ -102,7 +102,7 @@ export class Command implements IProvider {
             }
             cmd.range = range
             suggestions.push(cmd)
-            cmdList.push(this.getCmdName(cmd, true))
+            cmdList.push(this.getCmdName(cmd))
         })
 
         // Insert unimathsymbols
@@ -115,7 +115,7 @@ export class Command implements IProvider {
             }
             this.defaultSymbols.forEach(symbol => {
                 suggestions.push(symbol)
-                cmdList.push(this.getCmdName(symbol, true))
+                cmdList.push(this.getCmdName(symbol))
             })
         }
 
@@ -144,10 +144,10 @@ export class Command implements IProvider {
             const cmds = this.extension.manager.getCachedContent(tex)?.element.command
             if (cmds !== undefined) {
                 cmds.forEach(cmd => {
-                    if (!cmdList.includes(this.getCmdName(cmd, true))) {
+                    if (!cmdList.includes(this.getCmdName(cmd))) {
                         cmd.range = range
                         suggestions.push(cmd)
-                        cmdList.push(this.getCmdName(cmd, true))
+                        cmdList.push(this.getCmdName(cmd))
                     }
                 })
             }
--- a/src/providers/completer/commandlib/commandfinder.ts
+++ b/src/providers/completer/commandlib/commandfinder.ts
@@ -60,8 +60,8 @@ export class CommandFinder {
                     label: `\\${name}`,
                     kind: vscode.CompletionItemKind.Function,
                     documentation: '`' + name + '`',
-                    insertText: new vscode.SnippetString(name + this.getArgsFromNode(node)),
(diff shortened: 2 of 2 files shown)


tamuratak:
I cannot follow the logic. The issue is duplication. Why do we have to deal with tabStops?

https://github.com/James-Yu/LaTeX-Workshop/blob/b52b76457ef97feac61c0569a6249dee4d4be917/src/providers/completer/commandlib/commandfinder.ts#L130-L136

jlelong:
The computation of duplicates is based on the array `cmdList` in https://github.com/James-Yu/LaTeX-Workshop/blob/b52b76457ef97feac61c0569a6249dee4d4be917/src/providers/completer/command.ts#L86

For every package, we store in `cmdList` the signature of the commands it provides -- signature means `name` + one `{}` per argument

https://github.com/James-Yu/LaTeX-Workshop/blob/b52b76457ef97feac61c0569a6249dee4d4be917/src/providers/completer/command.ts#L380-L389

Currently, we do not store the arguments but only the name for commands defined by the user. Hence the bug you see in #3174: `\bm{}` is defined by the `bm` package, whereas the the `\bm` entry in the suggestion list comes from the `\bm{A}`.

1. In this PR, I suggest to also consider the `name` + one `{}` per argument signature for user defined commands.
2. An alternative and more restrictive approach would be to consider as duplicate any user defined commands with a name already part of `cmdList` no matter the number of arguments.

Maybe 2. would be better

tamuratak:
> Maybe 2. would be better

I don't have specific opinion.

>  signature means name + one {} per argument

The method name `getCmdName` is misleading. It should be `getCmdSignature`. Or, we had better create `getCmdSignature` as a new method.

https://github.com/James-Yu/LaTeX-Workshop/blob/b52b76457ef97feac61c0569a6249dee4d4be917/src/providers/completer/command.ts#L97

`cmdList` should be `Set`, not `Array`. When the size of `cmdList` is large, `Array.includes` can be slow.

tamuratak:
> In this PR, I suggest to also consider the name + one {} per argument signature for user defined commands.

When you do this kind of thing, I recommend defining a new class instead of encoding everything into a string.

jlelong:
> > In this PR, I suggest to also consider the name + one {} per argument signature for user defined commands.
> 
> When you do this kind of thing, I recommend defining a new class instead of encoding everything into a string.

Could you elaborate a bit on this? Are you suggesting to define something like

```ts
class CmdSignature {
    name: string,
    args: string,
    hasOptArgs?: boolean 
}
```

The field `hasOptArgs` would be useful for https://github.com/James-Yu/LaTeX-Workshop/blob/b52b76457ef97feac61c0569a6249dee4d4be917/src/providers/completer/command.ts#L382-L384

tamuratak:
> Are you suggesting to define something like

Yes.

Please notice that I don't recommend the option 1.

jlelong:
I will take some time to properly refactor all this. For commands provided by packages, we need to consider the complete signature, not only the name because many commands exist in several different forms leading to as many different suggested items.

I will think about user defined commands in more details to weight the pros and cons of the two approaches.

jlelong:
I am closing this PR in favour of a new one to start with the refactoring.

PR #3192, opened 2022-03-11
Refactor completion to better handle duplicates

jlelong:
This PR is a new attempt following #3182.

- [x] Use `Set` instead of `array` to speed up checking if a command is already present
- [x] Handle default commands (from `commands.json`) as those provided by packages: index them with their signature (`name` + `{}` for every argument). This solves this issue of having `\section` and `\section{title}` both listed as suggestions.
- [x] #3174

### DIFF
--- a/src/providers/completer/command.ts
+++ b/src/providers/completer/command.ts
@@ -187,15 +187,17 @@ export class Command implements IProvider {
             })
         }
 
-        // Start working on commands in tex
+        // Start working on commands in tex. To avoid over populating suggestions, we do not include
+        // user defined commands, whose name matches a default command or one provided by a package
+        const cmdNameList = new Set<string>(suggestions.map(e => e.signature.name))
         this.extension.manager.getIncludedTeX().forEach(tex => {
             const cmds = this.extension.manager.getCachedContent(tex)?.element.command
             if (cmds !== undefined) {
                 cmds.forEach(cmd => {
-                    if (!cmdSignatureList.has(getCmdName(cmd))) {
+                    if (!cmdNameList.has(getCmdName(cmd))) {
                         cmd.range = range
                         suggestions.push(cmd)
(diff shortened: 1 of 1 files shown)


jlelong:
A little summary of how command duplicates is computed. Commands come from 3 sources

1. **Default commands**: they are defined in [`commands.json`](https://github.com/jlelong/LaTeX-Workshop/blob/main/data/commands.json)
2. **Commands defined by packages**: they are defined `data/packages/*_cmd.json`. We only keep commands with a signature (ie. `name` + `{}` per argument) different from any default command.
3. **User defined commands**: they are obtained by parsing the tex file. We only keep commands with a **name** (not a signature!) different from any command already listed in 1 or 2. This choice corresponds to option 2 in https://github.com/James-Yu/LaTeX-Workshop/pull/3182#issuecomment-1060259872

There are two reasons for using names in 3 instead of signatures

- We do not want to over populate the suggestion list.
- It is very unlikely that a user will define a command with the same name as an existing one but taking different arguments.

tamuratak:
Because it is not easy to review codes under `src/providers/completer` from the before of this PR, I leave only general comments. 

To fix issues, we had better add classes rather than just add lines because TypeScript is an object-oriented language. We have been adding lines to the codes under `src/providers/completer`, which has become something we are seeing now.

For example, the following lines indicates that `Suggestion` should be a class instead of an interface, and those function should be its methods.

https://github.com/James-Yu/LaTeX-Workshop/blob/63d542419c306514ba2437841f8a18968247c70e/src/providers/completer/command.ts#L38-L55

One of the advantages of defining a class is that we can keep data structures and code dealing it in one place, which increases readability and maintainability. They lack in the current codes under `src/providers/completer`.

I am not requesting changes to this PR.

tamuratak (approved):

PR #3193, opened 2022-03-12
Add EventBus

tamuratak:
@jlelong 

Add EventBus, which allows us to register listeners for events.

For example, we keep adding lines to `findRoot`. They are called when the root file changes:

https://github.com/James-Yu/LaTeX-Workshop/blob/18f09066d90e0b032d3a5430f0fb6bd98b50c1a1/src/components/manager.ts#L354-L360

It reduces the maintainability of `findRoot`. With `EventBus`, we fire the `rootfilechanged` event in `findRoot`, and register listeners for the event.

Notice that we should not overuse `EventBus`. We should not use it for a task that must finish when `findRoot` ends.

Primary use cases of `EventBus` are CI tests. With `EventBus`, we can precisely know when the status of LaTeX Workshop changes.

### DIFF
--- a/src/components/eventbus.ts
+++ b/src/components/eventbus.ts
@@ -1,0 +1,69 @@
+import {EventEmitter} from 'stream'
+import type {PdfViewerState} from 'viewer/components/protocol'
+import type {Disposable} from 'vscode'
+
+export const BuildFinished = 'buildfinished'
+export const PdfViewerPagesLoaded = 'pdfviewerpagesloaded'
+export const PdfViewerStatusChanged = 'pdfviewerstatuschanged'
+export const RootFileChanged = 'rootfilechanged'
+export const FindRootFileEnd = 'findrootfileend'
+export const SyncTexForwardEnd = 'synctexforwardend'
+
+type EventArgTypeMap = {
+    [PdfViewerStatusChanged]: PdfViewerState,
+    [RootFileChanged]: string
+}
+
+export type EventName = typeof BuildFinished
+                    | typeof PdfViewerPagesLoaded
+                    | typeof PdfViewerStatusChanged
+                    | typeof RootFileChanged
+                    | typeof FindRootFileEnd
+                    | typeof SyncTexForwardEnd
+
+export class EventBus {
+    private readonly eventEmitter = new EventEmitter()
+
+    dispose() {
+        this.eventEmitter.removeAllListeners()
+    }
+
+    fire<T extends keyof EventArgTypeMap>(eventName: T, arg: EventArgTypeMap[T]): void
+    fire(eventName: EventName): void
+    fire(eventName: EventName, arg?: any): void {
+        this.eventEmitter.emit(eventName, arg)
+    }
+
+    onDidChangeRootFile(cb: (rootFile: EventArgTypeMap['rootfilechanged']) => void): Disposable {
+        return this.registerLisneter('rootfilechanged', cb)
+    }
+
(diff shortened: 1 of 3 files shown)


src/components/eventbus.ts
```diff
@@ -0,0 +1,69 @@
+import {EventEmitter} from 'stream'
```
jlelong, line 1: I thought  `EventEmitter` was provided by `events`. See https://nodejs.org/api/events.html#class-eventemitter

src/components/eventbus.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    fire<T extends keyof EventArgTypeMap>(eventName: T, arg: EventArgTypeMap[T]): void
+    fire(eventName: EventName): void
+    fire(eventName: EventName, arg?: any): void {
+        this.eventEmitter.emit(eventName, arg)
+    }
+
+    onDidChangeRootFile(cb: (rootFile: EventArgTypeMap['rootfilechanged']) => void): Disposable {
+        return this.registerLisneter('rootfilechanged', cb)
```
jlelong, line 38: Spelling typo everywhere: `registerListener`

src/main.ts
```diff
@@ -278,6 +279,7 @@ export function activate(context: vscode.ExtensionContext): ReturnType<typeof ge
 export class Extension {
     readonly extensionRoot: string
     readonly logger: Logger
+    readonly eventBus = new EventBus()
```
jlelong, line 282: For purely cosmetic reasons, I would have put the declaration here and the definition in `constructor()`

jlelong:
> It reduces the maintainability of `findRoot`. With `EventBus`, we fire the `rootfilechanged` event in `findRoot`, and register listeners for the event.

I am not sure to understand where you want to go with this. Do you plan to turn https://github.com/James-Yu/LaTeX-Workshop/blob/4c75a5aa52e5ccc20640db8af2a04e9311b0967b/src/components/manager.ts#L361-L362 into subscriptions to `onDidChangeRootFile`?

src/main.ts
tamuratak: We had better call `new EventBus()` here, which assures that `eventBus` already exists.

tamuratak:
It is up to you. Those lines are ones for which you are responsible. So, when you used to `EventBus`, and you think it is a good idea, please use `EventBus`.

https://github.com/James-Yu/LaTeX-Workshop/blob/4c75a5aa52e5ccc20640db8af2a04e9311b0967b/src/components/manager.ts#L361-L362

As I write the above comment,  primary use cases of `EventBus` are CI tests.  In the current implementation, we have to call `await sleep(...)` everywhere in tests.  With `EventBus`, we can know when tasks of LaTeX Workshop finishes. We can write tests as follows:

```ts
doSomething()
await eventPromise
checkRealExtensionState()
```

We don't have to call `await sleep(...)`.

src/main.ts
tamuratak: You can see at `out/src/main.js`:

```js
class Extension {
    constructor() {
        this.eventBus = new eventbus_1.EventBus();
        this.extensionRoot = path.resolve(`${__dirname}/../../`);
        // We must create an instance of Logger first to enable
        // adding log messages during initialization.
        this.logger = new logger_1.Logger();
        this.addLogFundamentals();
```

jlelong:
Great! It is indeed much better for CI tests.

PR #3200, opened 2022-03-15
Test new wordPattern

jlelong:
Related to https://github.com/jlelong/vscode-latex-basics/issues/17

@tamuratak Could you test if suggestions (and snippets?) work fine in Japanese ?

This PR is just for testing and must not be merged.

### DIFF
--- a/syntax/latex-language-configuration.json
+++ b/syntax/latex-language-configuration.json
@@ -111,5 +111,8 @@
 		"decreaseIndentPattern": "^\\s*\\\\end{(?!document)"
 	},
 	"autoCloseBefore": ";:.,=}])>\\` \n\t$",
-	"wordPattern": "(__)|(\\*\\*)|(\\.\\.\\.)|([^\\s`'\"~_!?|$#@%^&*\\-=+;:,.<>(){}[\\]\\\\\\/]{2,})"
+	"wordPattern": {
+		"pattern": "(__)|(\\*\\*)|(\\.\\.\\.)|((\\p{Alphabetic}|\\p{Number}|\\p{Nonspacing_Mark}){2,})",
+		"flags": "u"
+	}
 }


tamuratak:
No, it doesn't work:

```latex
\documentclass[uplatex]{jsarticle}
\usepackage[uplatex,jis2004]{otf}
\begin{document}

森鷗外と内田百閒とが髙島屋に行くところを想像した。

葛飾区の𠮷野家

\end{document}
```

The example is from https://texwiki.texjp.org/?OTF

You have to use `uplatex` to compile it.

スクリーンショット 2022-03-17 9 19 56

Notice that we cannot see `align` and `alignat` in suggestions.

tamuratak:
Instead of tweaking wordPattern, we had better implement `DocumentHighlightProvider`:

- https://github.com/microsoft/vscode/issues/104937#issuecomment-679045490

Notice that I am against implementing `DocumentHighlightProvider` on our side. It increases the maintenance burden.

tamuratak:
> Could you test that with this PR, suggestions (word based suggestions and snippets) work fine in a Japanese document?

No, they don't work.

I don't want to be bothered to review this kind of pull requests.

We should not try to add tweaks for natural languages. They can differ more than we can imagine. I don't want to be bothered to investigate such things. We should concentrate on the LaTeX language.

When I want to add tweaks for the Japanese language, I will make a new VS Code extension independently of LaTeX Workshop. I believe it is the best approach for each natural language.

---
It is a difficult task to separate a Japanese sentence into words. There are vast researches and libraries. It implies that we should not try to add tweaks on our side.

- http://www.phontron.com/kytea/
- https://taku910.github.io/mecab/
- https://github.com/daac-tools/vaporetto

jlelong:
Fine, I won't bother you but please note that I am not just playing around. This is actually discussed https://github.com/microsoft/vscode/issues/144913#issuecomment-1066468990

> This is the current built-in word definition for markdown:
> 
> https://github.com/microsoft/vscode/blob/3e5c7e2c570a729e664253baceaf443b69e82da6/extensions/markdown-basics/language-configuration.json#L55
> 
> And this is the current built-in word definition for LaTeX:
> 
> https://github.com/microsoft/vscode/blob/3e5c7e2c570a729e664253baceaf443b69e82da6/extensions/latex/latex-language-configuration.json#L110
> 
> alexr00 Given LaTeX is similar to markdown, forwarding to you to investigate if the LaTeX word definition could be improved using some new concepts that markdown is using.

I just wanted to make sure that using the new concepts does not make the situation worse as far as Japanese documents are concerned. As you do not feel concerned, I will take care of this change myself.

PR #3207, opened 2022-03-19
Enable user replacement of `@` snippets

jlelong:
This PR adds a new setting `latex-workshop.intellisense.snippetsJSON.replace` to enable users to remove, redefine or add `@` snippets.

Close https://github.com/jlelong/LaTeX-Workshop/issues/4


tamuratak:
It might not be helpful displaying only `User defined snippet`.

abcabc

tamuratak:
I am not positive about this PR. With this PR, we will have user-defined snippets `User/snippets/latex.json`, default LaTeX-Wokshop snippets `snippets/latex.json`, and `snippetsJSON.replace`. Things will get complicated.

One possible option is to move all the snippets `snippets/latex.json` to `package.json` like the default recipes, which allow users to rewrite the snippets. There are pros and cons. I am not sure which is better.

jlelong:
> It might not be helpful displaying only `User defined snippet`.

That is true but it makes clear where the item comes from and if we want the user to provide documentation we have to consider a more complicated type for the content of `snippetsJSON.replace`. I am not sure this is worth it.


> With this PR, we will have user-defined snippets `User/snippets/latex.json`, default LaTeX-Workshop snippets `snippets/latex.json`, and `snippetsJSON.replace`.

Actually, the name `snippets` might be misleading here as we are actually defining a completion provider responding to the trigger character `@`. These were originally snippets but snippets did not show to be very stable across VS Code releases and we decided to turn them into completions. The `@`snippets are defined in `data/snippets-as-commands.json`. I am not in favour of moving the entire content of `data/snippets-as-commands.json` into `package.json`, which is already quite large. Note the mechanisms behind user-defined snippets and default LaTeX-Wokshop snippets `snippets/latex.json` are complicated independent from the one of `@` snippets. In particular, note these `@` snippets are displayed even when `editor.suggest.showSnippets` is false.

I suggest to not use the name `snippets` anymore as it does not denote what VS Code calls [snippets](https://code.visualstudio.com/docs/editor/userdefinedsnippets). What about 
- `latex-workshop.intellisense.snippets.trigger.latex` --> `latex-workshop.intellisense.at-suggestions.trigger.latex`
- `latex-workshop.intellisense.snippetsJSON.replace` --> `latex-workshop.intellisense.at-suggestionsJSON.replace`
- Remove all references to the name `snippet` in the code and use `atSuggestions` instead

tamuratak:
I am not against changing names. However, I am not sure how it would make things better.

When typing `@`, snippets in `User/snippets/latex.json` including `@` also appears in suggestions. So, even if you change the name to `at-suggestionsJSON.replace`, users and you supporting the users will have to care about commands defined in `User/snippets/latex.json`.

jlelong:
> When typing `@`, snippets in `User/snippets/latex.json` including `@` also appear in suggestions.

The same already happens with user snippets containing `\`. 

In `User/snippets/latex.json`, put

```json
"pok": {
	"prefix": "\\pok",
	"body": "\\pokCommand",
	"description": "Test"
}
```


Capture d’écran 2022-03-20 à 14 57 33

Here is the plan
- Change names to avoid confusion between snippets and suggestions (aka intellisense). I will also update the wiki following the same lines.
- Use the documentation field to display `User defined at-suggestions`.

Then, I am pretty confident on dealing with issues as it will be quite easy to know where every suggestion comes from.

tamuratak:
I see. I am not against the plan.

package.json
```diff
@@ -1748,11 +1748,20 @@
           ],
           "markdownDescription": "Additional trigger characters for intellisense of LaTeX documents. You must reload VSCode to take into account a change in this configuration."
         },
-        "latex-workshop.intellisense.snippets.trigger.latex": {
+        "latex-workshop.intellisense.atSuggestion.trigger.latex": {
```
tamuratak, line 1751: Should not change the name. We had better add `intellisense.atSuggestion.trigger.latex` as a new entry, and set `markdownDeprecationMessage` to `intellisense.snippets.trigger.latex`.

- https://code.visualstudio.com/api/references/contribution-points
-  #2654

tamuratak (changes requested):

package.json
jlelong: Thanks for reminding me of #2654. Btw, it seems that calling 
https://github.com/James-Yu/LaTeX-Workshop/blob/2bd72a13cecf79c6be4f3748945acad539cb8896/src/components/configuration.ts#L16 is useless as https://github.com/James-Yu/LaTeX-Workshop/blob/4b418f4c02913af750423226695a879c61a0a59f/src/components/configurationlib/deprecated.ts#L13 is always empty.

Maybe, it is time to add a first setting to `deprecatedConfigurations`?

tamuratak: #3209

tamuratak (approved):

tamuratak:
You had better add an entry to `CHANGELOG.md` on this change.

Issue #3214, opened 2022-03-22
Custom entry field support for citation intellisense


nerdunio:
I found `latex-workshop.intellisense.citation.label` and `latex-workshop.intellisense.citation.type` to customize the citation completion. I set the type to "browser" to see more information about the entry. This is a nice feature.

However, it would be great if I could define a list of entry fields instead of a predefined string, and there's why.

In my bib files I use custom entry fields, for example "description". This field is like the title for some of my entries. With the current selection of "authors", "title" or "bibtex key" I can't target custom fields nor two fields simultaneously. 

So I suggest to extend the label from a string to an array, just like `latex-workshop.intellisense.citation.format`:

`"latex-workshop.intellisense.citation.label" : ["title", "description"],`

Hence, the two entries

```
@book{lamport1994latex,
  title={LATEX: A Document Preparation System : User's Guide and Reference Manual},
  author={Lamport, L. and Bibby, D. and Pearson Education},
  isbn={9780201529838},
  lccn={93039691},
  series={Addison-Wesley Series on Tools},
  url={https://books.google.ch/books?id=khVUAAAAMAAJ},
  year={1994},
  publisher={Addison-Wesley}
}

@book{lamport1994latex:p31,
  description={Helpful hints},
  page={31},
}
```

would result in the following suggestions.

```
LATEX: A Document Preparation System : User's Guide and Reference Manual
Helpful hints
```

And then I can type "latex guide" or "hints" and get one of the above suggestions.

Until how I used LaTeXTools for Sublime text. There I was able to customize the completion as follows:

`"cite_panel_format": ["{title} - {description}, ({keyword})","{author_short}"],`

I've gotten so used to this feature that I can't work efficiently without it.

I really like LaTeX Workshop so far and this feature would make it even better :)

Thanks in advance!

jlelong:
I am afraid this is not so easy. The setting `latex-workshop.intellisense.citation.label` must really be a single item because otherwise the rendering may become ugly: the label is displayed on a single line. 

Using the `inline` mode, you can display all the fields you want

<img width="792" alt="Capture d’écran 2022-03-22 à 17 46 36" src="https://user-images.githubusercontent.com/2910140/159531854-836cf38d-7c36-4b03-8fcb-a8cac21ef7f0.png">

I have just added `description` to `latex-workshop.intellisense.citation.format`. I hope that helps.

nerdunio:
Thank you for your fast reply!

[...]

PR #3216, opened 2022-03-23
Use keys from `latex-workshop.intellisense.citation.format` to filter text in citation intellisense

jlelong:
This a POC related to #3214 

In browser mode, display all the fields from `latex-workshop.intellisense.citation.format` on the second line. Order  in the array matters. In browser and inline mode, narrow down the search using all the fields listed in `latex-workshop.intellisense.citation.format`

Capture d’écran 2022-03-23 à 07 31 12

Capture d’écran 2022-03-23 à 07 30 03

Capture d’écran 2022-03-23 à 07 30 10

### DIFF
--- a/package.json
+++ b/package.json
@@ -1727,11 +1727,11 @@
             "authors"
           ],
           "default": "bibtex key",
-          "markdownDescription": "Defines what to show as suggestion labels when intellisense provides citation suggestions.",
+          "markdownDescription": "Defines what to show as suggestion labels when intellisense provides citation suggestions in inline mode.",
           "enumDescriptions": [
-            "Show bibtex keys in the inline intellisense.",
-            "Show publication titles in the inline intellisense.",
-            "Show publication authors in the inline intellisense."
+            "Show bibtex keys in the inline mode.",
+            "Show publication titles in the inline mode.",
+            "Show publication authors in the inline mode."
           ]
         },
         "latex-workshop.intellisense.citation.format": {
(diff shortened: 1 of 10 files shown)


James-Yu:
It works as intended and quite good on my side. Now the expanded hints show formatted raw data than a dirty-ly processed `publication`, which I think is better!

jlelong:
From https://github.com/James-Yu/LaTeX-Workshop/issues/3214#issuecomment-1076222351

> Using the format as secondary display in browser mode is a great idea. One thing to note is that the browser now shows the title field twice.

Just to keep track that I need to change this.

jlelong:
I think the PR is ready to be merged. 
@nerdunio could you test it?

nerdunio:
@jlelong I tested the new feature and it works perfectly both in `browser` and inline `mode`. I added my remaining custom entries to the format array and it was working as expected.

Thank you!

jlelong:
Close #3214

51616:
Hey, thanks for such a great VSCode extension!

Is it possible to add `bibtex key` as an option to `latex-workshop.intellisense.citation.format`? So, instead of making a separate "description" field, we could use the bib entry key as a "hint" directly.

51616:
Is it possible to also add `bibtex key` to `latex-workshop.intellisense.citation.format` as well? The browser mode seems to filter the bib entry keys but the inline does not even if I add `bibtex key` to `latex-workshop.intellisense.citation.format`.

jlelong:
I cannot reproduce the issue. The `bibtex key` is always added to the text used yo filter citation suggestions.
Please open a new issue and give a minimal working example to reproduce the issue.

PR #3220, opened 2022-03-30
Implement Structure view for bibtex

### LINKED ISSUE #3215: Support Document symbol for BibTeX file
I was wondering if you could consider supporting Document Symbol (Ctrl+Shift+.) feature for BibTex files. Having structural information would be greatly appreciated.

James-Yu:
This PR resolves #3215.

The implementation is primarily a simplified version of LaTeX structure. Thanks to the fabulous `latex-utensils` package by @tamuratak , parsing bibtex was never as easy.

One thing to note: I have check a lot of ways to dynamically hide/show the new bibtex structure view, with no good. Neither `when` clause with `editorLangId` in `contribution` nor programmatically setting `visible` to `false` work (indeed the attribute is read-only). I am not sure whether the current always-visible view is acceptable.


jlelong:
Great! What about reusing the document structure view from the TeX pane. I guess it would be enough to have a single `StructureTreeView` with two `_treeDataProvider`. Only one structure view is needed at a time in the TeX pane. 

It would also be great to have in the outline view of the explorer pane (see https://github.com/James-Yu/LaTeX-Workshop/blob/e625d6fabb0e8bf8ee8e1934a9bee949f22e5d3f/src/providers/docsymbol.ts).

jooyoungseo:
@James-Yu, this works nicely on my end. Thanks!

src/providers/structure.ts
```diff
@@ -40,8 +46,50 @@ export class SectionNodeProvider implements vscode.TreeDataProvider<Section> {
...
-        this._onDidChangeTreeData.fire(undefined)
+        this.refresh().finally(() => {this._onDidChangeTreeData.fire(undefined)})
+    }
+
+    async parseBibTeX() {
+        const document = vscode.window.activeTextEditor?.document
+        if (!document) {
+            return
+        }
+        const ast = await this.extension.pegParser.parseBibtex(document?.getText() || '').catch((error) => {
```
jlelong, line 57: No need to use `document?.` as we already check that `document` is well defined above. I guess calling `this.extension.pegParser.parseBibtex(document.getText())` would just be fine.

jlelong:
Even if this is not directly related to this PR,  I think we had better catch errors in all `parse` functions or in none to be able to use the same approach for all parsing in LW. 

https://github.com/James-Yu/LaTeX-Workshop/blob/74a5ae3552872fdcdb0d853477216f55d0bf4bbc/src/components/parser/syntax.ts#L30-L40


In https://github.com/James-Yu/LaTeX-Workshop/blob/18f4e61742ed57f98a336641adf8d7caeaa2b64d/src/providers/structure.ts#L57 you have to catch the errors while there is not need in https://github.com/James-Yu/LaTeX-Workshop/blob/74a5ae3552872fdcdb0d853477216f55d0bf4bbc/src/components/manager.ts#L954

src/providers/structure.ts
```diff
@@ -40,8 +46,50 @@ export class SectionNodeProvider implements vscode.TreeDataProvider<Section> {
     }
 
     update() {
-        this.refresh()
-        this._onDidChangeTreeData.fire(undefined)
+        this.refresh().finally(() => {this._onDidChangeTreeData.fire(undefined)})
+    }
+
+    async parseBibTeX() {
```
jlelong, line 52: My 2c suggestion. Maybe renaming this function to something like `buildBibtexModel` would be more coherent with the `buildModel` method (which could become `buildLatexModel`). I guess it would make the symmetry clearer between computing structures for bibtex and latex even if the fact this function is `async` makes his calls a bit different.

src/providers/structure.ts
```diff
@@ -30,31 +31,71 @@ export class SectionNodeProvider implements vscode.TreeDataProvider<Section> {
...
-        this._onDidChangeTreeData.fire(undefined)
+        this.refresh().finally(() => {this._onDidChangeTreeData.fire(undefined)})
+    }
+
+    async buildBibTeXModel() {
+        const document = vscode.window.activeTextEditor?.document
+        if (!document) {
+            return
+        }
+        const ast = await this.extension.pegParser.parseBibtex(document.getText())
```
jlelong, line 57: Don't we actually need to change https://github.com/James-Yu/LaTeX-Workshop/blob/74a5ae3552872fdcdb0d853477216f55d0bf4bbc/src/components/parser/syntax.ts#L38-L40 into

```ts
async parseBibtex(s: string, options?: bibtexParser.ParserOptions): Promise<bibtexParser.BibtexAst | undefined> {
    return (await this.proxy).parseBibtex(s, options).timeout(30000).catch(() => undefined)
}
```

and check `ast` as

```ts
if (!ast) {
    this.extension.logger.addLogMessage(`Cannot parse a BibTeX file: ${document.filename}`)
    this.ds = []
    return
}

Comment (James-Yu) [reply src/providers/structure.ts:57]: Let me have a check. In fact I did not read into `syntax.ts`.

**Comment by James-Yu** [comment]
One more thing: Maybe the parser should respect `intellisense.citation.maxfilesizeMB`. If so, the name of this config may not be appropriate and need rename. Will look into that.

**Comment by jlelong** [comment]
You are right, the structure should not be computed if the file is too large. It looks fine to use the same setting as for intellisense, which should probably be renamed. 

Ideally, we should also investigate if we could parse the bib files only once for computing both the structure and the intellisense suggestions

https://github.com/James-Yu/LaTeX-Workshop/blob/a160de34162f050b369f77b793060829f8d68474/src/providers/completer/citation.ts#L251-L257

**Comment by James-Yu** [comment]
> Ideally, we should also investigate if we could parse the bib files only once for computing both the structure and the intellisense suggestions. This clearly goes beyond the scope of this PR.

Great idea! To do list for record: 1. look into `buildLaTeXModel` (refactor) 2. Lazy parsing.

**Comment by James-Yu** [comment]
Looks like this PR is (almost) complete 😄

File: src/components/parser/syntax.ts [hunk 4]
```diff
@@ -31,12 +31,12 @@ export class UtensilsParser {
...
-        return (await this.proxy).parseLatexPreamble(s).timeout(500)
+    async parseLatexPreamble(s: string): Promise<latexParser.AstPreamble | undefined> {
+        return (await this.proxy).parseLatexPreamble(s).timeout(500).catch(() => undefined)
     }
 
-    async parseBibtex(s: string, options?: bibtexParser.ParserOptions): Promise<bibtexParser.BibtexAst> {
-        return (await this.proxy).parseBibtex(s, options).timeout(30000)
+    async parseBibtex(s: string, options?: bibtexParser.ParserOptions): Promise<bibtexParser.BibtexAst | undefined> {
+        return (await this.proxy).parseBibtex(s, options).timeout(30000).catch(() => undefined)
     }
```
jlelong, line 40: All in all, I am not sure about the best approach here. I may have pushed you into a slippery slope. Ensuring the same behaviour for the three functions is good but I am hesitating on the best approach 

- Catch any errors directly in the parse functinos and return `undefined`. This is easy to use but we loose the reason of the error and in particular the position at which it occured
- Catch the errors every time we use these functions. The code is more complicated but we get the exact position and reason of the problem.

In both cases, this requires to look at the places these functions are used and make the proper adjustements. For instance, look at https://github.com/James-Yu/LaTeX-Workshop/blob/fed20ebf4eca341f61fac83d82ca5c0dce7ef82f/src/providers/completer/citation.ts#L251-L258

Calling `catch` is useless as errors are already catched in `parseBibtex`. Maybe, we had rather replace `ast?.content` by
```ts
if (!ast) {
  return
}
ast.content
```

jlelong:
> > Ideally, we should also investigate if we could parse the bib files only once for computing both the structure and the intellisense suggestions. This clearly goes beyond the scope of this PR.
> 
> Great idea! To do list for record: 1. look into `buildLaTeXModel` (refactor) 2. Lazy parsing.

It would be great to list somewhere the things we would like to improve. Any suggestion on where to put this?

James-Yu:
It seems that removing the catches (which return undefined) after parsing does not yield any error, since they will be throwed at `await`. I made according changes.

> It would be great to list somewhere the things we would like to improve. Any suggestion on where to put this?
That's a quick note. We do need a some kind of todo list. Not sure creating issues + assign may suffice? A dedicated issue or even page seems a bit overwhelming, yet I agree that random notes in random pages are root of evil 🤣

jlelong:
> It seems that removing the catches (which return undefined) after parsing does not yield any error, since they will be thrown at `await`. I made according changes.

If we do not catch the errors, they appear in the _Developper Tools Console_  and a pop up error is raised in VS Code. I think it is better to catch errors and log them along with their position in the input files, don't you?

src/providers/structure.ts
```diff
@@ -30,31 +31,76 @@ export class SectionNodeProvider implements vscode.TreeDataProvider<Section> {
...
+        }
+        this.ds = []
+
+        const configuration = vscode.workspace.getConfiguration('latex-workshop', vscode.Uri.file(document.fileName))
+        if (document.getText().length >= (configuration.get('bibtex.maxFileSize') as number) * 1024 * 1024) {
+            this.extension.logger.addLogMessage(`Bib file is too large, ignoring it: ${document.fileName}`)
+            return
+        }
+        const ast = await this.extension.pegParser.parseBibtex(document.getText())
+        this.extension.logger.addLogMessage(`Cannot parse BibTeX file: ${document.fileName}`)
```
jlelong, line 65: This line should be in a catch

jlelong:
I have just reverted the changes to `parseLatex` to add `.catch(() => undefined)` back.  When `parserLatex` is called, we actually test its return value and perform specific actions when it is `undefined`.

jlelong (approved):

jlelong:
> That's a quick note. We do need a some kind of todo list. Not sure creating issues + assign may suffice? A dedicated issue or even page seems a bit overwhelming, yet I agree that random notes in random pages are root of evil 🤣

We can try to create one issue per todo item with the `dev` + `enhancement` labels. It has the advantage to allow separate discussions on every item. Let us try this.

PR #3225, opened 2022-04-01
Refractor the file content caching system

James-Yu:
This PR aims at completely rewrite the current content caching system, including latex and bibtex file watching, content caching, and API normalization. A side objective is to make the whole caching logic more clear. If possible, the fabulous `latex-utensils` will be used whenever appropriate, and I plan to reduce the non-`utensils` code as much as possible.

This is a work-in-progress. The following to-do list will be regularly updated whenever I remember something.
- [ ] LaTeX source file watching
- [ ] LaTeX source file caching
- [ ] BiBTeX source file watching
- [x] BiBTeX source file caching
  - [x] Caching
  - [x] Intellisense
  - [x] Structure
  - [ ] Auto-build
- [ ] Other project-related source file watching


tamuratak:
@James-Yu  it is not a good practice to put everything into one big pull request. You had better split this PR into small ones.

I am not positive about this PR. I am not sure whether this PR improves the maintainability.

@jlelong, if you are not against merging this PR, please notice that "looks good to me" is not enough. You must fully understand the codebase since you have to maintain it when James-Yu becomes inactive again. This PR is not an auxiliary feature that we can easily remove if it becomes unmaintained.

James-Yu:
After very extensive experiments, I found that the ast-based caching system is far slower than the current regex-based system for real-time doc symbol and structure parsing. Therefore I'd discard this PR.

James-Yu:
> 1. caching is actually a bad idea and we would rather re-parse every we need
> 2. the ast based parser is much slower that the regex-based parser for computing the symbols and structure of a project.

1. Caching is a good thing, though the current re-parse is in fact less computation hungry than I thought.
2. Yes, the parser cannot meet the demand. Though that can be just ~1 second, it is obvious in cases such as "build project then refresh outline". It is also vulnerable to docs with a lot of Latex syntax errors.

That being said, I will keep on thinking of this. @tamuratak made a good point on the maintainability, I may not make fundamental engineering changes in one go. That is indeed terrible to project management.

jlelong:
So, in terms of performance it is not good to use the ast parser to compute the structure. Yet, it would ease maintenance. 

I will keep thinking about this as it would improve the efficiency to compute the ast nodes only once on file save and then use it for structure and intellisense.

Anyway, VS Code tends to be a bit greedy with computational resources.

PR #3237, opened 2022-04-05
Add the feature to use xelatex on docker.

yanggl29:
Functions such as xelatex, pdflatex, etc. are not available when using docker. Added the above features to src/components/build.ts.


tamuratak:
Rejected due to the maintenance burden.

Please use VS Code Remote Containers

- https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers

PR #3252, opened 2022-04-16
Refactor structure to add caching

jlelong:
This PR caches the last latex and bibtex structures. 

Prior to this PR, every time an editor was revealed or saved, the structure was computed twice

- https://github.com/James-Yu/LaTeX-Workshop/blob/87666b1fe8194a398b02ad8cee9c397d0fa78fee/src/main.ts#L167
- https://github.com/James-Yu/LaTeX-Workshop/blob/6970e5d3a29b9328e2070ed5c91cfef0291f5e03/src/providers/structure.ts#L226 because a `onDidChangeTreeData` event is fired

Worse than that, when switching latex files inside a project, the structure of the whole project is recomputed twice

https://github.com/James-Yu/LaTeX-Workshop/blob/87666b1fe8194a398b02ad8cee9c397d0fa78fee/src/main.ts#L219-L223

The next step (in a separate PR) will be to investigate the use of `ast` to compute the structure of LaTeX projects.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -164,12 +164,13 @@ export function activate(context: vscode.ExtensionContext): ReturnType<typeof ge
             extension.logger.addLogMessage(`onDidSaveTextDocument triggered: ${e.uri.toString(true)}`)
             extension.manager.updateCachedContent(e)
             extension.linter.lintRootFileIfEnabled()
-            extension.structureViewer.update()
+            // Make sure to update cacheContent before calling structureViewer.computeTreeStructure
+            void extension.structureViewer.computeTreeStructure()
             void extension.manager.buildOnSaveIfEnabled(e.fileName)
             extension.counter.countOnSaveIfEnabled(e.fileName)
         }
         if (e.languageId === 'bibtex') {
-            extension.structureViewer.update()
+            void extension.structureViewer.computeTreeStructure()
         }
     }))
 
@@ -216,12 +217,6 @@ export function activate(context: vscode.ExtensionContext): ReturnType<typeof ge
     context.subscriptions.push(vscode.window.onDidChangeActiveTextEditor(async (e: vscode.TextEditor | undefined) => {
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
 
-        // We do not check langid of editor because the bibtex structure should
-        // reveal only when active editor is bibtex. This is done in the update
-        // function
-        // TODO: lazy loading for LaTeX structure?
-        extension.structureViewer.update()
-
         if (vscode.window.visibleTextEditors.filter(editor => extension.manager.hasTexId(editor.document.languageId)).length > 0) {
             extension.logger.status.show()
             if (configuration.get('view.autoFocus.enabled') && !isLaTeXActive) {
@@ -236,11 +231,14 @@ export function activate(context: vscode.ExtensionContext): ReturnType<typeof ge
         }
         if (e && extension.manager.hasTexId(e.document.languageId)) {
             await extension.manager.findRoot()
+            // Make sure to wait for findRoot to finish before calling refreshView
+            await extension.structureViewer.refreshView()
             extension.linter.lintRootFileIfEnabled()
+        } else if (e && extension.manager.hasBibtexId(e.document.languageId)) {
+            await extension.structureViewer.refreshView()
         } else {
             isLaTeXActive = false
         }
-
     }))
 
     context.subscriptions.push(vscode.window.onDidChangeTextEditorSelection((e: vscode.TextEditorSelectionChangeEvent) => {
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -307,7 +307,7 @@ export class Manager {
     /**
      * Returns `true` if the language of `id` is one of supported languages.
      *
-     * @param id The identifier of language.
+     * @param id The language identifier
      */
     hasTexId(id: string) {
         return ['tex', 'latex', 'latex-expl3', 'doctex', 'jlweave', 'rsweave'].includes(id)
@@ -316,7 +316,7 @@ export class Manager {
     /**
      * Returns `true` if the language of `id` is bibtex
      *
-     * @param id The identifier of language.
+     * @param id The language identifier
      */
     hasBibtexId(id: string) {
         return id === 'bibtex'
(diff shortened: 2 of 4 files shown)


src/main.ts
```diff
@@ -236,11 +230,13 @@ export function activate(context: vscode.ExtensionContext): ReturnType<typeof ge
         }
         if (e && extension.manager.hasTexId(e.document.languageId)) {
             await extension.manager.findRoot()
+            await extension.structureViewer.refreshView()
             extension.linter.lintRootFileIfEnabled()
+        } else if (e && extension.manager.hasBibtexId(e.document.languageId)) {
+            await extension.structureViewer.refreshView()
```
tamuratak, line 236: We had not better add more lines to `main.ts` if not necessary. How about calling `onDidSaveTextDocument` and `onDidChangeActiveTextEditor` in the constructor of `src/providers/structure.ts` and registering callbacks?

tamuratak (changes requested):

src/main.ts
```diff
@@ -164,12 +164,12 @@ export function activate(context: vscode.ExtensionContext): ReturnType<typeof ge
...
             extension.manager.updateCachedContent(e)
             extension.linter.lintRootFileIfEnabled()
-            extension.structureViewer.update()
+            void extension.structureViewer.update()
             void extension.manager.buildOnSaveIfEnabled(e.fileName)
             extension.counter.countOnSaveIfEnabled(e.fileName)
         }
         if (e.languageId === 'bibtex') {
-            extension.structureViewer.update()
+            void extension.structureViewer.update()
```
tamuratak, line 172: How about moving the line to `src/providers/structure.ts`?

src/main.ts
```diff
@@ -164,12 +164,12 @@ export function activate(context: vscode.ExtensionContext): ReturnType<typeof ge
             extension.logger.addLogMessage(`onDidSaveTextDocument triggered: ${e.uri.toString(true)}`)
             extension.manager.updateCachedContent(e)
             extension.linter.lintRootFileIfEnabled()
-            extension.structureViewer.update()
+            void extension.structureViewer.update()
```
tamuratak, line 167: How about moving the line to src/providers/structure.ts?

tamuratak (changes requested):

src/main.ts
jlelong: What is the general idea behind not adding more lines to `main.ts`? I agree that keeping functions short is a good practice but in this case I am not convinced that spreading subscriptions to `onDidSaveTextDocument` or `onDidChangeActiveTextEditor` in several places will really help.

The `hasBibTeX` part could be moved to a call to `onDidChangeActiveTextEditor` in `src/providers/structure.ts`. In the `hasTexId` part, refreshing the viewer must occur after `findRoot` has completed. It could be moved to `findRoot` but I think that not handling the structure refresh at the same place for LaTeX and BibTeX files will make maintenance harder.

src/main.ts
jlelong: It is possible to register this line in a call to `onDidSaveTextDocument` in `src/providers/structure.ts` but then it does not really make sense to log `onDidSaveTextDocument triggered`, does it?

src/main.ts
tamuratak: If we put everything into `main.ts`, we must always consider and remember all the things when we edit `main.ts`. Even if we can easily remember the reason of the conditions now, we would not be able to do a few weeks later.

tamuratak: > In the hasTexId part, refreshing the viewer must occur after findRoot has completed.

I have missed that. It is not easy to see the dependency. At least, you should write it in a comment.

tamuratak: Apart from the above discussion, the change of the active text editor does not necessarily mean the change of the rootfile. Should we really always call `structureViewer.refreshView()`?

src/main.ts
jlelong: Actually this is not so easy to move the line to an other call to `onDidSaveTextDocument` because `structureViewer.update` relies on the value of the `CachedContent`. We must make sure `extension.manager.updateCachedContent(e)` has finished first.

src/main.ts
jlelong: This is indeed the whole point of the PR. `structureViewer.refreshView` calls `SectionNodeProvider.build(force=false)`

https://github.com/James-Yu/LaTeX-Workshop/blob/2fe4d967aa7f07e58d6d1f5cee06f187f14eb754/src/providers/structure.ts#L57-L60

The structure is really rebuilt only when the `rootFile` changes

https://github.com/James-Yu/LaTeX-Workshop/blob/2fe4d967aa7f07e58d6d1f5cee06f187f14eb754/src/components/manager.ts#L377

src/main.ts
tamuratak: We have an implicit dependency here. It would increase the maintenance burden. One of the ways to make the dependency explicit is using `EventBus`. `updateCachedContent` fires an event, and the structure component calls `update` on the event.

At least we had better write a comment on the dependency.

src/main.ts
tamuratak: I must say it is not easy to see the difference between`update` and `refreshView` from the names. The should have much more different names.

src/main.ts
jlelong: I agree that we had better use `EventBus` to make dependency explicit.
I am afraid there are a lot of such dependencies in LW. This should be handled in a separate PR.

src/main.ts
jlelong: Any suggestions?  What about `updateData` and `updateView`?

tamuratak: They should not have similar names. They should not have the same word. How about `constructTreeStructure` and `refreshView`

src/main.ts
```diff
@@ -164,12 +164,13 @@ export function activate(context: vscode.ExtensionContext): ReturnType<typeof ge
             extension.logger.addLogMessage(`onDidSaveTextDocument triggered: ${e.uri.toString(true)}`)
             extension.manager.updateCachedContent(e)
             extension.linter.lintRootFileIfEnabled()
-            extension.structureViewer.update()
+            // Make sure to update cacheContent before calling structureViewer.update
```
tamuratak, line 167: > before calling structureViewer.update

Change `update` to `computeTreeStructure`

jlelong: VS Code rename feature does not act on comments 😂

tamuratak (approved):

PR #3260, opened 2022-04-21
Use eventbus to manage structure updates

jlelong:
Related to https://github.com/James-Yu/LaTeX-Workshop/pull/3252#discussion_r852681306

- the fucntion `updateCachedContent` fires an `EventBus.CachedContentUpdated`
- the structure provider relies on `EventBus` to properly update the LaTeX project structure (either the data tree or only the view).

### DIFF
--- a/src/providers/structure.ts
+++ b/src/providers/structure.ts
@@ -535,18 +535,12 @@ export class StructureTreeView {
         })
 
         vscode.workspace.onDidSaveTextDocument( (e: vscode.TextDocument) => {
-            if (extension.lwfs.isVirtualUri(e.uri)){
-                return
-            }
             if (extension.manager.hasBibtexId(e.languageId)) {
                 void extension.structureViewer.computeTreeStructure()
             }
         })
 
         vscode.window.onDidChangeActiveTextEditor((e: vscode.TextEditor | undefined) => {
-            if (e && extension.lwfs.isVirtualUri(e.document.uri)){
-                return
-            }
             if (e && extension.manager.hasBibtexId(e.document.languageId)) {
                 void extension.structureViewer.refreshView()
             }
--- a/src/providers/structure.ts
+++ b/src/providers/structure.ts
@@ -550,7 +550,7 @@ export class StructureTreeView {
             void this.computeTreeStructure()
         })
 
-        this.extension.eventBus.onDidChangeRootFile((_rootFile: string) => {
+        this.extension.eventBus.onDidChangeRootFile(() => {
             void this.computeTreeStructure()
         })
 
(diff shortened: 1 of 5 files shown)


src/providers/structure.ts
```diff
@@ -521,17 +520,49 @@ class StructureModel {
...
         this._treeDataProvider = new SectionNodeProvider(extension)
         this._viewer = vscode.window.createTreeView('latex-workshop-structure', { treeDataProvider: this._treeDataProvider, showCollapseAll: true })
         vscode.commands.registerCommand('latex-workshop.structure-toggle-follow-cursor', () => {
            this._followCursor = ! this._followCursor
         })
+
+        vscode.workspace.onDidSaveTextDocument( (e: vscode.TextDocument) => {
+            if (extension.lwfs.isVirtualUri(e.uri)){
+                return
+            }
```
tamuratak, line 540: Apart from this PR, we can make the structure view work on vitrualworkspaces since the feature does not depend on external commands like `pdflatex`, `chktex`, and `latexindent`.

src/providers/structure.ts
```diff
@@ -521,17 +520,49 @@ class StructureModel {
...
+            if (e && extension.manager.hasBibtexId(e.document.languageId)) {
+                void extension.structureViewer.refreshView()
+            }
+        })
+
+        this.extension.eventBus.onDidUpdateCachedContent(() => {
+            void this.computeTreeStructure()
+        })
+
+        this.extension.eventBus.onDidChangeRootFile((_rootFile: string) => {
```
tamuratak, line 559: You can remove the argument `_rootFile: string`

tamuratak (changes requested):

tamuratak:
I am not sure it works if `computeTreeStructure` is called again while `computeTreeStructure` is being executed. Can it happen with users' quick operation?

jlelong:
The function `computeTreeStructure` may yield to two concurrent calls of `buildBibTeXModel` or `buildLaTeXModel` but this is not an issue as these functions return a new tree structure and do not work directly on `this.ds`, see 946522a383095948563945c384c4081b759a2d2f

tamuratak:
aca36d54c49b2fe75077093b2cb76648141715f3

I am not saying that we can support vitrualworkspaces by just removing these lines.

jlelong:
> [aca36d5](https://github.com/James-Yu/LaTeX-Workshop/commit/aca36d54c49b2fe75077093b2cb76648141715f3)
> 
> Please notice that I am not saying that we can support vitrualworkspaces by just removing these lines.

Actually, I think we can. These two functions https://github.com/jlelong/LaTeX-Workshop/blob/a1913ee97eb8022fedf230fd324e16f74bf549b7/src/providers/structure.ts#L532-L542 only handle the case of a current text editor with `bibtex` language. The function `buildBibTeXModel` only accesses the editor content using `.getText()`. So as far as I understand, everything is fine.

Note that this does make the LaTeX structure work on virtual workspaces. I do not intend to support it.

tamuratak:
> I am not saying that we can support vitrualworkspaces by just removing these lines.

I see. Thank you.

tamuratak (approved):

PR #3266, opened 2022-04-24
Turn the `Suggestion` interface into a class

jlelong:
Related to https://github.com/James-Yu/LaTeX-Workshop/pull/3192#issuecomment-1066195082

This is the first PR of a series to improve the maintenance of the code under `src/providers/completer`.

### DIFF
--- a/src/providers/completer/command.ts
+++ b/src/providers/completer/command.ts
@@ -22,7 +22,7 @@ export interface CmdItemEntry {
 
 export interface CmdSignature {
     name: string, // name without leading `\`
-    args: string // ({}*[]*)*
+    args: string // {} for mandatory args and [] for optional args
 }
 
 function isCmdItemEntry(obj: any): obj is CmdItemEntry {
@@ -61,15 +61,15 @@ export class Suggestion extends vscode.CompletionItem implements ILwCompletionIt
 
     /**
      * Return the signature, ie the name + {} for mandatory arguments + [] for optional arguments.
-     * The leading backward slash is not part of the signature;
+     * The leading backward slash is not part of the signature
      */
-    cmdSignature(): string {
+    cmdSignatureAsString(): string {
         return this.signature.name + this.signature.args
     }
 
     /**
      * Return the name without the arguments
-     * The leading backward slash is not part of the signature;
+     * The leading backward slash is not part of the signature
      */
     cmdName(): string {
         return this.signature.name
@@ -154,7 +154,7 @@ export class Command implements IProvider {
             }
             cmd.range = range
             suggestions.push(cmd)
-            cmdSignatureList.add(cmd.cmdSignature())
+            cmdSignatureList.add(cmd.cmdSignatureAsString())
         })
 
         // Insert unimathsymbols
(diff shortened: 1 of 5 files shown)


src/providers/completer/command.ts
```diff
@@ -71,9 +47,37 @@ export function splitSignatureString(signature: string): CmdSignature {
...
+        this.label = label
+        this.package = pkg
+        this.signature = signature
+    }
+
+    /**
+     * Return the signature, ie the name + {} for mandatory arguments + [] for optional arguments.
+     * The leading backward slash is not part of the signature
+     */
+    cmdSignatureAsString(): string {
```
tamuratak, line 66: The prefix `cmd` seems not necessary.

src/providers/completer/command.ts
```diff
@@ -71,9 +47,37 @@ export function splitSignatureString(signature: string): CmdSignature {
...
+        return this.signature.name + this.signature.args
+    }
+
+    /**
+     * Return the name without the arguments
+     * The leading backward slash is not part of the signature
+     */
+    cmdName(): string {
```
tamuratak, line 74: The prefix `cmd` seems not necessary.

src/providers/completer/command.ts
```diff
@@ -71,9 +47,37 @@ export function splitSignatureString(signature: string): CmdSignature {
...
+        return this.signature.name
+    }
+
+    cmdHasOptionalArgs(): boolean {
```
tamuratak, line 78: The prefix `cmd` seems not necessary.

src/providers/completer/environment.ts
```diff
@@ -3,9 +3,9 @@ import * as fs from 'fs'
 import {latexParser} from 'latex-utensils'
 
 import type {Extension} from '../../main'
-import type {IProvider, ILwCompletionItem} from './interface'
+import type {IProvider} from './interface'
 import {resolveCmdEnvFile} from './commandlib/commandfinder'
-import {cmdHasOptionalArgs, CmdSignature, getCmdSignature} from './command'
+import {Suggestion} from './command'
```
tamuratak, line 8: We had better write a comment on why we can use the `Suggention` class for `environment` while it is defined in `./command` . I am not sure where we should write it. It might not be appropriate at this line.

src/providers/completer/command.ts
```diff
@@ -21,38 +21,14 @@ export interface CmdItemEntry {
 }
 
 export interface CmdSignature {
-    name: string,
-    args: string
+    name: string, // name without leading `\`
+    args: string // {} for mandatory args and [] for optional args
```
tamuratak, line 25: You can add `readonly` to the properties.

You can also add `readonly` to the properties of `interface CmdItemEntry` except `snippet`.

src/providers/completer/command.ts
```diff
@@ -71,9 +47,37 @@ export function splitSignatureString(signature: string): CmdSignature {
...
 }
 
-export interface Suggestion extends ILwCompletionItem {
-    package: string,
+export class Suggestion extends vscode.CompletionItem implements ILwCompletionItem {
+    label: string
+    package: string
     signature: CmdSignature
+
+    constructor(label: string, pkg: string, signature: CmdSignature, kind?: vscode.CompletionItemKind) {
```
tamuratak, line 55: We can change `kind?` to `kind`.

tamuratak (changes requested):

src/providers/completer/environment.ts
jlelong: This is because every environment can also be inserted using `\envname`. All these commands are computed in the `Environment` class and then called from `Command`

https://github.com/James-Yu/LaTeX-Workshop/blob/db9044cc915d57f5e9013bf5f03cc910c88ae693/src/providers/completer/command.ts#L188

tamuratak: > every environments can also be inserted using \envname

I see. We had better write the explanation as a comment somewhere on `environment.ts`.

tamuratak (approved):

PR #3279, opened 2022-05-03
onEnterKey supports multi-cursors when all selected lines start with \item

### LINKED ISSUE #3278: PR #3255 breaks indentation for lines without "\item"
PR #3255 seems to change the indentation behavior for lines without "\item" in an unintended way:
``` latex
% with these changes:

    Some line. %<-press enter here
%<- cursor ends up here


% without these changes:

    Some line. %<-press enter here
    %<- cursor ends up here
```

_Originally posted by @ManuelHentschel in https://github.com/James-Yu/LaTeX-Workshop/issues/3255#issuecomment-1115119401_

jlelong:
Fix #3278. This PR is a followup on #3255, which turned to be buggy.

Multi-cursors is only partially supported because of a limitation of vscode's API to insert a new line and indent it. So `onEnterKey` only works when all the selected lines start with `\s*\\item`. If not, we simply return a normal `Enter` keystroke.

### DIFF
--- a/src/commander.ts
+++ b/src/commander.ts
@@ -438,33 +438,27 @@ export class Commander {
         }
 
         void editor.edit(editBuilder => {
+            // If we arrive here, all the cursors are at the end of a line starting with `\s*\\item`.
+            // Yet, we keep the conditions for the sake of maintenance.
             for (const selection of editor.selections) {
                 const cursorPos = selection.active
                 const line = editor.document.lineAt(cursorPos.line)
+                let matches: RegExpExecArray | null
 
-                // The line only consists of \item or \item[], delete its content
                 if (/^\s*\\item(\[\s*\])?\s*$/.exec(line.text)) {
+                    // The line is an empty \item or \item[]
                     const rangeToDelete = line.range.with(cursorPos.with(line.lineNumber, line.firstNonWhitespaceCharacterIndex), line.range.end)
                     editBuilder.delete(rangeToDelete)
-                    continue
-                }
-
-                // The line starts with \item and has more text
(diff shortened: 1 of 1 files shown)


src/commander.ts
```diff
@@ -421,50 +421,45 @@ export class Commander {
...
+        const allCursorsOnItem = editor.selections.every((selection: vscode.Selection) => {
+                const cursorPos = selection.active
+                const line = editor.document.lineAt(cursorPos.line)
+                return /^\s*\\item/.test(line.text) && (line.text.substring(cursorPos.character).trim().length === 0)
+        })
+        if (!allCursorsOnItem) {
+            return vscode.commands.executeCommand('type', { source: 'keyboard', text: '\n' })
+        }
 
         void editor.edit(editBuilder => {
```
tamuratak, line 440: Instead of `void`, it would be better to return the value that `editor.edit` returns.

tamuratak (changes requested):

tamuratak:
On the following example, typing enter does nothing.

```latex
\begin{itemize}
  \item \item (Enter here)
\end{itemize}
```

jlelong:
@tamuratak Thanks a lot for catching this one!

I have refactor the three possible cases
-  Empty `\item` or `\item[]`
- `\item[blabla]`
- `\item blabla` or `\item[.*] blabla`

Considering we only enter `editor.edit` if all the selected cursors are at the end of a line starting with  `\s*\item`, the above three cases should cover all the situations. Yet, as a fallback, a newline is added with the same indentation as the current one.

tamuratak (approved):

PR #3292, opened 2022-05-13
Remove setViewer 

tamuratak:
Remove setViewer to reduce the maintenance burden.

- The default had better be `tab` rather than `none`
- We should not modify users' `settings.json` on our side

@jlelong how do you think?


jlelong:
I agree that we can remove `setViewer`.

Issue #2393, opened 2020-11-25
Support lacheck


michaelcarno:
**Is your feature request related to a problem? Please describe.**
chktex is very old, and works bad with non english langs;

**Describe the solution you'd like**
lacheck works very well and fast, it suppot unicode;

jlelong:
This is worth having a look. I will try to make a PR

James-Yu:
I am looking into this feature. Is this still valid? I may work on it in the following days.

A trivial design things. Shall we unitfy `chktex` and the potential `lacheck` into a `linter` configuration class? Or treat these two as individual features? @jlelong

jlelong:
I have never used `lacheck`, so I do not have a specific opinion on how it outperforms `chktex`. Anyway, you are welcome to give it a try. Supporting a new linter also requires to implement a parser for its output.

I think it will be easier to maintain if one class is derived from the base `Linter` class for each linter . The same probably applies to the `LinterLogParser` class. As far as configuration variables are concerned, what about making the following changes

- Renaming `latex-workshop.chktex.run` into `latex-workshop.linting.run`, `latex-workshop.chktex.enabled` into `latex-workshop.linting.enabled`
 and `latex-workshop.chktex.delay` into `latex-workshop.linting.delay`.
- Adding `latex-workshop.linting.linter` with possible values `lacheck` or `chktex` (default)
- It would probably be more coherent to rename all the other configuration variables `latex-workshop.chktex` into `latex-workshop.linting.chktex`, but this would make a lot of changes.
- Add variables of the form `latex-workshop.linting.lacheck` 

Btw, there are many deprecated variables `latex-workshop.chktex.*`, which may now be removed?

PR #3355, opened 2022-06-29
Add deprecated config alert on activating extension

James-Yu:
This PR add a function to check if any of the deprecated configs are assigned with values different from the default one. If so, a warning is displayed to alert the user.

This PR prepares itself for #2393, which may involve many config changes. It's hard to request users to always read the changelog.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1542,7 +1542,7 @@
           "type": "string",
           "default": "#ffffff",
           "markdownDeprecationMessage": "**Deprecated**: Please use `#latex-workshop.view.pdf.color.light.backgroundColor#` and `#latex-workshop.view.pdf.color.dark.backgroundColor#` instead.",
-          "deprecationMessage": "Deprecated: Please use `#latex-workshop.view.pdf.color.light.backgroundColor#` and `#latex-workshop.view.pdf.color.dark.backgroundColor#` instead."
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1542,7 +1542,7 @@
           "type": "string",
           "default": "#ffffff",
           "markdownDeprecationMessage": "**Deprecated**: Please use `#latex-workshop.view.pdf.color.light.backgroundColor#` and `#latex-workshop.view.pdf.color.dark.backgroundColor#` instead.",
-          "deprecationMessage": "Deprecated: Please use `#latex-workshop.view.pdf.color.light.backgroundColor#` and `#latex-workshop.view.pdf.color.dark.backgroundColor#` instead."
```
jlelong, line 1545: The syntax `#config.variable#` is actually used be vs code to create a link to the new configuration variable. See https://code.visualstudio.com/api/references/contribution-points paragraph _Linking to settings_.

James-Yu: The latest doc changed the scope as follows:

> `#target.setting.id#`. This will work in markdownDescription, markdownEnumDescriptions, and markdownDeprecationMessage.

Therefore, only deprecateMessages are changed.

jlelong: My bad. I did not pay attention to this change.

PR #3356, opened 2022-06-30
Resolve #2607 Include <<child='...'>> in structure

### LINKED ISSUE #2607: Inclusion of Rnw files using <<child=...>> does not work

James-Yu:
In #3335, we change to use `latex-utensils` to generate project structure. However, the tool does not support Rnw sub-file syntax in chunk, i.e., `<<some-def, child='...', some-more-def>>`.

In this PR, we first use the existing `PathRegExp` class to generate a list of child files. When looping AST nodes, the line-of-appear of child files are also checked and inserted in function `buildLaTeXSectionFromFile`. Note that as this function only parses contents from one file, only line number is checked.

### DIFF
--- a/src/providers/structure.ts
+++ b/src/providers/structure.ts
@@ -4,6 +4,7 @@ import {latexParser,bibtexParser} from 'latex-utensils'
 
 import type { Extension } from '../main'
 import * as utils from '../utils/utils'
+import { PathRegExp, MatchPath, MatchType } from '../components/managerlib/pathutils'
 
 export class SectionNodeProvider implements vscode.TreeDataProvider<Section> {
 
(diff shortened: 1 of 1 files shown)


src/providers/structure.ts
```diff
@@ -4,6 +4,7 @@ import {latexParser,bibtexParser} from 'latex-utensils'
 
 import type { Extension } from '../main'
 import * as utils from '../utils/utils'
+import { PathRegExp, MatchPath, MatchType } from '../components/managerlib/pathutils'
```
tamuratak, line 7: Please stop importing `managerlib/pathutils` directly here, which violates the hierarchy of libraries. It would be better to consider moving `managerlib/pathutils` to `components/` and expose it as a component.

tamuratak (changes requested):

src/providers/structure.ts
James-Yu: That's exactly what I thought when I import that lib. Quite strange why it appears there.

James-Yu: After looking into the `pathutils` code, I think it would be more appropriate to refactor that class in another dedicated PR. I will work on it immediately after this PR merged.

PR #3357, opened 2022-06-30
Refactor cleaner glob methods

XuehaiPan:
Implementation cleanup for `Cleaner.cleanGlob` method for better understanding.

cc @jlelong

### DIFF
--- a/src/components/cleaner.ts
+++ b/src/components/cleaner.ts
@@ -17,17 +17,19 @@ function unique(sequence: string[]): string[] {
 /**
  * Globs all given patterns into absolute paths. The result will be sorted in
  * reverse order and all tailing slashes will be stripped.
+ *
+ * The result is sort in descending dictionary order, make sure the children are sorted before the parents.
+ * For example: [..., 'out/folder1/folder2/', 'out/folder1/', ...] ('out/folder1/folder2/' > 'out/folder1/' in directory order)
  */
 function globAll(globs: string[], cwd: string): string[] {
     return unique(
         globs.map(g => glob.sync(g, { cwd }))
-             // Reduce the array of arrays to a single array containing all the elements
-             .reduce((all, curr) => all.concat(curr), [])
-             // Resolve absolute path (tailing slashes will be stripped)
+             // Flatten the array of arrays to a single array containing all the elements
+             .flat()
+             // Resolve absolute paths (tailing slashes will be stripped)
              .map((globedPath: string): string => path.resolve(cwd, globedPath))
     )
-    // Sort in descending dictionary order, make sure the children are sorted before the parents
-    // For example: [..., 'out/folder1/folder2/', 'out/folder1/', ...] ('out/folder1/folder2/' > 'out/folder1/' in directory order)
+    // Sort in descending dictionary order
     .sort((a, b) => b.localeCompare(a))
 }
 
@@ -64,7 +66,7 @@ export class Cleaner {
 
     /**
      * Splits the given glob patterns into three distinct groups (duplicates will be ignored)
-     *   1. file globs (not end with tailing slashes)
+     *   1. file or folder globs (not end with tailing slashes)
      *   2. globs explicitly for folders
      *   3. folder globs with globstar (`**`)
      *
@@ -74,8 +76,8 @@ export class Cleaner {
      * @param globs a list of glob patterns
      * @returns three distinct groups of glob patterns
      */
-    private static splitGlobs(globs: string[]): { fileGlobs: string[], folderGlobsExplicit: string[], folderGlobsWithGlobstar: string[] } {
-        const fileGlobs: string[] = []
+    private static splitGlobs(globs: string[]): { fileOrFolderGlobs: string[], folderGlobsExplicit: string[], folderGlobsWithGlobstar: string[] } {
+        const fileOrFolderGlobs: string[] = []
         const folderGlobsWithGlobstar: string[] = []
         const folderGlobsExplicit: string[] = []
 
@@ -96,12 +98,11 @@ export class Cleaner {
                     folderGlobsExplicit.push(pattern)
                 }
             } else {
-                // The pattern can match both files and folders
-                fileGlobs.push(pattern)
+                fileOrFolderGlobs.push(pattern)
             }
         }
 
-        return { fileGlobs, folderGlobsExplicit, folderGlobsWithGlobstar }
+        return { fileOrFolderGlobs, folderGlobsExplicit, folderGlobsWithGlobstar }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


James-Yu:
Please use latest commit as base.

jlelong:
I think a rebase on `master` plus a force-push should make it. No need to close the PR.

XuehaiPan:
> Please use latest commit as base.

Already using the latest commit. The old logic was implemented to glob all patterns at once.

Since we do cleaning in a two-stage manner, we can split the glob patterns into distinct groups and gob separately. Because the file globs can match folders (e.g. `abc/**` can match folder` abc`). We should get all glob results (files and folders) in a single function and remove the duplicates.

src/components/cleaner.ts
```diff
@@ -38,6 +62,49 @@ export class Cleaner {
...
+     *   2. globs explicitly for folders
+     *   3. folder globs with globstar (`**`)
+     *
+     * We will remove the <1.> type paths if they are files, remove the <2.> type
+     * paths if they are empty folders, and ignore the <3.> type paths.
+     *
+     * @param globs a list of glob patterns
+     * @returns three groups of glob patterns
+     */
+    private static splitGlobs(globs: string[]): [string[], string[], string[]] {
```
James-Yu, line 77: Better use object with labels as return instead of array of arrays.

James-Yu:
LGTM @jlelong

src/components/cleaner.ts
```diff
@@ -38,6 +62,48 @@ export class Cleaner {
...
+                    // Contains globstar `**` in the last component
+                    folderGlobsWithGlobstar.push(pattern)
+                } else {
+                    // All glob patterns EXPLICITLY given for folders, the glob should be end with a slash (`path.sep`)
+                    // and the last component should not contain globstar `**`
+                    //
+                    // Positive examples: ['abc/', 'abc*/', '**/abc*/', 'abc/**/def/', 'abc/**/def*/']
+                    // Negative examples:
+                    //   - not end with a slash: ['abc', 'abc*', 'abc/**/def*']
+                    //   - contain globstar `**` in the last component: ['**', '**/', 'abc/**', 'abc/**/', 'abc/def**/', 'abc/d**ef/']
```
jlelong, line 95: Move these explanations in the documentation of the function. Long comments tend make the code hard to read and become counterproductive.

src/components/cleaner.ts
```diff
@@ -2,11 +2,35 @@ import * as vscode from 'vscode'
...
+function globAll(globs: string[], cwd: string): string[] {
+    return unique(
+        globs.map(g => glob.sync(g, { cwd }))
+             // Reduce the array of arrays to a single array containing all the elements
+             .reduce((all, curr) => all.concat(curr), [])
+             // Resolve absolute path (tailing slashes will be stripped)
+             .map((globedPath: string): string => path.resolve(cwd, globedPath))
+    )
+    // Sort in descending dictionary order, make sure the children are sorted before the parents
+    // For example: [..., 'out/folder1/folder2/', 'out/folder1/', ...] ('out/folder1/folder2/' > 'out/folder1/' in directory order)
```
jlelong, line 30: Move this comment to the function's documentation

src/components/cleaner.ts
```diff
@@ -2,11 +2,35 @@ import * as vscode from 'vscode'
...
+/**
+ * Globs all given patterns into absolute paths. The result will be sorted in
+ * reverse order and all tailing slashes will be stripped.
+ */
+function globAll(globs: string[], cwd: string): string[] {
+    return unique(
+        globs.map(g => glob.sync(g, { cwd }))
+             // Reduce the array of arrays to a single array containing all the elements
+             .reduce((all, curr) => all.concat(curr), [])
+             // Resolve absolute path (tailing slashes will be stripped)
```
jlelong, line 26: It seems to me that we do not care about trailing slashes being stripped anymore. That was important prior to this refactoring.

src/components/cleaner.ts
```diff
@@ -52,49 +118,22 @@ export class Cleaner {
...
-            ([_, realPath]) => !uniqueRealPaths.has(realPath + path.sep)
-        )
+        // All folders explicitly specified by the user, remove them if empty after cleaning files
+        const explicitFolders: string[] = globAll(folderGlobsExplicit, outdir)
 
-        await this.cleanGlobFiles(pathPairsToRemove)
-        await this.cleanGlobFolders(globs, pathPairsToRemove)
-    }
+        // All files to remove. Note that the file globs can match directories, we will check it with `fs.stat`
+        // Pattern `abc/**` can match folder as "abc", remove duplicates
```
jlelong, line 131: Remove the comment.

src/components/cleaner.ts
```diff
@@ -52,49 +118,22 @@ export class Cleaner {
...
+        const explicitFoldersSet: Set<string> = new Set(explicitFolders)
+        const files: string[] = globAll(fileGlobs, outdir).filter(file => !explicitFoldersSet.has(file))
```
jlelong, line 133: We had better rename `files` into `filesOrDirs`.

src/components/cleaner.ts
```diff
@@ -112,27 +151,9 @@ export class Cleaner {
...
-        // Sort by descending dictionary order, make sure the children folders are sorted before the parents
-        // For example: [..., 'out/folder1/folder2/', 'out/folder1/', ...] ('out/folder1/folder2/' > 'out/folder1/' in directory order)
-        .sort((a, b) => b.localeCompare(a))
-
-        // Remove empty folders EXPLICITLY specified
-        for (const folderRealPath of explicitFolderPathsToRemove) {
+        // Then remove empty folders EXPLICITLY specified
+        for (const folderRealPath of explicitFolders) {
             try {
                 // We are sure that this is a folder because it's globed by pattern with a trailing slash
```
jlelong, line 158: I guess the name `explicitFolders` is sufficiently meaningful to know that it only contains folders.

jlelong:
With this PR (the previous one actually), subfolders can be cleaned even if the variable `latex-workshop.latex.clean.subfolder.enabled` is set to `false` by simply passing `**` to `latex-workshop.latex.clean.fileTypes`.

src/components/cleaner.ts
XuehaiPan: > That was important prior to this refactoring.

The old implementation globs all patterns at once. Here we split the globs into two groups and get the results separately. We don't need the tailing slashes anymore.

src/components/cleaner.ts
XuehaiPan: Glob pattern `*/` cannot feed into multi-line comment `/** ... */`.

XuehaiPan:
> With this PR (the previous one actually), subfolders can be cleaned even if the variable `latex-workshop.latex.clean.subfolder.enabled` is set to `false` by simply passing `**` to `latex-workshop.latex.clean.fileTypes`.

The cleaner with glob `"**"` only removes files but keeps the empty folders.

```console
$ tree out 
out
├── a.aux
└── temp
    └── b.aux

1 directory, 2 files
```

- After cleaning with `"latex-workshop.latex.clean.fileTypes": ["**"]`:

    ```console
    $ tree out
    out
    └── temp
    
    1 directory, 0 files
    ```

- After cleaning with `"latex-workshop.latex.clean.fileTypes": ["*.aux"]`:

    ```console
    $ tree out
    out
    └── temp
    
    1 directory, 0 files
    ```

- After cleaning with `"latex-workshop.latex.clean.fileTypes": ["*.aux", "temp"]`:

    ```console
    $ tree out
    out
    └── temp
    
    1 directory, 0 files
    ```

- After cleaning with `"latex-workshop.latex.clean.fileTypes": ["*.aux", "temp/"]`:

    ```console
    $ tree out
    out
    
    0 directory, 0 files
    ```

jlelong:
Let me rephrase. With this PR and the setting `latex-workshop.latex.clean.fileTypes: ["**/*.aux"]`, `.aux` are removed in all subfolders even if `latex-workshop.latex.clean.subfolder.enabled: false` whereas https://github.com/James-Yu/LaTeX-Workshop/blob/92d336e63c2a84c38039449d008dcd17e40ef075/package.json#L1214-L1219

XuehaiPan:
> With this PR and the setting `latex-workshop.latex.clean.fileTypes: ["**/*.aux"]`, `.aux` are removed in all subfolders even if `latex-workshop.latex.clean.subfolder.enabled: false`

Without this PR and the previous one, we also remove files in sub-folders with glob `"**/*.aux"` anyway. Globstar `**` will search in sub-folders even if `latex-workshop.latex.clean.subfolder.enabled: false`.

jlelong:
> Without this PR and the previous one, we also remove files in sub-folders with glob `"**/*.aux"` anyway. Globstar `**` will search in sub-folders even if `latex-workshop.latex.clean.subfolder.enabled: false`.

Indeed, you are right. Then, I am wondering if this setting is really needed. @James-Yu

jlelong (approved):

XuehaiPan:
> Then, I am wondering if this setting is really needed.

I think `latex-workshop.latex.clean.subfolder.enabled` is a convenient way to touch every pattern in the array. Otherwise, the user will need to do it manually.

James-Yu:
> Indeed, you are right. Then, I am wondering if this setting is really needed. @James-Yu

I will think of it soon (TM).

PR #3358, opened 2022-07-01
Refactor `pathutils`

James-Yu:
This PR handles https://github.com/James-Yu/LaTeX-Workshop/pull/3356#discussion_r911516376

The PR does two things:

1. The original `pathutils` are split into two. This is because that the classes `PathRegExp` and `PathUtils` indeed do two different things and are not coupled at all. The former concerns both `manager` and `structure` (and maybe more to come) while the latter only relates with `manager`. Currently, the former is moved to a more general place `utils` and the latter remains in `managerlib`.
2. As `PathRegExp` is moved to `utils`, it shall adhere to the practice in the folder. I checked the other files in the same folder and found that all files export only functions. Therefore `PathRegExp` is refactored into dedicated functions. This is also natural as the purpose of the original class is to hold the two `RegExp` objects, which are reused on the same piece of text.

I have also thought of other solutions, e.g., making `PathRegExp` a component like builder. However, this may involve explicitly resetting the index of regexps, which can be a pain. The current implementation is analogous to the original one:
- Previous: Create an object of `PathRegExp` class, use the object to exec and parse path.
- Now: Create an object of `RegExp` using either `createInputRegExp()` or `createChildRegExp()`, then use the object to exec (and implicitly parse path, which was strongly coupled with `exec` in previous code. Also, as such, type `MatchPath` needs not to be exposed anymore (and enum `MatchType`).


jlelong:
I think splitting `PathRegExp` into dedicated functions is not a good idea. All the functions are linked together and the typing system will not be able to check the correct arguments are passed. It will make maintenance much more difficult and error prone. I am making my point precise by commenting the code directly.

James-Yu:
I am quite neutral to either function-based or a class-based implementation. The only concern is to keep consistent with other files in `utils` folder. If that is not a thing (a confirmation is greatly appreciated!), I am more than happy to switch back to the original class-based implementation.

src/utils/inputfilepath.ts
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Return the matched input path. If there is no match, return undefined
+ *
+ * @param content the string to match the regex on
+ * @param regexp the input regex created by {@link createInputRegExp()}
+ * @param currentFile is the name of file in which the regex is executed
+ * @param rootFile
+ */
+export function execInputRegExp(content: string, regexp: RegExp, currentFile: string, rootFile: string): {path: string, match: MatchPath} | undefined {
+    const result = regexp.exec(content)
```
jlelong, line 52: The `regexp` argument must be the one returned by `createInputRegExp`. The content of `result` is tight to this particular regexp. In particular, if the regexp in `createInputRegExp` is modified, then the parsing should also modified in this function. I think ti would better that the architecture links the two functions together into a class.

src/utils/inputfilepath.ts
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Return the matched child path. If there is no match, return undefined
+ *
+ * @param content the string to match the regex on
+ * @param regexp the child regex created by {@link createChildRegExp()}
+ * @param currentFile is the name of file in which the regex is executed
+ * @param rootFile
+ */
+export function execChildRegExp(content: string, regexp: RegExp, currentFile: string, rootFile: string): {path: string, match: MatchPath} | undefined {
+    const result = regexp.exec(content)
```
jlelong, line 76: Same comment as above

src/utils/inputfilepath.ts
```diff
@@ -0,0 +1,131 @@
...
+ * undefined
+ *
+ * @param content the string to match the regex on
+ * @param regexps the input and child regexes created by
+ *        {@link createInputRegExp()} and {@link createChildRegExp()}
+ * @param currentFile is the name of file in which the regex is executed
+ * @param rootFile
+ */
+export function execInputChildRegExps(content: string, regexps: {input: RegExp, child: RegExp}, currentFile: string, rootFile: string): {path: string, match: MatchPath} | undefined {
+    return execInputRegExp(content, regexps.input, currentFile, rootFile)
```
jlelong, line 102: Same comment as above concerning `regexps: {input: RegExp, child: RegExp}`.

jlelong:
> I am quite neutral to either function-based or a class-based implementation. The only concern is to keep consistent with other files in `utils` folder. If that is not a thing (a confirmation is greatly appreciated!), I am more than happy to switch back to the original class-based implementation.

I understand your concern about keeping the design of all elements consistent across `utils:`, but I tend to think the cons outweigh the pros. The conclusion might also be that `utils` is actually not the right place. I can think of two solutions 

1. Extract `PathRegExp` to `components/helperlib/inputfile.ts`.
2. Extract `PathRegExp` to `components/` and call the existing `.resetlastIndex` method before using the component.

Approach 2 keeps the design of components more consistent but approach 1 is more flexible and probably easier to maintain.

Any other suggestions?

jlelong:
I have a slight preference for the class based approach is it keeps the definition of the regex's and the analysis of the results together. I think it is easier to maintain.

James-Yu:
> I have a slight preference for the class based approach is it keeps the definition of the regex's and the analysis of the results together. I think it is easier to maintain.

Yes. Later when we forgot about these functions, class makes it easier to know that these "functions" are coupled. Will do the change shortly.

PR #3360, opened 2022-07-01
Add support to LaCheck

### LINKED ISSUE #2393: Support lacheck

James-Yu:
This PR resolves #2393 

Both `Linter` and `LinterLogParser` are abstracted, which are extended by `chktex` and `lacheck`. Further extension to other linters are possible.

Hope the file changes are straightforward.

edit: Now the linter log parsers are located in the respective linter ts file, instead of `components/parser`. I'm not sure if this is a good practice as we now have three classes (one abstract, two for linter log parser implementations). Putting them all in `parser` folder may make things confusing. I'm happy to move them back if it is considered more appropriate.


jlelong:
In #2393, I wrote

> I think it will be easier to maintain if one class is derived from the base `Linter` class for each linter . The same probably applies to the `LinterLogParser` class.

I am giving a second thought to this. I spoke about abstract classes because I am more involved in C++ programming than typescript actually but I think interfaces may be more appropriate here. Considering a parser is actually tight to a specific linter, I would rather keep the `parseLog` method inside each linter definition.

tamuratak:
@jlelong  I hope you request @James-Yu write tests. It is one of the reviewers' jobs. You never do that. Even if you have no time to write tests, he seems to have the time. You would have recognized that if tests had been available, rewriting such as #3335 would have been much more smooth.

We should share the thought that writing tests is the essential part of maintainance and is also a real and respectful contribution. We cannot call rewriting without tests "refactoring".

James-Yu:
> We should share the thought that writing tests is the essential part of maintainance and is also a real and respectful contribution. We cannot call rewriting without tests "refactoring".
> 
> Not writing tests is a bad habit of software in academics. We don't have to stick to it.

Though harsh, I agree with you. I just need much time to learn how the current test framwork functions. Also in my local dev environment, running `npm run test` is causing heavy file pollution in the fixture folders.

@tamuratak do you have a weblink to some online tutorials introducing the system and how to use it? I am perfectly fine to write tests (and make up the structure ones and others.)

tamuratak:
The shortest path is to imitate the following:

Test for a function:

https://github.com/James-Yu/LaTeX-Workshop/blob/170dbc1dc3856e0794880c5b55be622cb0c02fd5/test/unittests/providers/completer/commandlib/commandfinder.test.ts#L8-L10

Test for a component:

https://github.com/James-Yu/LaTeX-Workshop/blob/170dbc1dc3856e0794880c5b55be622cb0c02fd5/test/unittests/components/locatorlib/synctex.test.ts#L15-L25

Run `npm run test unittest`. You can refer to the reference https://nodejs.org/api/assert.html

As you can see, it would be convenient to put test files in the same place as the file we want to test. You can also test `exports.realExtension.someComponent` directly instead of constructing the component.

A much more difficult part is making a component "testable" when it is difficult to write tests for it straightly. 

It would be good to read the followings although not free:

- Martin Fowler "Refactoring" 2nd edition (Sample Code written in JavaScript)
- Michael Feathers "Working Effectively with Legacy Code" (Sample Code written in C++, still good, but old)

src/components/linter.ts
```diff
@@ -1,248 +1,50 @@
...
-import * as path from 'path'
-import * as fs from 'fs'
-import * as os from 'os'
-import {ChildProcessWithoutNullStreams, spawn} from 'child_process'
-import {EOL} from 'os'
 
 import type {Extension} from '../main'
+import { Linter as AbstractLinter } from './linters/linter'
+import { ChkTeX } from './linters/chktex'
+import { LaCheck } from './linters/lacheck'
```
tamuratak, line 6: Please rename the directory `linters` to `linterlib` by convention.

tamuratak (changes requested):

tamuratak:
We had better stop using inheritance as @jlelong suggested. One of the implementations without inheritance I can think of is the following:

https://github.com/tamuratak/LaTeX-Workshop/commits/linter-refactor_composition

tamuratak:
I have a question. Why isn't it possible to use `chktex` and `lacheck` at the same time? Is there any reason?

jlelong:
> I have a question. Why isn't it possible to use `chktex` and `lacheck` at the same time? Is there any reason?

Both could used together. I do not think of any reason to make their use exclusive even though it would generate many double errors/warnings. This means that we have to replace `latex-workshop.linting.linter` by `latex-workshop.linting.chktex.enabled` and `latex-workshop.linting.lacheck.enabled`.

src/components/linter.ts
```diff
@@ -1,248 +1,52 @@
...
-import {ChildProcessWithoutNullStreams, spawn} from 'child_process'
-import {EOL} from 'os'
 
 import type {Extension} from '../main'
+import { ChkTeX } from './linterlib/chktex'
+import { LaCheck } from './linterlib/lacheck'
 
 export class Linter {
-    private readonly extension: Extension
+    protected readonly extension: Extension
```
tamuratak, line 8: Please change `protected` to `private`.

tamuratak (changes requested):

src/components/linter.ts
```diff
@@ -1,248 +1,52 @@
...
+        this.lacheck = new LaCheck(extension)
     }
 
-    private get rcPath() {
-        let rcPath: string
-        // 0. root file folder
-        const root = this.extension.manager.rootFile
-        if (root) {
-            rcPath = path.resolve(path.dirname(root), './.chktexrc')
+    private getLinters(scope?: vscode.ConfigurationScope | null): (ChkTeX | LaCheck)[] {
```
tamuratak, line 19: We can change the type `vscode.ConfigurationScope | null` to `vscode.ConfigurationScope`.

tamuratak (changes requested):

src/components/linterlib/chktex.ts
```diff
@@ -0,0 +1,307 @@
...
+export class ChkTeX {
+    readonly #linterName = 'ChkTeX'
+    readonly linterDiagnostics: vscode.DiagnosticCollection = vscode.languages.createDiagnosticCollection(this.#linterName)
+    readonly #linterUtil: LinterUtil
+
+    constructor(private readonly extension: Extension) {
+        this.#linterUtil = new LinterUtil(extension)
+    }
+
+    async lintRootFile() {
```
jlelong, line 19: The methods `lintRootFile` and `lintFile` share a lot of common code. Actually, they only differ by a couple of lines. I suggest to refactor these two methods to extract the command part to ease maintainance.

src/components/linterlib/lacheck.ts
```diff
@@ -0,0 +1,141 @@
...
+            if ('stdout' in err) {
+                stdout = err.stdout as string
+            } else {
+                return
+            }
+        }
+        this.parseLog(stdout)
+    }
+
+    async lintFile(document: vscode.TextDocument) {
```
jlelong, line 42: The methods `lintRootFile` and `lintFile` share a lot of common code. Actually, they only differ by a couple of lines. I suggest to refactor these two methods to extract the command part to ease maintainance.

src/components/linter.ts
```diff
@@ -1,248 +1,50 @@
...
+        this.lacheck = new LaCheck(extension)
     }
 
-    private get rcPath() {
-        let rcPath: string
-        // 0. root file folder
-        const root = this.extension.manager.rootFile
-        if (root) {
-            rcPath = path.resolve(path.dirname(root), './.chktexrc')
+    private getLinters(scope?: vscode.ConfigurationScope): (ChkTeX | LaCheck)[] {
```
jlelong, line 17: Why not defining a linter interface that `chkTeX` and `LaCheck` would implement? Something like
```ts
interface ILinter {
  readonly linterDiagnostics: vscode.DiagnosticCollection,
  readonly #linterName: string,
  async lintRootFile(),
  async lintFile(document: vscode.TextDocument),
  parseLog(log: string, filePath?: string)
}
```

This would allow `getLinters` to return a `ILinter[]` and more importantly it would clarify what has to be implemented to add a new linter.

src/components/linter.ts
```diff
@@ -1,248 +1,50 @@
...
+    private readonly chktex: ChkTeX
+    private readonly lacheck: LaCheck
     private linterTimeout?: NodeJS.Timer
-    private readonly currentProcesses = Object.create(null) as { [linterId: string]: ChildProcessWithoutNullStreams }
 
-    constructor(extension: Extension) {
-        this.extension = extension
+    constructor(private readonly extension: Extension) {
+        this.chktex = new ChkTeX(extension)
+        this.lacheck = new LaCheck(extension)
```
jlelong, line 14: I find it weird to explicitly instantiate these two classes before knowing if a linter will actually be used.

jlelong (commented):
Sorry for not responding faster but I was busy refactoring the LaTeX grammar, which had become very hard to maintain.

src/components/linterlib/chktex.ts
```diff
@@ -0,0 +1,300 @@
+import * as vscode from 'vscode'
+import * as path from 'path'
+import * as fs from 'fs'
+import * as os from 'os'
+
+import type {Extension} from '../../main'
+import { LinterUtil } from './linterutil'
+import { convertFilenameEncoding } from '../../utils/convertfilename'
+
+export class ChkTeX {
```
jlelong, line 10: This class should implement `ILinter`
```ts
export class ChkTeX implements ILinter
```

src/components/linterlib/lacheck.ts
```diff
@@ -0,0 +1,143 @@
+import * as vscode from 'vscode'
+import * as path from 'path'
+import * as fs from 'fs'
+
+import type { Extension } from '../../main'
+import { LinterUtil } from './linterutil'
+import { convertFilenameEncoding } from '../../utils/convertfilename'
+
+export class LaCheck {
```
jlelong, line 9: This class should implement `ILinter`
```ts
export class LaCheck implements ILinter
```

src/components/linter.ts
```diff
@@ -1,248 +1,66 @@
...
+export interface ILinter {
+    readonly linterDiagnostics: vscode.DiagnosticCollection,
+    lintRootFile(): void,
+    lintFile(document: vscode.TextDocument): void,
+    parseLog(log: string, filePath?: string): void
+}
 
 export class Linter {
-    private readonly extension: Extension
+    readonly #linters: {[key: string]: ILinter} = {}
```
tamuratak, line 14: We had better add `#lacheck: ILinter | undefined` and `#chktex: ILinter | undefined` instead of `#linters`.

We should avoid using `object` to prevent prototype pollution. See jlelong/LaTeX-Workshop-security/issues/3.

tamuratak (changes requested):

src/components/linter.ts
James-Yu: I agree. Just wondering which one is better, explicit variables on each linter or a `Map<string, ILinter>`. I am a little bit inclined on the latter, which can be (very trivially) more friendly to possible future linters? Any idea?

jlelong: Using a `Map` looks good to me. @tamuratak ?

tamuratak: It is up to you.

src/components/linter.ts
```diff
@@ -1,248 +1,68 @@
...
-            return
-        }
-        if (fs.existsSync(rcPath)) {
-            return rcPath
+    private get chktex(): ChkTeX {
+        const linterId = 'chktex'
+        if (!this.#linters.has(linterId)) {
+            this.#linters.set(linterId, new ChkTeX(this.extension))
         }
+        return this.#linters.get(linterId) as ChkTeX
```
tamuratak, line 24: Please remove `as ChkTeX`. Please stop using type assertions casually. When you use type assertions, it is just wrong generally.

If you don't know how to fix it, please use `#chktex` as I suggested.

src/components/linter.ts
```diff
@@ -1,248 +1,68 @@
...
-        }
-        if (fs.existsSync(rcPath)) {
-            return rcPath
+    private get lacheck(): LaCheck {
+        const linterId = 'lacheck'
+        if (!this.#linters.has(linterId)) {
+            this.#linters.set(linterId, new LaCheck(this.extension))
         }
-        return undefined
+        return this.#linters.get(linterId) as LaCheck
```
tamuratak, line 32: Please stop using type assertions.

tamuratak (changes requested):

PR #3364, opened 2022-07-03
Remove console.log in deprecate config check

James-Yu:


src/main.ts
```diff
@@ -62,7 +62,6 @@ function deprecateConfigCheck() {
     deprecatedConfigs.forEach(config => {
         const defaultValue = configuration.inspect(config)?.defaultValue
         const configValue = configuration.get(config)
-        console.log(defaultValue, configValue)
         if (JSON.stringify(defaultValue) !== JSON.stringify(configValue)) {
             const fullConfig = `latex-workshop.${config}`
             void vscode.window.showWarningMessage(`Config "${fullConfig}" is deprecated. ${packageDef.contributes.configuration.properties[fullConfig].deprecationMessage}`)
```
jlelong, line 67: It may be worth logging this also using `logger.addLogMessage` to help identify false issues if any.

James-Yu: Exactly. I removed this when creating #3365 as it produces random logs in ci. Adding log to trace the deprecacy is great, will add shortly.

PR #3365, opened 2022-07-03
Add tests for structuring

James-Yu:
This PR adds tests on project structure. The following cases are tested

- Section hierarchy
- Configs `view.outline.sections` and `view.outline.numbers.enabled`
- Parsing labels properly with their correct line number
- Floats, frames and their captions
- Config `view.outline.floats.enabled`
- Subfiles parsing and testable edge use cases in https://github.com/James-Yu/LaTeX-Workshop/pull/3335#issuecomment-1158632737

This is my first attempt on writing tests for LW. Please feel free to comment.


jlelong:
Something seems to have gone wrong with the indentation of `package.json`.

tamuratak:
Use [assert.strictEqual](https://nodejs.org/docs/latest-v16.x/api/assert.html#assertstrictequalactual-expected-message) instead of `assert.equal`.

- https://nodejs.org/docs/latest-v16.x/api/assert.html#assertequalactual-expected-message

tamuratak:
Don't change things not related to this PR.  Do it in a separated PR.

> Tabsize is 2

tamuratak:
As I said before, please consider writing tests as unit tests, no as CI tests. CI tests tend to be slow and unreliable. If you can not know how to write, please ask me.

James-Yu:
> @James-Yu, As I said before, please consider writing tests as unit tests, no as CI tests. CI tests tend to be slow and unreliable. If you don't know how to write, please ask me.

TBH I don't know the difference of the current structure tests and the ones under `unittest` folder. Structure is coupled with manager, or at least the `rootFile`. It is unclear to me how to transform the existing tests into unit tests besides moving the tests under `unittest`.

tamuratak:
>  I don't know the difference of the current structure tests and the ones under unittest folder.

I wrote the difference.

> CI tests tend to be slow and unreliable.

CI tests with `runTestWithFixture` tend to be slow and unreliable since we launch a new process for each call. On some platforms, we have to skip some tests because of unknown reasons. On unit tests with `runUnitTestWithFixture`, we don't have to launch it, which makes the tests fast and reliable.

>  Structure is coupled with manager, or at least the rootFile. It is unclear to me how to transform the existing tests into unit tests

There are some ways. If it is difficult, you don't have to do it.

Fake method:

1. Make a fake `extension` object which returns only `extension.manager.rootFile` and other necessary properties. 
2. Construct a `SectionNodeProvider` with the fake `extension`. 
3. You can test the component with any `rootFile` you want.

Decoupling:

1. Decouple `Structure` with `manager`. For example, move `this.extension.manager.rootFile` to the argument of the method, and call each method with `rootFile`.
2. Then, you can test the method with any `rootFile` you want.

tamuratak:
Before you implement something in a different way from someone suggest, you had better ask them the new approach would be appropriate. There should be a reason they suggest. Implementing in a wrong way wastes time.

James-Yu:
I'm closing this PR in favor of https://github.com/James-Yu/LaTeX-Workshop/pull/3365#issuecomment-1173062768. Will come back in few days.

tamuratak:
Notice that there are disadvantages for above methods. For example, it would be complicated to make the fake object.

There are similar words, mock, stub, and, fake. Please refer to https://stackoverflow.com/a/346440

It would be better to write tests for `src/components/cleaner.ts` too although it seems not testable.

PR #3370, opened 2022-07-05
Add a log message on deprecated config

James-Yu:


tamuratak:
You had better move `deprecateConfigCheck` to `src/components/configuration.ts`. If you need the logger in the function, it indicates that the function should be the method of a certain component.

You should not add functions to `src/main.ts`, which makes maintenance harder.

PR #3377, opened 2022-07-11
Refactor commands duplication detection for intellisense

jlelong:
This PR refactors the duplication detection for command intellisense. It introduces two new classes
- `CommandSignatureDuplicationDetector`: the complete signature is used.
- `CommandNameDuplicationDetector`: only the command name is used.

### DIFF
--- a/src/providers/completer/command.ts
+++ b/src/providers/completer/command.ts
@@ -95,6 +95,10 @@ export class CommandSignatureDuplicationDetector {
 export class CommandNameDuplicationDetector {
     private readonly cmdSignatureList: Set<string> = new Set<string>()
 
+    constructor(suggestions: Suggestion[] = []) {
+        this.cmdSignatureList = new Set<string>(suggestions.map(s => s.name()))
+    }
+
     add(cmd: Suggestion): void
     add(cmdName: string): void
     add(cmd: any): void {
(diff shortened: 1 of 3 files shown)


tamuratak:
You had better move export `class CommandSignatureDuplicationDetector` and `class CommandNameDuplicationDetector` under  `completer/commandlib`. You can re-export them in `completer/command.ts` if you need.

It seems weird that `commandfinder.ts ` imports components from `command.ts`. `command.ts` should import from `commandlib`.

jlelong:
> It seems weird that `commandfinder.ts ` imports components from `command.ts`. `command.ts` should import from `commandlib`.

For an other PR, should we be moving the declaration of `Suggestion` to `commandfinder`?

https://github.com/James-Yu/LaTeX-Workshop/blob/5811c2430ed50aebef5210e3aac924bdab603f48/src/providers/completer/commandlib/commandfinder.ts#L5

I guess it would also make more sense to move `CmdSignature` and `splitSignatureString` to `commandfinder`.

https://github.com/James-Yu/LaTeX-Workshop/blob/5811c2430ed50aebef5210e3aac924bdab603f48/src/providers/completer/command.ts#L23-L48

tamuratak:
> should we be moving the declaration of Suggestion to commandfinder?

> I guess it would also make more sense to move CmdSignature and splitSignatureString to commandfinder.

It is up to you. I have no specific opinion.

We had better rename `Suggestion` to `CmdEnvEntry`. `Suggestion` is too general.

PR #3378, opened 2022-07-11
Move obsolete configuration detection to Configuration

jlelong:
This PR is a follow-up on #3370 and moves the check of modified deprecated setting to the `Configuration` class. 

@James-Yu The current implementation may fail to detect modified deprecated settings defined in a specific workspace inside a [multi-root workspace](https://code.visualstudio.com/docs/editor/multi-root-workspaces). 

Instead of https://github.com/James-Yu/LaTeX-Workshop/blob/070599da15b01588e163705025f1e4c5f8419a55/src/components/configuration.ts#L65
we should be checking the configuration value for every workspace in `vscode.workspace.workspaceFolders`

https://github.com/James-Yu/LaTeX-Workshop/blob/070599da15b01588e163705025f1e4c5f8419a55/src/components/configuration.ts#L37-L39


James-Yu (approved):

jlelong:
with [242382c](https://github.com/James-Yu/LaTeX-Workshop/pull/3378/commits/242382c416046b96f748e22777b921b10e8920dc), we check for deprecated settings in all workspaces.

PR #3379, opened 2022-07-14
Make Suggestion class names more explicit

jlelong:
This PR renames the different `Suggestion` classes with a more explicit name

- `CiteSuggestion`
- `CmdEnvSuggestion`
- `GlossarySuggestion`

### DIFF
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -12,7 +12,7 @@ import type {Extension} from '../main'
 import * as eventbus from './eventbus'
 import type {CmdEnvSuggestion as CmdEnvEntry} from '../providers/completer/command'
 import type {CiteSuggestion as CiteEntry} from '../providers/completer/citation'
-import type {Suggestion as GlossEntry} from '../providers/completer/glossary'
+import type {GlossarySuggestion as GlossaryEntry} from '../providers/completer/glossary'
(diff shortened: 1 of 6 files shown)


src/components/manager.ts
```diff
@@ -10,9 +10,9 @@ import {InputFileRegExp} from '../utils/inputfilepath'
 
 import type {Extension} from '../main'
 import * as eventbus from './eventbus'
-import type {Suggestion as CiteEntry} from '../providers/completer/citation'
-import type {Suggestion as CmdEnvEntry} from '../providers/completer/command'
-import type {Suggestion as GlossEntry} from '../providers/completer/glossary'
+import type {CmdEnvSuggestion as CmdEnvEntry} from '../providers/completer/command'
+import type {CiteSuggestion as CiteEntry} from '../providers/completer/citation'
+import type {GlossarySuggestion as GlossaryEntry} from '../providers/completer/glossary'
```
tamuratak, line 15: What is the reason for using aliases here? We would have no reason to use the aliases.

James-Yu:
@jlelong Really sorry that I accidentially overwrote this PR... May you please re-commit the changes? I promise that mistake won't happen again...

jlelong:
@James-Yu No worries. That's fixed.

PR #3394, opened 2022-07-23
 Clarify with type annotations that Extension is a kind of service locator

tamuratak:
Clarify with type annotations that Extension is a kind of service locator, which makes components testable with fake or mock Extension.

See
- https://martinfowler.com/articles/injection.html#UsingAServiceLocator

Especially, read the section "Using a Segregated Interface for the Locator".

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -54,6 +54,7 @@
         "@typescript-eslint/semi": ["error", "never", { "beforeStatementContinuationChars": "always" }],
         "@typescript-eslint/type-annotation-spacing": ["error", { "before": false, "after": true, "overrides": { "arrow": { "before": true, "after": true }}}],
         "@typescript-eslint/consistent-type-assertions": ["error", { "assertionStyle": "as", "objectLiteralTypeAssertions": "never" }],
+        "@typescript-eslint/no-empty-interface": [ "error", { "allowSingleExtends": true } ],
         "@typescript-eslint/no-floating-promises": "error",
         "@typescript-eslint/no-invalid-void-type": "error",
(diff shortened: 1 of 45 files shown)


James-Yu:
This change might be good for testing, but it introduces tedious steps for development.

For example, when I want to inspect into the `getDirtyConent()` function below and see how it is implemented
https://github.com/James-Yu/LaTeX-Workshop/blob/c04cb6ff9cd6e79b7591b5b5fadd3b5572123095/src/providers/structure.ts#L145
Hovering on the function yields
Screen Shot 2022-07-26 at 10 30 24
cmd+click yields
Screen Shot 2022-07-26 at 10 31 05
cmd+click again yields
Screen Shot 2022-07-26 at 10 31 30
You need to find out the correct one and double click to get the function comments and implementations.

This essentially breaks the hinting and intellisense of typescript during development.

Suppose some time later, this `structure.ts` needs to be amended,
Screen Shot 2022-07-26 at 10 33 30
this is the only suggestions provided.

Any workaround?

James-Yu:
Looked around on DI and IoC and found the following articles:

https://medium.com/@samueleresca/inversion-of-control-and-dependency-injection-in-typescript-3040d568aabe
https://khalilstemmler.com/articles/tutorials/dependency-injection-inversion-explained/
https://dev.to/jefftian/inversion-of-control-implementation-based-on-typescript-3fl6

It seems that the interfaces need serious documentations. Will look into it soon.

Issue resolved.

tamuratak:
> when I want to inspect into the getDirtyConent()

Please use `Go to Implementations` or `Peek Implementations`

- https://code.visualstudio.com/docs/editor/editingevolved#_go-to-implementation

PR #3396, opened 2022-07-26
Add pdflatex-only compiler

riccardotosi:
This PR adds compilation with pdflatex only. I am working on Windows using MiKTeX without Perl and this recipe works well. Not sure if this was considered before or not but I couldn't find info [here](https://github.com/James-Yu/LaTeX-Workshop/wiki/Compile#misc).

### DIFF
--- a/package.json
+++ b/package.json
@@ -824,6 +824,12 @@
               "tools": [
                 "tectonic"
               ]
+            },
+            {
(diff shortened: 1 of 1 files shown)


James-Yu:
This is a personalized config setting, therefore won't be merged unfortunately. Thank you for the PR.

PR #3426, opened 2022-08-11
Add .dtx file type while inferring a language ID

T0nyX1ang:
The modified code can infer a dtx file automatically, so it may help users to create their own recipes to compile dtx files instead of encountering a wrong message saying `Cannot find LaTeX root file`.

### DIFF
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -280,6 +280,8 @@ export class Manager implements IManager {
             return 'jlweave'
         } else if (this.rsweaveExt.includes(ext)) {
             return 'rsweave'
+        } else if (ext === '.dtx') {
+            return 'doctex'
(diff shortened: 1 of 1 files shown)


tamuratak:
@jlelong Please check this PR is fine when you have time.

jlelong:
This PR does not harm but it is not really sufficient to compile `.dtx` files with the recipe mechanism. 

Only the method `Manager.findRootFromActive` can detect a `.dtx` file as a root file. Moreover, if we really want to support `doctex`, we should probably add a case for `doctex` in https://github.com/James-Yu/LaTeX-Workshop/blob/dc3283bf17221ec130c86662333e487edf38d396/src/components/builder.ts#L425-L442

PR #3456, opened 2022-09-06
To enable `-shell-escape` for engine `xelatex`?

likanzhan:
To enable `-shell-escape` for engine `xelatex`?


jlelong:
We will never enable `shell-escape` per default as it creates a security hole by allowing the execution of arbitrary code from the LaTeX file.

Issue #3470, opened 2022-09-10
Add custom shell config option to use for builds


xubuntutux:
Invoking a build command launches a child process with `cp.spawn()` [builder.ts#L273](https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/components/builder.ts#L273) or [builder.ts#L282](https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/components/builder.ts#L282).  On *NIX, this is always `/bin/sh` per [node.js docs](https://nodejs.org/api/child_process.html#child_processspawncommand-args-options).  It is desirable to allow the user to specify an alternative shell via a setting available in `settings.json` to use instead of the default.

James-Yu:
Can you please provide a comprehensive use case on the necessity of defining shell? Thanks.

xubuntutux:
The TeX Live installation, including a corporate provided texmf tree, exists inside a Singularity container.  Both the TeX Live build commands and Singularity tools are loaded via [Lmod](https://github.com/TACC/Lmod).  Thus, the build process normally goes this like:

```sh
module load singularity
module load texlive
# At this point, latexmk and company are available from the command prompt;
# however, latexmk first calls `singularity exec <sif image>`
# where singularity is also defined by the Lmod module

# Build commands executed via LaTeX-Workshop
latexmk <%DOC%>
```

A normal shell invoked with `/bin/sh -c` as is done by Node.js will not allow proper loading of the Lmod modules.  Thus, there is a need to specify an alternate shell by which preloading the modules can be achieved.

jlelong:
What about using [external build command](https://github.com/James-Yu/LaTeX-Workshop/wiki/Compile#external-build-command). You could launch the correct shell inside your script.

Btw, I think the right solution is to load the modules prior to launching vscode such that any sub-shell spawned from inside vscode would inherit the correct paths.

xubuntutux:
Yes, an external build command could work for one of the builds, but we have several types for a single project depending on whether we need extra/alternate flags passed to `latexmk`, as well as whether we clean (`latexmk -C`) prior to building or not.  Thus, we have a need for several recipes.

[...]

PR #3480, opened 2022-09-14
Add a new `latex-workshop.shell` config to define shell of `spawn`

James-Yu:
This PR adds a new `latex-workshop.shell` config to control the shell option of all `child_process.spawn` (in fact, `cross-spawn.spawn`).

Currently we may have an issue on Linux compiling, working on it now.

### DIFF
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -270,7 +270,6 @@ export class Builder implements IBuilder {
                 command += ' ' + args[0]
             }
             this.extension.logger.addLogMessage(`cwd: ${path.dirname(rootFile)}`)
-            // The shell option below should be true. Cannot recall why. Need to check. @James @22-Sep-14
             this.currentProcess = cs.spawn(command, [], {cwd: path.dirname(rootFile), env: envVars, shell: getShell() || true})
         } else {
             let workingDirectory: string
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -270,7 +270,8 @@ export class Builder implements IBuilder {
                 command += ' ' + args[0]
             }
             this.extension.logger.addLogMessage(`cwd: ${path.dirname(rootFile)}`)
-            this.currentProcess = cs.spawn(command, [], {cwd: path.dirname(rootFile), env: envVars, shell: getShell()})
+            // The shell option below should be true. Cannot recall why. Need to check. @James @22-Sep-14
+            this.currentProcess = cs.spawn(command, [], {cwd: path.dirname(rootFile), env: envVars, shell: getShell() || true})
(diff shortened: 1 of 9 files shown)


src/components/builder.ts
```diff
@@ -270,7 +270,7 @@ export class Builder implements IBuilder {
                 command += ' ' + args[0]
             }
             this.extension.logger.addLogMessage(`cwd: ${path.dirname(rootFile)}`)
-            this.currentProcess = cs.spawn(command, [], {cwd: path.dirname(rootFile), env: envVars, shell: true})
+            this.currentProcess = cs.spawn(command, [], {cwd: path.dirname(rootFile), env: envVars, shell: getShell()})
```
jlelong, line 273: The function `getShell` returns `false` if no specific shell is specified. it changes the default behaviour.

The `command` variables holds the complete command + its arguments, so we must pass `shell: true` to make it work;

James-Yu: Exactly same time! Fixed in the new commit. May need to add a small note here.

jlelong:
I am concerned with the following two comments 

- From child_process: If the shell option is enabled, do not pass unsanitized user input to this function. Any input containing shell metacharacters may be used to trigger arbitrary command execution.
- From cross_spawn: https://www.npmjs.com/package/cross-spawn#optionsshell-support. While cross-spawn adds support for options.shell in node <v4.8, all of its enhancements are disabled. This mimics the Node.js behavior. More specifically, the command and its arguments will not be automatically escaped nor shebang support will be offered. This is by design because if you are using options.shell you are probably targeting a specific platform anyway and you don't want things to get into your way.

src/components/builder.ts
```diff
@@ -270,7 +270,8 @@ export class Builder implements IBuilder {
                 command += ' ' + args[0]
             }
             this.extension.logger.addLogMessage(`cwd: ${path.dirname(rootFile)}`)
-            this.currentProcess = cs.spawn(command, [], {cwd: path.dirname(rootFile), env: envVars, shell: getShell()})
+            // The shell option below should be true. Cannot recall why. Need to check. @James @22-Sep-14
```
jlelong, line 273: This is because, magic comments define the command + its arguments in a single variable

https://github.com/James-Yu/LaTeX-Workshop/wiki/Compile#magic-comments

James-Yu:
> I am concerned with the following two comments.

I agree. My second thought is, enabling custom shell is more or less equal to making all recipes `latex.external.build`. Maybe I try come up with a better solution to https://github.com/James-Yu/LaTeX-Workshop/issues/3470#issuecomment-1243763407.

Issue #3570, opened 2022-12-03
Running any recipe from the TeX badge actually calls the default one in addition when buildOnFileChange is used


jlelong:
@James-Yu This is a follow up on #3568  

There is still an issue when using build on file change. The call https://github.com/James-Yu/LaTeX-Workshop/blob/cc9186f3e6caaa2fa0dc315173cc880dca999789/src/components/builder.ts#L179 actually triggers a build with the default recipe because of https://github.com/James-Yu/LaTeX-Workshop/blob/7281fb04a219941219fa1edb5ed4ca09e61f7c95/src/components/manager.ts#L889

To reproduce
- Consider a modified unsaved tex editor
- Run a recipe from the TeX badge
- Observe that two builds are triggered (one with the correct recipe and one with the default one)

See the logs where the correct recipe is  `latexmk (latexmkrc)` and the default one  `latexmk 🔃`

```
[16:35:19] RECIPES command invoked.
[16:35:19] BUILD command invoked.
[16:35:19] The document of the active editor: file:///Users/jl/devel/LaTeX-Workshop/mytests/poo.tex
[16:35:19] The languageId of the document: latex
[16:35:19] Current workspace folders: ["file:///Users/jl/devel/LaTeX-Workshop/mytests"]
[16:35:19] Found root file from active editor: /Users/jl/devel/LaTeX-Workshop/mytests/poo.tex
[16:35:19] Keep using the same root file: /Users/jl/devel/LaTeX-Workshop/mytests/poo.tex
[16:35:19] Building root file: /Users/jl/devel/LaTeX-Workshop/mytests/poo.tex
[16:35:19] Build root file /Users/jl/devel/LaTeX-Workshop/mytests/poo.tex
[16:35:19] onDidSaveTextDocument triggered: file:///Users/jl/devel/LaTeX-Workshop/mytests/poo.tex
[16:35:19] outDir: /Users/jl/devel/LaTeX-Workshop/mytests
[16:35:19] Preparing to run recipe: latexmk (latexmkrc)
[16:35:19] Recipe step 1: latexmk
[16:35:19] Recipe step 1 args: ["/Users/jl/devel/LaTeX-Workshop/mytests/poo"]
[16:35:19] Recipe step env: {}
[16:35:19] Recipe step root file: /Users/jl/devel/LaTeX-Workshop/mytests/poo.tex
[16:35:19] cwd: /Users/jl/devel/LaTeX-Workshop/mytests
[...]

PR #3572, opened 2022-12-05
Revert to use delay to prevent simultaneous auto build by file save or change

### LINKED ISSUE #3568: Running any recipe from the TeX badge may actually call the default one when buildOnsave is set
### Disable all the other extensions except for LaTeX Workshop, restart VS Code, and check that you still see this issue. [Required]

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

Minimal working example:
```
\documentclass{article}
\usepackage{fontspec}
\begin{document}
    Some text.
\end{document}
```

When compiling the above with `Recipe: latexmk (lualatex)` the "Recipe terminated with error." and "Recipe terminated with error. Retry building the project." messages appear. However, the document appears to be compiled correctly, despite of the messages. Moreover, if the document is compiled for a second time (without edits), no error messages appear. The problem is that, because of these errors, the whole document is compiled every time there is an edit, making long documents take too long to compile. I believe this started to happen recently, maybe due to a recent update?

### To Reproduce

1. Consider the above minimal working example.
2. Go to LaTeX extension > Commands > Build LaTeX project > Recipe: latexmk (lualatex).
3. See the error messages.

### Expected behavior

No errors should appear. The whole document shouldn't be compiled every time, especially under small edits in `\include{filename}` files.

## Logs [Required]

```
Win CP console initial and current in/out Win: (437, 65001), (1252, 1252)
Coding system for system and terminal: 'CP1252'
---
Rc files read:
  NONE
Latexmk: This is Latexmk, John Collins, 17 Mar. 2022. Version 4.77, version: 4.77.
Latexmk: applying rule 'pdflatex'...
Rule 'pdflatex': File changes, etc:
   Changed files, or newly in use since previous run(s):
  c:/Users/Carlos/Desktop/New folder (2)/main.tex
Rule 'pdflatex': The following rules & subrules became out-of-date:
  pdflatex
------------
Run number 1 of rule 'pdflatex'
------------
------------
Running 'pdflatex  --max-print-line=10000 -synctex=1 -interaction=nonstopmode -file-line-error -recorder  "c:/Users/Carlos/Desktop/New folder (2)/main.tex"'
------------
This is pdfTeX, Version 3.141592653-2.6-1.40.24 (MiKTeX 22.10) (preloaded format=pdflatex.fmt)
 restricted \write18 enabled.
entering extended mode
(c:/Users/Carlos/Desktop/New folder (2)/main.tex
LaTeX2e <2022-06-01> patch level 5
L3 programming layer <2022-10-26> (C:\Users\Carlos\AppData\Local\Programs\MiKTeX\tex/latex/base\article.cls
Document Class: article 2021/10/04 v1.4n Standard LaTeX document class
(C:\Users\Carlos\AppData\Local\Programs\MiKTeX\tex/latex/base\size10.clo)) (C:\Users\Carlos\AppData\Local\Programs\MiKTeX\tex/latex/fontspec\fontspec.sty (C:\Users\Carlos\AppData\Local\Programs\MiKTeX\tex/latex/l3packages/xparse\xparse.sty (C:\Users\Carlos\AppData\Local\Programs\MiKTeX\tex/latex/l3kernel\expl3.sty (C:\Users\Carlos\AppData\Local\Programs\MiKTeX\tex/latex/l3backend\l3backend-pdftex.def)))

C:\Users\Carlos\AppData\Local\Programs\MiKTeX\tex/latex/fontspec\fontspec.sty:45: Fatal Package fontspec Error: The fontspec package requires either XeTeX or
(fontspec)                      LuaTeX.
(fontspec)                      
(fontspec)                      You must change your typesetting engine to,
(fontspec)                      e.g., "xelatex" or "lualatex" instead of
(fontspec)                      "latex" or "pdflatex".

Type <return> to continue.
 ...                                              
                                                  
l.45 \msg_fatal:nn {fontspec} {cannot-use-pdftex}
                                                 

Latexmk: If appropriate, the -f option can be used to get latexmk
  to try to force complete processing.
Reverting Windows console CPs to (in,out) = (437,65001)
Latexmk: Getting log file 'main.log'
Latexmk: Examining 'main.fls'
PWD line not in UTF-8
Latexmk: Examining 'main.log'
Latexmk: Errors, so I did not complete making targets
Collected error summary (may duplicate other messages):
  pdflatex: Command for 'pdflatex' gave return code 1
      Refer to 'main.log' for details

```

## Desktop [Required]

 - OS: Windows 11
 - VS Code version: 1.73.1
 - Extension version: 9.1.0
 - TeX distribution version: MiKTeX 22.10

James-Yu:
This PR should fix #3568 and #3570.

The issue rises as the `onDidSaveTextDocument` emits asynchronously with `saveAll` in the builder, so that the event listener may escape the screening of `saving` variable. Same thing also happened to the `.on('changed')` event of `chokidar`.

In commits d91bd72b5fbfdc3499ab6618459eefb193c07209 and 785b6923c117a9e896399afe12e2a3df70449f03 I tried to temporarily remove both listeners during file-saving. This works for the vscode `onDidSaveTextDocument`, which seems to invoke the event before returning `saveAll`. However, `chokidar` is still not synchronous.

Therefore, this PR uses the old-style delay-after-save to prevent the issue. As long as either events will be triggered within 500ms after the saving function returns, we are good.

I am yet to find a better and more elegant way to handle the issue. I myself don't like hard-coded delays....

### DIFF
--- a/src/components/builder.ts
+++ b/src/components/builder.ts
@@ -9,11 +9,10 @@ import { replaceArgumentPlaceholders } from '../utils/utils'
 import { BuildFinished } from './eventbus'
 
 export class Builder {
-    private lastAutoBuild: number = 0
+    private lastBuild: number = 0
     private prevLangId: string | undefined
     private prevRecipe: Recipe | undefined
     private building: boolean = false
-    private saving: boolean = false
     private process: cp.ChildProcessWithoutNullStreams | undefined
 
     private readonly isMiktex: boolean = false
@@ -74,7 +73,6 @@ export class Builder {
      * queue. After that, this function tries to initiate a {@link buildLoop} if
      * there is no one running.
      *
-     * @param type Whether the build is triggered manually or via auto-build.
      * @param command The external command to be executed.
      * @param args The arguments to {@link command}.
      * @param pwd The current working directory. This argument will be overrided
@@ -82,25 +80,19 @@ export class Builder {
      * folder is used as the current working directory.
      * @param rootFile Path to the root LaTeX file.
      */
-    async buildExternal(type: 'manual' | 'auto', command: string, args: string[], pwd: string, rootFile?: string) {
+    async buildExternal(command: string, args: string[], pwd: string, rootFile?: string) {
         if (this.building) {
             void this.extension.logger.showErrorMessageWithCompilerLogButton('Please wait for the current build to finish.')
             return
         }
 
-        if (!this.canBuild(type)) {
-            this.extension.logger.addLogMessage('Auto Build Run is temporarily disabled for `latex.autoBuild.interval`.')
-            return
-        }
-        if (type === 'auto') {
-            this.lastAutoBuild = Date.now()
-        }
+        this.lastBuild = Date.now()
 
         if (rootFile) {
             this.extension.manager.ignorePdfFile(rootFile)
         }
 
-        await this.saveAll()
+        await vscode.workspace.saveAll()
 
         const workspaceFolder = vscode.workspace.workspaceFolders?.[0]
         const cwd = workspaceFolder?.uri.fsPath || pwd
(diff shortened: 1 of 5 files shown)


src/components/builder.ts
```diff
@@ -168,20 +167,20 @@ export class Builder {
...
     }
 
     private async saveAll() {
-        await this.extension.removeBuildOnSaveEvent(this.buildOnSaveEvent)
-        this.extension.manager.unregisterbuildOnFileChanged()
+        this.saving = true
         await vscode.workspace.saveAll()
-        this.buildOnSaveEvent = this.extension.createBuildOnSaveEvent()
-        this.extension.manager.registerbuildOnFileChanged()
+        setTimeout(() => this.saving = false, 500)
```
jlelong, line 183: Maybe we could use `latex-workshop.latex.autoBuild.interval` instead of `500` here.

James-Yu: I agree. The new implementation should make the whole logic more clear.

James-Yu:
In 840343a, the config `latex.autoBuild.interval` is used for both auto-builds triggered by users and by file saves from previous building processes. To facilitate this implementation, the definition of this config is tweaked: auto builds must not trigger within the milliseconds from the previous build (manual or auto), where previously it was the interval between auto builds. I checked the original implementation before the refactor, and discovered that the current logic is indeed consistent.

jlelong:
I have successfully run a few different test cases. This PR looks good to me.

PR #3579, opened 2022-12-16
Optional argument intellisense

### LINKED ISSUE #2098: Completion for options of minted package
I have used this extension quite a lot in the past months and I'd like to give back a little!
I'm still pretty new to VS Code extension development, so maybe I'm misjudging the scope of my idea. Please let me know your thoughts!

What I have in mind is completion support for the options of the minted package. I figured this might be an easy thing to contribute. I.e.,

```latex
\inputminted[firstli<ctrl+space>
```

would then offer to complete `firstli` to `firstline=` and do the same thing for the other macro options in the minted package.

Thanks for this wonderful extension and again, hit me with your feedback, please!

James-Yu:
This PR provides the intellisense feature for optional arguments, also called keyvals by TeXStudio.

For demo, the optional arguments of `minted`package are added, which is parsed from the cwl file of TeXStudio. This closes #2098.

Note that the `minted.cwl` is not obtained from https://github.com/LaTeXing/LaTeX-cwl but instead https://github.com/texstudio-org/texstudio/blob/master/completion/minted.cwl . This means that:
- https://github.com/jlelong/LaTeX-cwl/ may need updates on any packages whose optional arguments are to appear in the extension.
- I've consulted license experts and confirmed that incorporating the information instead of code (in `cwl` form for our case) does not cause GPL to override MIT.


jlelong:
> Note that the `minted.cwl` is not obtained from https://github.com/LaTeXing/LaTeX-cwl but instead https://github.com/texstudio-org/texstudio/blob/master/completion/minted.cwl . This means that:
> 
> * https://github.com/jlelong/LaTeX-cwl/ may need updates on any packages whose optional arguments are to appear in the extension.

I had not really paid attention to the completion directory of `texstudio` but  the `.cwl` files from `texstudio` are much more up-to-date than those in https://github.com/LaTeXing/LaTeX-cwl. I think we could simply drop https://github.com/jlelong/LaTeX-cwl/ and directly get the `.cwl` files from `texstudio`. Yet, it is not easy to extract a subdirectory from a github repo as github does not allow the use of `git-archive`.

PR #3588, opened 2022-12-21
Pull cwl from texstudio-org/texstudio

James-Yu:
This PR changed the source of `cwl` files from the repo @jlelong maintained to the official one from texstudio-org/texstudio. I am not sure whether this is a good idea though, as the change is buffing the package size of the extension from around ten MB to more than 50. Not sure if cherry picking is a good solution, or a more complex scheme to let users to download aynchronously.


jlelong:
Considering LW is already quite heavy, I do not think this is a good idea to add all these files to the extension. I am afraid that increasing the size of the extension will lead to a lot of complaints/issues and users would probably be right. Note that there are several issues about the size of the package.

The `.cwl` base from `tex-studio-org/studio` looks quite exhaustive and most users will only use a few of these files. We could try to keep only the _core_ ones (related to mostly used packages even though I do not know exactly how to quantify this). We can create a separate repo with all the files to ease asynchronous download.

Btw, the size of LW has doubled between `8.29.0` and `9.1.1`.

James-Yu:
Will look into the size increase very shortly.

jlelong:
I compared the content of the  two `.vsix` files. Here is a summary (sorted in decreasing order) of how much modules or directories (`dir_data`, `dir_out`, `dir_viewer`) have grown by

```
('canvas', 10231507)
('nan', 423681)
('tr46', 268388)
('tar', 188475)
('@mapbox', 166631)
('node-fetch', 152240)
('are-we-there-yet', 136397)
('dir_data', 104648)
('semver', 87418)
('make-dir', 75579)
('rimraf', 55064)
('whatwg-url', 49886)
('glob', 48964)
('gauge', 48549)
('emoji-regex', 47828)
('minipass', 43852)
('debug', 42352)
('string_decoder', 38246)
('nopt', 25840)
('agent-base', 24763)
('detect-libc', 19526)
('minizlib', 17309)
('https-proxy-agent', 17046)
('npmlog', 16566)
('lru-cache', 15643)
('yallist', 14752)
('fs-minipass', 14089)
('console-control-strings', 12669)
('webidl-conversions', 12370)
('signal-exit', 9958)
('color-support', 9233)
('aproba', 8047)
('delegates', 7459)
('mathjax-full', 6936)
('ms', 6842)
('object-assign', 5493)
('util-deprecate', 5481)
('ansi-regex', 4865)
('abbrev', 4782)
('simple-concat', 4624)
('wide-align', 4467)
('is-fullwidth-code-point', 4445)
('string-width', 4369)
('set-blocking', 4224)
('strip-ansi', 3660)
('dir_out', 3623)
('has-unicode', 3436)
('dir_viewer', 3427)
```

PR #3589, opened 2022-12-21
Add output directory for pdflatex

JoepdeJong:
Make sure that `pdflatex` auxiliary files end up in the `out` dir.


James-Yu:
You can always make local changes on the configurations.

PR #3594, opened 2022-12-22
Load cwl from texstudio-org/texstudio, only a few

James-Yu:
This one should be good to go compared to #3588. In this PR, only the already-loaded packages are parsed from the texstudio-org/texstudio source. Later on, we may edit the new `cwl.list` file to determine which packages should be included.


James-Yu:
It seems that TeXStudio had a discussion on `latex-mathsymbols` and split the file to three https://github.com/texstudio-org/texstudio/pull/783. It seems better to adhere to their changes for mantainability?

jlelong:
I agree that to ease maintenance, we should stick to the way TeXStudio organises its `.cwl` files.

jlelong:
I have some basic (maybe naive) questions regarding the content of the `.json` files. 

- The two fields `keyvals` and `keyvalindex` work together. They are most of the time empty, which increases the sizes of the `.json` files with no reason. What about only keeping them when they are not empty? It seems that the completer code is already almost ready for this https://github.com/James-Yu/LaTeX-Workshop/blob/e4a9276b7c13b6ccbfc5caf1fbcf72befc605751/src/providers/completer/command.ts#L17-L18
- The `option` field seems not used in the completion code. Maybe, we could remove it from the `.json` file? At least, when it is empty What about using an option in the `python` machinery.
- The `keyvals` field is described as `{key: string, snippet: string}[]` in https://github.com/James-Yu/LaTeX-Workshop/blob/e94680b791e122bf2f2e0dc7b9add548e1a9be5f/dev/README.md#L78 but as `string[]` in https://github.com/James-Yu/LaTeX-Workshop/blob/e4a9276b7c13b6ccbfc5caf1fbcf72befc605751/src/providers/completer/command.ts#L18

James-Yu:
> * The two fields `keyvals` and `keyvalindex` work together. They are most of the time empty, which increases the sizes of the `.json` files with no reason. What about only keeping them when they are not empty? It seems that the completer code is already almost ready for this https://github.com/James-Yu/LaTeX-Workshop/blob/e4a9276b7c13b6ccbfc5caf1fbcf72befc605751/src/providers/completer/command.ts#L17-L18

I totally agree! Will do the change.

> * The `option` field seems not used in the completion code. Maybe, we could remove it from the `.json` file? At least, when it is empty What about using an option in the `python` machinery.

This option can be included with ease, but I forgot it 😓 Will do shortly.

> * The `keyvals` field is described as `{key: string, snippet: string}[]` in https://github.com/James-Yu/LaTeX-Workshop/blob/e94680b791e122bf2f2e0dc7b9add548e1a9be5f/dev/README.md#L78
>    but as `string[]` in https://github.com/James-Yu/LaTeX-Workshop/blob/e4a9276b7c13b6ccbfc5caf1fbcf72befc605751/src/providers/completer/command.ts#L18

Fixed!

James-Yu:
From 2190430b to 30b06775, the vsix size is reduced from 9.88MB to 9.74MB.

jlelong:
> From [2190430](https://github.com/James-Yu/LaTeX-Workshop/commit/2190430b28b069f40039cc5161241a7e3aad03a2) to [30b0677](https://github.com/James-Yu/LaTeX-Workshop/commit/30b06775415961d9298d303434db97bcbaa00793), the vsix size is reduced from 9.88MB to 9.74MB.

Even this looks a small gain, I still think it is worth it. JSON files are very much compressed inside vsix bundles but they take more place once installed and more importantly from a maintenance point of view shorter files are easier to navigate.

James-Yu:
> > From [2190430](https://github.com/James-Yu/LaTeX-Workshop/commit/2190430b28b069f40039cc5161241a7e3aad03a2) to [30b0677](https://github.com/James-Yu/LaTeX-Workshop/commit/30b06775415961d9298d303434db97bcbaa00793), the vsix size is reduced from 9.88MB to 9.74MB.
> 
> Even this looks a small gain, I still think it is worth it. JSON files are very much compressed inside vsix bundles but they take more place once installed and more importantly from a maintenance point of view shorter files are easier to navigate.

I totally agree. Also by removing the canvas package, the reduction looks more notable. Will do some tests and merge shortly.

Issue #1932, opened 2020-02-04
latexmk -output-directory=fullpath -cd causes makeindex to fail


phdoerfler:
### Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue.

As far as I can see this is completely unrelated to other extensions.

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki? Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually? Yes

## Describe the bug

I can compile a TeX document manually using this command line:

```shell
latexmk -synctex=1 -interaction=nonstopmode -file-line-error -pdf -outdir=/Users/phi/Documents/workspace/FAUGitlab/spiht/Documentation
```

Clean auxiliary files and add `-cd` (which is what LaTeX-Workshop does) and it no longer builds:

```shell
latexmk -synctex=1 -interaction=nonstopmode -file-line-error -pdf -outdir=/Users/phi/Documents/workspace/FAUGitlab/spiht/Documentation -cd /Users/phi/Documents/workspace/FAUGitlab/spiht/Documentation/documentation
```

In both cases I invoked this from the directory where `.tex` file resides and which is also the `outdir`.
Unfortunately LaTeX-Workshop uses the `-cd` argument of `latexmk`.

### Expected behavior

I expect LaTeX-Workshop to not use the `-cd` argument, but to invoke `latexmk` with the correct working directory set.
`child_process.spawn` of node.js allows to specify the `cwd`.

### LaTeX Workshop Output

https://gist.github.com/phdoerfler/e38fa3bb2a5fe189e31351317dd15a77

### Developer Tools Console

I don't see anything suspicious there, but here you go:

```
/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:3930 Overwriting grammar scope name to file mapping for scope source.scala.
Old grammar file: file:///Users/phi/.vscode/extensions/daltonjorge.scala-0.0.5/syntaxes/Scala.tmLanguage.
New grammar file: file:///Users/phi/.vscode/extensions/scala-lang.scala-0.3.9/syntaxes/Scala.tmLanguage.json
register @ /Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:3930
[...]

PR #3604, opened 2022-12-28
Refactor test, all

James-Yu:
This PR refactors the current CI and unit tests, but leave those problematic ones untouched.

A major change is to use as least number of vscode instances as possible, so that the testing time can be dramatically decreased. This is achieved by dynamically creating and deleting test `tex` files.

This PR also goes through existing tests and validates their validity. Some are not correct or do not cover all use cases any more.

By this PR, the testing time on Linux is reduced from the 8m0s of c55dd84 to 4m4s of 95c86d88c64abc3455b632ae3fceb62921c29245 with possibility of further reduction. The error rate is also notably reduced theoretically, which may need later observations.


src/components/builder.ts
```diff
@@ -235,6 +235,9 @@ export class Builder {
         } else if (!step.isExternal) {
             if (step.command === 'latexmk' && step.rootFile === this.extension.manager.localRootFile && this.extension.manager.rootDir) {
                 cwd = this.extension.manager.rootDir
+                if (step.args && !step.args.includes('-cd')) {
+                    step.args.push('-cd')
+                }
```
jlelong, line 240: We have had a lot of troubles with the `-cd` option of `latexmk`. See for instance #1932
Why do we need to enforce `-cd` here?

James-Yu: Without enforcing `-cd`, the test `build sub.tex to outdir` will fail. This test has a structure like this:
- main.tex: `\subfile{sub/s}`
- sub/s.tex: `\input{./subsub/infile`, `\documentclass[../main.tex]{subfiles}`
- sub/subsub/infile.tex

As `sub/s.tex` is a `subfile`, it should be able to compile independently. So the `input` should receive `./subsub/infile` which is a relative path to `sub/s.tex`.

However, without a `-cd`, the compilation initiated from `sub/s.tex` will fail, as the cwd is `./` instead of `./sub/`

`cwd = this.extension.manager.rootDir`

I have no idea why this test passed in the previous test implementation. But I think the current behavior and fail is reasonable.

As to #1932, I will write a quick test to see how to work around it.

James-Yu: I've added a new test on #1932 with subfile and makeindex, compiled with latexmk. The test passed with `-cd`! Maybe `latexmk` has fixed the bug? https://github.com/James-Yu/LaTeX-Workshop/issues/1932#issuecomment-583258858

James-Yu: I should have noted that the new test on `makeindex` with `-cd` is `build sub.tex to outdir with makeindex` in `01_build`. Will add the issue no. to the test name.

James-Yu:
The testing time on Linux is finalized to 2m11s of [d01b7e2](https://github.com/James-Yu/LaTeX-Workshop/pull/3604/commits/d01b7e2c8eef7e312dbb68943a6830fe173ed89f), only 27% of https://github.com/James-Yu/LaTeX-Workshop/commit/c55dd846cca3da21fca0d8f3a7c68a93155fcad8

Issue #3591, opened 2022-12-21
PDF viewer won't refresh or jump after returning from suspend


edemaine:
### Disable all the other extensions except for LaTeX Workshop, restart VS Code, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: n/a

## Describe the bug [Required]

Just like #1804 and #1304, the connection to the PDF viewer seems to be broken after my Windows 11 machine comes back from suspend state. When the PDF changes (e.g. from building in LaTeX Workshop), the PDF window does not refresh. SyncTeX jumping doesn't work in either direction.

The logs below show that LaTeX Workshop attempts to reconnect, but fails.

### To Reproduce

1. Open a LaTeX project and PDF file.
2. Suspend machine (probably needs to be Windows).
3. Modify LaTeX and build PDF.

### Expected behavior

I expect the PDF window to stay up-to-date with the PDF file.

### Workarounds

Closing and re-opening the PDF window causes it to start working again.

## Logs [Required]

Please paste the whole log messages here, not parts of ones. The log should start with `Initializing LaTeX Workshop`. It is very important to identify problems.

### LaTeX Workshop Output [Required]

In the logs below, the wake-from-suspend occurs at 14:30. The logs beforehand are likely not relevant, but included for completeness.

My project path within my home directory has been abbreviated to `...` for security.

```
To access the log, click the 'TeX' icon on the Activity Bar on the left side, select 'View Log Messages', then select 'View LaTeX Workshop extension log'.

[13:25:29] Initializing LaTeX Workshop.
[13:25:29] Extension root: c:\Users\edemaine\.vscode\extensions\james-yu.latex-workshop-9.1.1
[...]

PR #3607, opened 2022-12-30
Fixed sleep bug

obsgolem:
As documented in #3591, #1812 did not solve all the issues with reconnection. I can reproduce consistently on my machine simply by going to sleep and waking. Perhaps we could solve this by extending the timeout instead, but it seems more robust to just repeatedly reconnect until we succeed. This usually occurs within 2-4 attempts for me. Perhaps this has something to do with the time from waking to unlocking the computer.

Additionally, there is a bug in the ping code which causes it to repeatedly produce connection error messages. This was also causing us to leak `WebSocketPort` objects, as the interval was keeping them alive. Clearing the interval should solve that issue as well.

### DIFF
--- a/viewer/latexworkshop.ts
+++ b/viewer/latexworkshop.ts
@@ -409,23 +409,26 @@ class LateXWorkshopPdfViewer implements ILatexWorkshopPdfViewer {
 
             // Since WebSockets are disconnected when PC resumes from sleep,
             // we have to reconnect. https://github.com/James-Yu/LaTeX-Workshop/pull/1812
-            const reconnect = (i: number) => async () => {
-                console.log('Try to reconnect to LaTeX Workshop.')
-                this.connectionPort = createConnectionPort(this)
+            const reconnect = (tries: number) => async () => {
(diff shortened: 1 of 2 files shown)


viewer/latexworkshop.ts
```diff
@@ -409,13 +409,14 @@ class LateXWorkshopPdfViewer implements ILatexWorkshopPdfViewer {
...
             // we have to reconnect. https://github.com/James-Yu/LaTeX-Workshop/pull/1812
-            setTimeout(() => {
+            const id = setInterval(() => {
                 console.log('Try to reconnect to LaTeX Workshop.')
                 this.connectionPort = createConnectionPort(this)
                 this.connectionPort.onDidOpen(() => {
                     document.title = this.documentTitle
                     this.setupConnectionPort()
                     console.log('Reconnected: WebScocket to LaTeX Workshop.')
+                    clearInterval(id)
```
James-Yu, line 419: It would be better to set a maximum retries instead of indefinitely connect. What about every 2+i seconds for i=1..10 times?

James-Yu:
Thanks for the contrib!

PR #3618, opened 2023-01-03
Make all config items effective without reloading vscode

James-Yu:
This PR makes the following config items independent of vscode reload to take effect:

- `intellisense.triggers.latex`
- `intellisense.argumentHint.enabled`
- `intellisense.commandsJSON.replace`
- `latex.watch.usePolling`
- `latex.watch.interval`
- `latex.watch.delay`
- `latex.watch.pdf.delay`

This PR also resolves an issue where the original `latex.pdfWatch.delay` config does not work due to a typo in 897dd1813c6f1f7ef653315adaf0ff9eb909f556

### DIFF
--- a/package.json
+++ b/package.json
@@ -1102,25 +1102,25 @@
           "scope": "window",
           "type": "boolean",
           "default": false,
-          "markdownDescription": "When set to true, polling is used to watch changes on files. When TeX files are placed on network drives or OneDrive, this option should be turned on. Setting this option to true might lead to high CPU utilization. Reload vscode to make any change in this configuration effective."
+          "markdownDescription": "When set to true, polling is used to watch changes on files. When TeX files are placed on network drives or OneDrive, this option should be turned on. Setting this option to true might lead to high CPU utilization."
         },
         "latex-workshop.latex.watch.interval": {
           "scope": "window",
           "type": "number",
           "default": 300,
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -1102,25 +1102,25 @@
...
         },
         "latex-workshop.latex.watch.delay": {
           "scope": "window",
           "type": "number",
           "default": 250,
-          "markdownDescription": "Delay before starting builds, in milliseconds. Reload vscode to make any change in this configuration effective."
+          "markdownDescription": "Delay before starting builds, in milliseconds."
         },
-        "latex-workshop.latex.pdfWatch.delay": {
+        "latex-workshop.latex.watch.pdf.delay": {
```
jlelong, line 1119: Maybe, we could have marked `latex-workshop.latex.pdfWatch.delay` as a deprecated setting to notify users that it has changed. Deprecated settings could be removed when the major release number changes for instance.

James-Yu: A minor issue is do we need a dedicated pdf watch delay or just use the normal one? Looks like 897dd1813c6f1f7ef653315adaf0ff9eb909f556 is for the former, but a typo prevents it from properly functioning.

jlelong: If I remember well, we may need a longer delay for pdf reload to avoid reloading a corrupted pdf file. It is probably better to keep a separate setting.

PR #3630, opened 2023-01-07
Fixed sleep bug again

obsgolem:
I have verified that one of the changes you made in #3607 actually broke my fix.

Note that reconnect must be async and didOpen must be awaited in order for reconnect to function. This is because otherwise the exception is raised in a background task rather than in the reconnect task, and is thus lost.

### DIFF
--- a/viewer/latexworkshop.ts
+++ b/viewer/latexworkshop.ts
@@ -417,17 +417,8 @@ class LateXWorkshopPdfViewer implements ILatexWorkshopPdfViewer {
 
             // Since WebSockets are disconnected when PC resumes from sleep,
             // we have to reconnect. https://github.com/James-Yu/LaTeX-Workshop/pull/1812
-            const reconnect = (tries: number) => async () => {
-                console.log(`Try to reconnect to LaTeX Workshop: (${tries}/10).`)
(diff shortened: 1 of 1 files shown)


viewer/latexworkshop.ts
```diff
@@ -417,11 +417,11 @@ class LateXWorkshopPdfViewer implements ILatexWorkshopPdfViewer {
 
             // Since WebSockets are disconnected when PC resumes from sleep,
             // we have to reconnect. https://github.com/James-Yu/LaTeX-Workshop/pull/1812
-            const reconnect = (tries: number) => () => {
+            const reconnect = (tries: number) => async () => {
                 console.log(`Try to reconnect to LaTeX Workshop: (${tries}/10).`)
                 try {
                     this.connectionPort = createConnectionPort(this)
-                    this.connectionPort.onDidOpen(() => {
+                    await this.connectionPort.onDidOpen(() => {
```
James-Yu, line 424: We have a `'await' has no effect on the type of this expression.ts(80007)` warning on this line. It is not advised to use `await` to force raising exceptions. You may need to use explicit try-catch inside.

PR #3650, opened 2023-01-15
New config `intellisense.command.user` to add/change/delete default cmd

James-Yu:
This PR adds a new config `intellisense.command.user`, which is almost an in-place substitution of the original `intellisense.commandsJSON.replace`. The new config shares the same data format of the previous one, and can also change snippets of default commands or remove them by empty snippets. In addition, entries in the new config are added to the command suggestion list, making the config add/change/delete commands compared to the previous change/delete only.

This PR is primarily inspired by #3646 . I cannot recall the original initiative on preventing `intellisense.commandsJSON.replace` to add entries.

There are also other `*JSON.replace` configs. This PR does not touch them, and I am not sure if it is a good idea to make their behavior and name consistent. It seems that the current @-suggestion.replace can already add.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1796,6 +1796,15 @@
           "default": 1000,
           "markdownDescription": "Defines the delay in milliseconds for the extension to update current active file content for intellisense after stopped typing. This config works only when `intellisense.update.aggressive.enabled` is enabled. Lower this value to let the extension know newly defined commands/references/environments more quickly, at the cost of more frequent content parsing: more computational burden."
         },
+        "latex-workshop.intellisense.command.user": {
+          "scope": "resource",
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          },
+          "default": {},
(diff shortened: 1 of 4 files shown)


jlelong (approved):

jlelong:
Concerning the 3 other `*JSON.replace` settings, I would not change their behaviour. 

- The setting `latex-workshop.intellisense.atSuggestionJSON.replace` already allows adding new entries. Its behaviour is similar to the one implemented by this PR. Maybe, we should rename it to `latex-workshop.intellisense.atSuggestion.user` for coherence.
- The settings `latex-workshop.intellisense.biblatexJSON.replace` and `latex-workshop.intellisense.bibtexJSON.replace` are used to change the list of default fields used in a given bibliography entry type. Allowing the user to add new entries would result in declaring new bibliography entry types. This is of course not possible as they would not be recognised by `bibtex` or `biblatex`.

PR #3655, opened 2023-01-18
[DocTeX Syntax] Remove hard-coded whitespace requirement

nvlang:
`%    \begin{macrocode}` is currently only highlighted properly if there's exactly 4 spaces between the `%` and `\begin{macrocode}` parts. The proposed change would make the line be highlighted properly no matter how many space or tab characters are used instead. Same for `%    \end{macrocode}`.

### DIFF
--- a/syntax/DocTeX.tmLanguage.json
+++ b/syntax/DocTeX.tmLanguage.json
@@ -17,7 +17,7 @@
       "name": "meta.function.verb.latex"
     },
     {
-      "begin": "^(%)    (\\\\begin\\{macrocode\\})",
+      "begin": "^(%)[ \\t]*(\\\\begin\\{macrocode\\})",
       "captures": {
         "1": {
           "name": "comment.line.percentage.doctex"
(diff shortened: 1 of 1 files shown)


jlelong:
From https://mirror.ibcp.fr/pub/CTAN/info/dtxtut/dtxtut.pdf, at the bottom of page 16

> There must be exactly four spaces between the `%` and the `\begin{macrocode}` or `\end{macrocode}`. Otherwise, Doc won’t detect the end of the code fragment.

There is also a footnote on page 17

> Only the `\end{macrocode}` needs this precise spacing and then, only for typesetting the documentation. Nevertheless, it’s good practice to use `%    ` for the `\begin{macrocode}`, as well.

It seems to me that the current PR may be misleading to users by letting them think that the four spaces are optional. Unless you have very strong arguments, I will close this PR shortly.

nvlang:
Huh, I didn't realize this was actually a requirement imposed by DocTeX itself, that's 100% my bad. Thank you for your comment!

Issue #3664, opened 2023-01-24
PDF viewer doesn't work


nidrissi:
### Disable all the other extensions except for LaTeX Workshop, restart VS Code, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

## Describe the bug [Required]

The PDF viewer does not work. The internal webview is empty. Opening the URL of the viewer in an external browser produces a `ERR_CONNECTION_REFUSED` error.

### To Reproduce

1. Compile any document.
2. Click on "View PDF".
3. The webview is empty.

### Expected behavior

I should be able to see the PDF.

### LaTeX Workshop Output [Required]

```
[09:37:45][Logger] New log placeholder %WS1% registered for /home/najib/tmp .
[09:37:45][Config] Configuration for workspace: file://%WS1% .
[09:37:45][Config] editor.acceptSuggestionOnEnter: "on" .
[09:37:45][Config] latex-workshop.bind.enter.key: false .
[09:37:45][Config] latex-workshop.docker.enabled: false .
[09:37:45][Config] latex-workshop.docker.image.latex: "" .
[09:37:45][Config] latex-workshop.hover.preview.mathjax.extensions: [] .
[09:37:45][Config] latex-workshop.intellisense.package.enabled: true .
[09:37:45][Config] latex-workshop.intellisense.update.aggressive.enabled: false .
[09:37:45][Config] latex-workshop.intellisense.update.delay: 1000 .
[09:37:45][Config] latex-workshop.latex.autoBuild.run: "never" .
[09:37:45][Config] latex-workshop.latex.build.forceRecipeUsage: true .
[09:37:45][Config] latex-workshop.latex.outDir: "%DIR%" .
[09:37:45][Config] latex-workshop.latex.recipes: [
 {
  "name": "latexmk",
  "tools": [
   "latexmk"
  ]
 },
 {
  "name": "latexmk (latexmkrc)",
  "tools": [
   "latexmk_rconly"
  ]
 },
 {
  "name": "latexmk (lualatex)",
  "tools": [
   "lualatexmk"
  ]
 },
 {
  "name": "latexmk (xelatex)",
  "tools": [
   "xelatexmk"
  ]
 },
 {
  "name": "pdflatex -> bibtex -> pdflatex * 2",
  "tools": [
[...]

PR #3665, opened 2023-01-24
Have the PDF viewer listen on any address

nidrissi:
The PDF viewer only listens on `127.0.0.1`. This causes issues such as #3664. The fix is to allow the viewer's HTTP server to listen on any address, rather than the hardcoded IPv4 address `127.0.0.1`. See https://nodejs.org/api/net.html#serverlisten for more information.


James-Yu:
Considering firewall, you are suggested to set a fixed port for the extension and create an exception for the port.

PR #3674, opened 2023-01-27
Fix synctex with .Rnw formats with patchDVI

LucunJi:
This fixes the synctex issues when compiling .Rnw files:
- Originally, synctex (forward & backward) only works with the generated .tex file.
The generated .tex and *-concordance.tex files are also outside of the `%OUTDIR%` directory.
- With this fix, bidirectional synctex works with the .Rnw file, and all generated files are inside `%OUTDIR%`.

Note: For this fix to work, the user needs to install `patchDVI` R package.
I'm not familiar with VSCode extensions, but you may add some alert box, or use R code to check if the package is installed and possibly use the old compiling code as a fallback.

### DIFF
--- a/package.json
+++ b/package.json
@@ -832,8 +832,7 @@
             {
               "name": "Compile Rnw files",
               "tools": [
-                "rnw2tex",
-                "latexmk"
+                "rnw2tex"
               ]
             },
             {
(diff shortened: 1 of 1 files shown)


James-Yu:
Thanks for the PR. We typically don’t merge config changes unless have to. You can create the tools on your local side, and optionally provide a short instruction in the discussion section of this repository for others to reference.

PR #3676, opened 2023-01-30
add shortcut for `\frac` command

nvlang:


jlelong:
This is not the right to create a shortcut for a snippet.
See https://code.visualstudio.com/updates/v1_9#_insert-snippets

nvlang:
Thank you for the hint, I didn't realize this was possible — sorry about that.

Issue #3687, opened 2023-02-03
Cleaning with latex-workshop.latex.clean.command (latexmk) fails on Windows 11 with MiKTeX


t4rf9:
### Pre-check

- [X] The issue remains after disabling all other extensions and restarting Visual Studio Code.
- [X] The [FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) cannot address the issue.
- [X] If the issue is related to compiling a document, it can be successfully compiled in OS terminal.

### Extension Version

9.6.0

### Operating System and Version

Windows 11

### Visual Studio Code Version

1.74.3

### LaTeX Distribution Version

MiKTeX 23.1

### Extension Environment

- [ ] I am using VSCodium.
- [ ] I am using the Snap or Flatpack version of Visual Studio Code.
- [ ] I am using LaTeX Workshop with Visual Studio Code Remote Containers.
- [ ] I am using LaTeX Workshop with Visual Studio Code Remote SSH.
- [ ] I am using LaTeX Workshop with Visual Studio Code Remote WSL.

### The Issue

With MiKTeX on Windows, cleaning with latex-workshop.latex.clean.command (latexmk) fails.

This was reported in #3679 , but was closed before the issue is solved.
After some investigation, I found that latexmk.exe generated by TeXLive works well but the one by MiKTeX fails.
I presented an issue to MiKTeX, and received an [illustration](https://github.com/MiKTeX/miktex/issues/1278#issuecomment-1415334296).
It is the MiKTeX feature of installing TeX packages on-the-fly that causes the problem.
With the install-on-the-fly option turned on, MiKTeX executables CANNOT be run as a detached process.

So I think it is wise to modify https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/components/cleaner.ts#L160 by setting `detached: false`.
If this will cause any side-effects, I hope we can find a better solution.

Besides, if possible, I suggest adding MiKTeX to unit tests in addition to TeXLive.

### Reproduction Steps

1. Set latex-workshop.latex.clean.command to `latexmk`
2. Add "-c", "-auxdir=%OUTDIR%/.aux" and "%TEX%" to latex-workshop.latex.clean.args.
3. Compile a tex file with latexmk -auxdir=%OUTDIR%/.aux
4. Click "Clean up auxilliary files"
5. Fails on Windows

### Expected Behavior

The cleaning should succeed.

### Actual Behavior

The cleaning process fails.
[...]

Issue #3679, opened 2023-01-31
Cleaning with latex-workshop.latex.clean.command (latexmk) fails on Windows 11


t4rf9:
### Disable all the other extensions except for LaTeX Workshop, restart VS Code, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: Yes

## Describe the bug [Required]

When I try to configure latex-workshop to clean with 
`latexmk -c -auxdir=%OUTDIR%/.aux %TEX%` (compiled with `latexmk -auxdir=%OUTDIR%/.aux`), 
I added "-c", "-auxdir=%OUTDIR%/.aux" and "%TEX%" to latex-workshop.latex.clean.args.

Then, it works well on macOS 13, but fails on Windows 11.

After investigating the source code, I found the problem comes from https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/components/cleaner.ts#L160,
where `cross-spawn.spawn` is used.
Using `cross-spawn-async.spawn` would work well.
That's because after parsing, `cross-spawn.spawn` directly returns the cleaning command `latexmk`, namely
```
  command: 'latexmk',
  args: [
    '-c',
    '-bibtex',
    '-auxdir=c:/Users/t4rf9/Programs/test/.aux',
    'c:\\Users\\t4rf9\\Programs\\test\\main.tex'
  ],
  options: {
    cwd: 'c:\\Users\\t4rf9\\Programs\\test',
    detached: true,
    windowsVerbatimArguments: true
  }
```
whilst `cross-spawn-async.spawn` returns `cmd.exe` as the command, namely
```
  command: 'C:\\WINDOWS\\system32\\cmd.exe',
  args: [
    '/s',
    '/c',
    '"latexmk "-c" "-bibtex" "-auxdir=c:/Users/t4rf9/Programs/test/.aux" "c:\\Users\\t4rf9\\Programs\\test\\main.tex""'
  ],
  options: {
    cwd: 'c:\\Users\\t4rf9\\Programs\\test',
    detached: true,
    windowsVerbatimArguments: true
  }
```

The latter wraps the cleaning command in a separated CMD, I think that's what makes it a success.
[...]

PR #3688, opened 2023-02-03
set detached: false for cleaning command

t4rf9:
This allows latexmk.exe generated by MiKTeX to run on windows.
#3679 #3687

### DIFF
--- a/src/components/cleaner.ts
+++ b/src/components/cleaner.ts
@@ -157,7 +157,8 @@ export class Cleaner {
         }
         logger.logCommand('Clean temporary files command', command, args)
         return new Promise((resolve, _reject) => {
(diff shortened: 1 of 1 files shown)


James-Yu:
@jlelong Looks like the detached flag was set in 486d474c08becf9ab96a74ca7f7c0551772babe7 . May you still recall the primary reason behind detaching the process?

James-Yu:
@t4rf9 by default, `detached` is `false`. There's no need to explicitly state a `false` value.

t4rf9:
> @t4rf9 by default, `detached` is `false`. There's no need to explicitly state a `false` value.

That's true.
My reason is that in case a future edition adds it back again, the explicit `false` may be a reminder.
Or if you like, of course you can remove the detached option completely, if there is not a reason to keep it true.

PR #3708, opened 2023-02-08
Update environments.json

likanzhan:
Are these typos, or I understand them incorrectly?

### DIFF
--- a/data/environments.json
+++ b/data/environments.json
@@ -15,10 +15,10 @@
   "math": {},
   "displaymath": {},
   "split": {},
-  "array{}": {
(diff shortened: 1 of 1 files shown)


James-Yu:
Meaningless commit message. No PR description.

likanzhan:
When I input the following:
```
\array
```

it is supposed to be 

```
\begin{array}{cols}
    
\end{array}
```

But it turns out to be 

```
\begin{array{}}{cols}
    
\end{array{}}
```

So if I understand correctly, the symbols `{}` should be deleted from `array{}`

PR #3724, opened 2023-02-15
Add configurable default hostname

asztrikx:
Currently there is no option to set default hostname for which the server binds to without using gui. For premade environments it would be vital if one could preconfigure this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1449,6 +1449,12 @@
           "default": "0",
           "markdownDescription": "Define the port to listen on for communicating with the internal viewer. The default value \"0\" means the port is chosen randomly by the application."
         },
+        "latex-workshop.viewer.pdf.internal.hostname": {
+          "scope": "window",
+          "type": "string",
+          "default": "127.0.0.1",
+          "markdownDescription": "Define the hostname to bind to for communicating with the internal viewer."
+        },
         "latex-workshop.view.pdf.internal.synctex.keybinding": {
           "scope": "window",
           "type": "string",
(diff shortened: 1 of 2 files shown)


James-Yu:
#1271

asztrikx:
I'm not into vscode extension development, but how about creating a popup just like when changing the hostname manually from the gui before creating the server if a different hostname is defined in the configuration file. The popup would have the same "type confirm" dialog like the manual one has.

With this the security risk would be the same as in the current manual system. If I read #1271 carefully this was the concern against implementing this feature.

James-Yu:
I personally am against this idea. A pop up means yes-by-default, which is quite a security risk to less experienced users.

jlelong:
I do agree that considering the security risk created by changing the hostname, it has to come from a user action and cannot be reduced to an "I agree" popup.

asztrikx:
Seems reasonable. However if other commands with a security risk like this will be introduced, then i would suggest to create a setting for each of them but not apply them by default. They could only be applied by a command like "change hostname" (manual, has a confirm dialog). The downside would be that the confirm dialog would have to include all information that would be seen by executing each command (like change hostname) one by one. Resetting all of them should also be possible like resetting hostname.

Also *maybe* this decision (applying unsecure settings) could be remembered so that next time vscode started you wouldn't have to do this again. This doesn't mean to reapply settings each startup as user only authorized unsecure settings at a specific state. Of course this should also not be a default behaviour and should have a warning at startup if it is remembered.

But currently this isn't necessary as there isn't many unsecure commands like this which have to be applied manually each time.

Issue #3720, opened 2023-02-14
Highlight syntax errors in bibliography with biber


LinqLover:
> Please fill the following fields with a star (*) and provide as much related information as possible.

## Pre-checks*
> Please change the following `[ ]` to `[x]` for confirmation.
- [x] The feature request has not been suggested in this repository.

## The Missed*
> Is your feature request related to a problem? Please provide a clear and concise description of what the problem is.
- I often insert syntax errors in bib files (missing commas between properties of an entry, unescaped accents, missing/additional curly braces, ...). These are hard to debug because I always need to look into the logs or do manual bisect-debugging for finding the buggy entry (i.e., removing half of the entries until the file compiles again).
   Example:
   ```bib
      @article{zingheim2002downturn
   	title	= {In a Downturn Do You Cut Pay, Slash the Workforce or Protect Precious Talent?},
   	author	= {Zingheim, Patricia K and Schuster, Jay R},
   	journal	= {Strategy & Leadership},
   	volume	= {30},
   	number	= {1},
   	pages	= {23--26},
   	year	= {2002},
   	publisher	= {MCB UP Ltd}
   	doi	= {10.1108/10878570210414417}
   }
   ```

## The Solution*
> Please provide a solution you would like to have.
- Highlight syntax errors in the bib file. This could either be done through the grammar by being less forgiving in the parser or by generating annotations from the outputs of the parsers (probably the former is easier as different bibliography backends might create different kinds of logs, but I'm neither an expert in that nor in parser design :D).

## Anything Else?
> Add any other context about the feature request below.
- Thank you for developing and maintaining this awesome extension!

jlelong:
With your exemple I get
<img width="371" alt="Capture d’écran 2023-02-16 à 11 34 08" src="https://user-images.githubusercontent.com/2910140/219341804-beb47041-b8b7-49e2-bc6c-acbac26c9c1d.png">

<img width="468" alt="Capture d’écran 2023-02-16 à 11 36 01" src="https://user-images.githubusercontent.com/2910140/219341824-fa34af06-b5b4-4238-842e-446fdbf2f205.png">

As you can see, the bitex error is reported in the _Problems_ pane et the `doi` line is not highlighted as other lines. The line reported by `bibtex` is wrong but we cannot do anything about it. 

@LinqLover Could you elaborate on what you would like on top of that?

LinqLover:
Hmm, this looks different to me:

[...]

PR #3740, opened 2023-02-22
Add a biber logs parser to populate the problems pane

jlelong:
Related to #3720 

This PR implements a parser for biber logs in order to show related errors and warnings in the problem pane.

As I am not a biber user, the parse might be incomplete but we can always improve it later. 
Biber can use several `.bib` files at the same time, so it is clear which file missing entry warnings should be associated with. I can think of three solutions
1. the tex root file
2. the last bib file read
3. all bib files 

@LinqLover @James-Yu Do you have any opinion?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1999,6 +1999,15 @@
           "default": true,
           "markdownDescription": "Show badbox information in the problems panel."
         },
+        "latex-workshop.message.biberlog.exclude": {
+          "scope": "window",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "default": [],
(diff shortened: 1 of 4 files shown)


James-Yu (approved):
I am good with the changes 🙇 

One thing that comes to me is that these log parsers can be refactored a bit like how linters (chktex, lacheck) were normalized. This should be another PR and need to first have tests covered. Will do later.

LinqLover:
That sounds great! But the change has not yet been deployed @James-Yu right? Is there a manual VSIX bundle that I can try out?

Issue #3645, opened 2023-01-12
Jump between paired symbols


gaoqiangks:
I switched from vim+vimtex to vscode+latexworkshop several weeks ago. I like your extension very much. There's a feature in vimtex that  "when the percentage symbol % is pressed in normal, you can jump to the symbol matching the one under the cursor". Certainly this is a feature built in vim. However, vim only allows you to jump between "(" and ")", or "{" and "}".  vimtex allows you to jump between "\begin{equation}" and "\end{equation}".   Can you please add this feature ?

James-Yu:
I don’t understand what you mean. Please fill in all contents in the feature request template.

jlelong:
I think the command `latex-workshop.navigate-envpair` does what you want. See https://github.com/James-Yu/LaTeX-Workshop/wiki/Environments#navigating-and-selecting

You can even define a shortcut for this command. See at the bottom of page https://code.visualstudio.com/docs/getstarted/keybindings

gaoqiangks:
> I think the command `latex-workshop.navigate-envpair` does what you want. See https://github.com/James-Yu/LaTeX-Workshop/wiki/Environments#navigating-and-selecting
> 
> You can even define a shortcut for this command. See at the bottom of page https://code.visualstudio.com/docs/getstarted/keybindings

Thank you very much!

TryerGit:
> I think the command `latex-workshop.navigate-envpair` does what you want

This does work for some environment pairs. In particular this does not work for the following structures/packages:

```
\usepackage{ifthen} 
\newif\ifanswer 
```

The typical usage then within the `.tex` is:

```tex
\begin{document}
\usepackage{ifthen} 
\newif\ifanswer 
\answertrue % this produces the answer key, set this to \answerfalse for the qQuestion paper only with no answers

Question : How many hours are in a day?
\ifanswer
    24
\else
\bigskip
\fi
\end{document}
```

From cursor on `\ifanswer`, it would be great if it were possible to navigate to `\else` and then subsequently to `\fi` and then cycle back.

It may not be easy to implement -- I do not know. A Vim extension manages to accomplish this for LaTeX documents. Please see [here](https://github.com/lervag/vimtex/issues/2609#event-8116014295)

jlelong:
For the time being, the command `latex-workshop.navigate-envpair` only works for environments, ie `\begin{...} ... \end{...}` constructions.

jlelong:
[...]

PR #3770, opened 2023-03-11
Refactor EnvPair module

jlelong:
This PR completely rewrites the `EnvPair` module. We get rid of the old regex logic and use the `latexParser` to build an AST `CommandPair` of all the pairs in the current file.

- It solves #3645 by taking into account pairs of commands in addition to `\begin`/ `end`, inline math and display math
- It solves #3744 
   - _LaTeX Workshop: Select the current environment content_ selects only the inner content of the current environment
   - _LaTeX Workshop: Select the current environment_ selects the whole of the current environment (content + begin / end statements)

Because the current editor may be dirty, it is not easy to cache the AST computed by `parserLatex`. I am still wondering if there is something to be really gained from trying to reuse the AST computed in `buildLaTeXSectionFromFile` for the structure view.

https://github.com/James-Yu/LaTeX-Workshop/blob/f1161d1c0e5f27d258073cfe6f37043ea28007c1/src/providers/structurelib/latex.ts#L86 

Things yet to be done
- [x] Add some tests
- [x] Handle pairs like `\if`/ `\else`, `\if` / `\fi` and `\else` / `\fi` to allow cycling between `\if`, `\else` and `\fi`.


jlelong:
@James-Yu As you redesigned the tests, may I ask there is an easy way of running a single `.test.ts` file in VS Code debug mode without locally modifying the glob statement in`test/suites/index.ts`?

James-Yu:
@jlelong in the launch.json file, there is a LATEXWORKSHOP_SUITE constant. You can insert the test suite file name (or part of it e.g. 04) to just run tests in the file. For a single test, you may change the test.run function to test.only

jlelong:
Thanks a lot for your explanation. I am just realising that this was explained in the test README. RTFM. My bad!

James-Yu:
Will look into this and the issue backlog in a few days. Extremely busy in the past weeks irl 😫

James-Yu:
The PR looks good to me. (p.s. I'm back!)

James-Yu (approved):

PR #3786, opened 2023-03-24
Add config `latex-workshop.latex.jobname`

### LINKED ISSUE #3769: View the pdf according to the "jobname" and "output" of latexmk
> _Please fill the following fields with a star (*) and provide as much related information as possible._

- If we specify the `-jobname` and `-outdir` options in latexmk, the generated PDF should be `[outdir]/[jobname].pdf`.
- When we call the viewer,  **it should open the above PDF, not the one with same name as tex file (e.g., `main.pdf` for `main.tex`)**. If `latex-workshop.latex.outDir` has been changed, the extension still searches for `main.pdf` in this directory.

- Enable `[outdir]/[jobname].pdf` to override `latex-workshop.latex.outDir`, **if either of them is specified**.

- We use `settings.json` to set the two options, like:
```json
        {
            "name": "xelatexmk",
            "command": "latexmk",
            "args": [
                "-xelatex",
                "%DOC%",
                "-jobname=[jobname]",
                "-outdir=[outdir]",
                "-auxdir=[temp file dir]",
                "-synctex=1",
                "-interaction=nonstopmode",
                "-file-line-error"
            ],
            "env": {}
        }
```

James-Yu:
This PR resolves #3769 by adding a new config item `latex-workshop.latex.jobname` to let the extension follow the `-jobname` argument passed to tex.

Similar to `latex-workshop.latex.outDir`, `latex-workshop.latex.jobname` also need to be manually defined besides including the `-jobname` argument in tools.

### DIFF
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -197,7 +197,10 @@ export class Manager {
      * @param respectOutDir If `true`, the 'latex.outDir' config is respected.
      */
     tex2pdf(texPath: string) {
-        return path.resolve(path.dirname(texPath), this.getOutDir(texPath), path.basename(`${texPath.substring(0, texPath.lastIndexOf('.'))}.pdf`))
+        const config = vscode.workspace.getConfiguration('latex-workshop')
+        const jobname = config.get('latex.jobname') as string
+        const texname = texPath.substring(0, texPath.lastIndexOf('.'))
+        return path.resolve(path.dirname(texPath), this.getOutDir(texPath), path.basename(`${jobname || texname}.pdf`))
     }
 
     /**
--- a/src/components/manager.ts
+++ b/src/components/manager.ts
@@ -191,16 +191,24 @@ export class Manager {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/components/manager.ts
```diff
@@ -190,14 +190,25 @@ export class Manager {
         }
     }
 
+    /**
+     * Returns the jobname. If empty, return the name of the input `texPath`.
+     *
+     * @param texPath The path of a LaTeX file.
+     */
+    jobname(texPath: string) {
+        const config = vscode.workspace.getConfiguration('latex-workshop')
```
jlelong, line 199: To access a configuration setting with scope `resource`, we need to pass a `ConfigurationScope` to `getConfiguration` like

```ts
vscode.workspace.getConfiguration('latex-workshop', vscode.Uri.file(texPath))
```

jlelong (changes requested):

src/components/manager.ts
James-Yu: Done!

jlelong (approved):

PR #3788, opened 2023-03-24
Fix scope mismatch between declaration and access

jlelong:
This PR tries to fix scope mismatch between some settings declarations and how they are accessed in the extension.

Any setting with scope `resource` must be accessed with `getConfiguration(section, scope)`, whereas `scope` is either a workspace folder or a file Uri. I think that settings used to initialise intellisense data cannot be tied to a file otherwise we would need to reset default data every time we change the active file (or at least when the root file changes).


James-Yu (approved):

James-Yu:
Thanks for the fix! I always forget about the scope things...

jlelong:
So do I! But looking at your PR earlier today reminded me of this!

PR #3794, opened 2023-03-25
No need to call findRoot on did open text document

### LINKED ISSUE #3792: Undefined `activeTextEditor` in `findRootFromActive` when called from `onDidOpenTextDocument`
Consider a directory with 2 or more possible main tex files, say `main1.tex` and `main2.tex`. 

1. Open the directory in VS Code and the two main files
2. Reload VS Code
3. Put a break point at https://github.com/James-Yu/LaTeX-Workshop/blob/b49efcb4ea355a14ee44b97accc6629f7ec1f4fb/src/components/manager.ts#L369
4. Observe that `vscode.window.activeTextEditor` is undefined

This breaks discovering the correct root File.

When opening a new file, VS Code logic is first to fire a `onDidOpenTextDocument` event and then a `onDidChangeActiveTextEditor` event once the content is loaded into the editor. Nothing happens inside `onDidChangeActiveTextEditor` because `prevTeXDocumentPath` is already to what should be the root file.

I am wondering if there is any reason to call `findRoot` inside `onDidOpenTextDocument`.

jlelong:
Close #3792.

On opening a new document into vscode, two events are emitted sequentially
1. onDidOpenTextDocument: inside the registered callbacks, the active
editor is undefined.
2. onDidChangeActiveTextEditor: we always call findRoot in there

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -38,17 +38,6 @@ export function activate(extensionContext: vscode.ExtensionContext) {
 
     /** The previous active TeX document path. If this changed, root need to be re-searched */
     let prevTeXDocumentPath: string | undefined
-    // This function will be called when a new text is opened, or an inactive editor is reactivated after vscode reload
-    lw.registerDisposable(vscode.workspace.onDidOpenTextDocument(async (e: vscode.TextDocument) => {
-        if (lw.lwfs.isVirtualUri(e.uri)){
-            return
-        }
-        if (lw.manager.hasTexId(e.languageId) && e.fileName !== prevTeXDocumentPath) {
-            prevTeXDocumentPath = e.fileName
-            await lw.manager.findRoot()
-        }
-    }))
-
     let isLaTeXActive = false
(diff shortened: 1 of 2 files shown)


James-Yu (approved):

jlelong:
This PR seems to break https://github.com/James-Yu/LaTeX-Workshop/blob/26b733195103cac96dcb0c291530c4974e5989a0/test/suites/99_multiroot.test.ts#L190

I do not really understand what happens there and see that this test only runs in Linux. @James-Yu any idea?

James-Yu:
> I do not really understand what happens there. I see that this test only is only active under Linux, what is the reason? 

This test was designed to test whether the structure provider can provide the new outline when switching from multiroot A to B. This test is highly problematic due to some vscode problems that are out of reach from the extension side, and produces a large number of false alarms on win32 and darwin. This was the reason why it was only enabled on linux. I do think we can safely remove this test as the structure depends on root detection, which was tested in earlier tests in `multiroot`.

jlelong:
Thanks for your explanations @James-Yu. Then, I am commenting out this test for the time being.

PR #3811, opened 2023-03-31
\include suggestion does not have file extension

### LINKED ISSUE #3808: intellisense file completion in \include not according to standard
This one is simple, the intellisense autocomplete for files inside the \include{} command always shows the file name including the extension.
image
Once accepted, the command is
`\include{introduction.tex}`
But this does not compile on any system, the` .tex` extension should be ommitted as only
`\include{introduction}`
is valid.

James-Yu:
Resolves #3808 

This PR introduces a patch to remove the file extensions from `\include{}` file suggestions. The label of suggestions include the extension to distinguish multiple files with the same name, but the inserted text excludes the extension.

A new patch on test is included to test this feature.

### DIFF
--- a/src/providers/completer/input.ts
+++ b/src/providers/completer/input.ts
@@ -129,7 +129,7 @@ abstract class InputAbstract implements IProvider {
                         }
                         item.range = range
                         item.detail = dir
-                        if (['include'].includes(command)) {
+                        if (['include', 'includeonly', 'excludeonly'].includes(command)) {
                             item.insertText = path.parse(file).name
                         }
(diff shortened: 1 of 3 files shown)


jlelong (commented):
I think the following commands also only expects the base name without extension: `includeonly`, `excludeonly`.

The commands `import|includefrom|inputfrom`, `(sub(?:import|includefrom|inputfrom))` seem to accept the file name with and without extension. So, we can keep the current behaviour completion for them.

jlelong (approved):

Issue #3796, opened 2023-03-26
code-server not loading PDFs from built-in PDF viewer


raulbalanza:
## Pre-checks*
- [x] The issue has not been reported in this repository.
- [x] The issue remains after disabling all other extensions and restarting Visual Studio Code.
- [x] The [FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) cannot address the issue.
- [x] The issue is not related to compiling a document, or the document can be successfully compiled in the OS terminal but not in Visual Studio Code with this extension.

## Environment*
- Operating System: Ubuntu Server (through Docker)
- Visual Studio Code Version: 1.76.1
- LaTeX Workshop Version: 9.7.0
- TeX Distribution Version: TeX Live 2022
- I am using code-server whose version is v4.11.0 inside a Docker container.

## The Issue*
I cannot view PDF files in the built-in PDF viewer of LaTeX Workshop. However, I can do so by using another extensions (e.g., Papyrus PDF Preview).

### Reproduction Steps
1. Deploy code-server (I used this Docker image: https://hub.docker.com/r/linuxserver/code-server)
2. Install texlive from the command line
3. Install LaTeX Workshop
4. Try to view a PDF file (no other extensions installed)

### Expected Behavior
- I expect to see the loaded PDF file like I do on my local VSCode/LaTeX install

### LaTeX Workshop Output*
```
[18:23:45][Logger] New log placeholder %WS1% registered for /config/workspace .
[18:23:45][Config] Configuration for workspace: file://%WS1% .
[18:23:45][Config] editor.acceptSuggestionOnEnter: "on" .
[18:23:45][Config] latex-workshop.bind.enter.key: true .
[18:23:45][Config] latex-workshop.docker.enabled: false .
[18:23:45][Config] latex-workshop.docker.image.latex: "" .
[18:23:45][Config] latex-workshop.hover.preview.mathjax.extensions: [] .
[18:23:45][Config] latex-workshop.intellisense.package.enabled: true .
[18:23:45][Config] latex-workshop.intellisense.update.aggressive.enabled: false .
[18:23:45][Config] latex-workshop.intellisense.update.delay: 1000 .
[18:23:45][Config] latex-workshop.latex.autoBuild.run: "onFileChange" .
[18:23:45][Config] latex-workshop.latex.build.forceRecipeUsage: true .
[18:23:45][Config] latex-workshop.latex.outDir: "%DIR%" .
[...]

PR #3819, opened 2023-04-03
viewer.js updated deprecated function and file pathing

Pascal-D:
This PR resolves the closed Issue #3796.

The use of "../" caused issues with URLs, resulting in the path to the file changing from "localhost/folder/file" to "localhost/file."
The creation of the custom event 'webviewerloaded' utilized a deprecated function, which caused issues with code-server and could lead to problems in the future with the vscode webview and browsers.


Clashsoft (approved):

James-Yu:
Thanks for the contribution. Unfortunately, we typically do not introduce changes directly to `viewer.js`, as it will be overwritten at the next `pdf.js` update. Changes need to find their place in `latexworkshop.ts`, or make themselves as minimal as possible to fit in `dev/editviewer.py`.

Pascal-D:
how does dev/editviewer.py work ? i checked the documentation. Maybe I did not see the part regarding this.

James-Yu:
The script is used to patch (replace) strings in pdf.js files. The code is straightforward, and there is no documentation. It's simply a python script of string replacements.

PR #3832, opened 2023-04-11
Change caption in BTA to top, add a BTB snippet

### LINKED ISSUE #3831: Move caption in the Table snippet to the top
The current table snippet (`BTA`) placed the Table's caption below the Table. It is uncommon in practice (that position is for the figure). The common position for the caption position of the Table is above the Table.

## Pre-checks*
- [ x] The feature request has not been suggested in this repository.

## Current Behavior
```
\begin{table}[htbp]
  \centering\begin{tabular}{<columns>}
    
  \end{tabular}
  \caption{<caption>}
  \label{<label>}
\end{table}
```

```
\begin{table}[htbp]
  \centering
  \caption{<caption>}
  \begin{tabular}{<columns>}
    
  \end{tabular}
  \label{<label>}
\end{table}
```

Thanks for this great plugin. This feature will prevent us from moving the caption manually (cut and paste) from the produced Table by `BTA` snippet.

James-Yu:
This PR fixes #3831.

This PR changed the position of table caption in snippet `BTA` to the beginning of `table` environment, before `tabular` environment.

The original `BTA` layout is preserved in a new snippet `BTB`, in which the second `B` stands for ta**b**le and also **b**ottom.

### DIFF
--- a/data/latex-snippet.json
+++ b/data/latex-snippet.json
@@ -91,6 +91,11 @@
 	},
 	"table": {
 		"prefix": "BTA",
+		"body": "\\begin{table}[${1:htbp}]\n\t\\centering\t\\caption{${2:<caption>}}\n\t\\label{${3:<label>}}\n\\begin{tabular}{${4:<columns>}}\n\t\t${0:${TM_SELECTED_TEXT}}\n\t\\end{tabular}\n\\end{table}",
+		"description": "table"
+	},
+	"table and bottom caption": {
+		"prefix": "BTB",
 		"body": "\\begin{table}[${1:htbp}]\n\t\\centering\\begin{tabular}{${4:<columns>}}\n\t\t${0:${TM_SELECTED_TEXT}}\n\t\\end{tabular}\n\t\\caption{${2:<caption>}}\n\t\\label{${3:<label>}}\n\\end{table}",
 		"description": "table"
 	},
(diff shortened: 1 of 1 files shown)


jlelong (commented):
Looking closely at the table environment, I do not think we should change `TBA`. Tables with captions below the table are very common. This snippet has been around for at least 2 years and nobody has ever complained with the position of the caption.

I would rather recommend to add a new snippet for putting the caption above the table. 
Users can also remove the snippet provided by the extension and define its own.

James-Yu:
> Looking closely at the table environment, I do not think we should change `TBA`. Tables with captions below the table are very common. This snippet has been around for at least 2 years and nobody has ever complained with the position of the caption.
> 
> I would rather recommend to add a new snippet for putting the caption above the table. Users can also remove the snippet provided by the extension and define its own.

Fair enough, will make the change.

jlelong (commented):
Maybe we could make the descriptions of `BTA` and `BTT` more explicit. What about something like _table with bottom caption_ and _caption with top caption_?

James-Yu:
Changed to `table (caption after tabular)` and `table (caption before tabular)` respectively. Shall the latter abbreviation renamed to `BTB` for **b**efore 😆 ?

jlelong:
You are right, `BTB` is more consistent.

PR #3841, opened 2023-04-13
Viewer.js edit with editviewer.py

Pascal-D:
Like in #3819 discussed i made changes to the editviewer.py script. in the else clause is the old code 

Plus
- .replace('''pdfjsLib = require("../build/pdf.js");''','''pdfjsLib = require("./build/pdf.js");''')
- .replace('''value: "../build/pdf.worker.js",''', '''  value: "./build/pdf.worker.js",''')

Minus
- .replace('''console.error(`webviewerloaded: ''', '''// console.error(`webviewerloaded: ''')

the if clause edits only the function webViewerLoad() function


James-Yu:
I’m fine with the path changes. However, it is unclear on why a new event is included. You may need to elaborate more on that or rather, use a more simple and straightforward way to address the issue for maintainability.

James-Yu:
Understood. However, I don't think the deprecated function fix should appear in LaTeX-Workshop. Rather, a PR or fix should appear in https://github.com/mozilla/pdf.js/ . The current edits require monitoring whether and when the upstream pdf.js would fix the issue, which can be a bit tedious to my maintenance.

Will the fix still work without changing the event creation? I myself is a bit against including the corresponding edits.

Pascal-D:
I can understand why you are against modifying the event. If we only adjust the pathing, at least the display of the pdf in codeserver works. I will remove the part where I adjust the function, then you can merge without worry.

James-Yu:
One final thing. This patch should be applied to `viewer.js` and committed, so that the fix will be live in the next release.

Pascal-D:
I have looked further into the problem today and I think I have found actual cause of the problem why the WebView got stuck at the promisses. The function does not need to be changed. I would look into the problem more, so as not to annoy you unnecessarily. Would you be against such a change ?

````
  try {
    parent.document.dispatchEvent(event);
  } catch (ex) {
    // console.error(`webviewerloaded: ${ex}`);
    document.dispatchEvent(event);
  }
````
to

````
  try {
    document.dispatchEvent(event);
    parent.document.dispatchEvent(event);
  } catch (ex) {
    // console.error(`webviewerloaded: ${ex}`);
  }
````

**Comment by James-Yu** [comment]
No. I agree that adding one line to an existing try-catch is minimal enough to fit in the `editviewer` file. You may search for `parent.document.dispatchEvent(event);` and replace it by `parent.document.dispatchEvent(event);\ndocument.dispatchEvent(event);`.

**Comment by Pascal-D** [comment]
I have added it now. If you want you can merge the branch now.

File: dev/editviewer.py [hunk 1]
```diff
@@ -32,7 +32,10 @@
                     .replace('''(!event.shiftKey || window.chrome || window.opera)) {''', '''(!event.shiftKey || window.chrome || window.opera)) {\n    if (window.parent !== window) {\n      return;\n    }''')
                     .replace('''console.error(`webviewerloaded: ''', '''// console.error(`webviewerloaded: ''')
                     .replace('''console.log(`PDF ${pdfDocument.''', '''// console.log(`PDF ${pdfDocument.''')
-            )
+                    .replace('''pdfjsLib = require("../build/pdf.js");''','''pdfjsLib = require("./build/pdf.js");''')
+                    .replace('''value: "../build/pdf.worker.js",''', '''  value: "./build/pdf.worker.js",''')
```
James-Yu, line 36: Extra leading spaces here.

viewer/viewer.js
```diff
@@ -753,7 +753,7 @@ const defaultOptions = {
     kind: OptionKind.WORKER
   },
   workerSrc: {
-    value: "../build/pdf.worker.js",
+      value: "./build/pdf.worker.js",
```
James-Yu, line 756: Same here.

Pascal-D:
Thanks for the codereview. I removed these hickups

PR #3850, opened 2023-04-21
More cleaning methods

### LINKED ISSUE #3849: Delete only the auxiliary files associated with the focused file
## Pre-checks*

## The Missed*

Related: https://github.com/James-Yu/LaTeX-Workshop/issues/3848

Currently, `latex-workshop.clean`, when `"glob"` (`latex-workshop.latex.clean.method`), deletes all auxiliary files in the directory containing the focused file, similar to `latexmk -c`. This also deletes any ancillary files associated with files other than the one in focus.

Consider making this deletion method more restrictive and providing a way to delete only the ancillary files associated with the focused file.

Even with latexmk, `latexmk -c file.tex` deletes only the auxiliary files related to file.tex.

## The Solution*

Two solutions are possible:

1. In addition to the current method of deleting all auxiliary files, provide a method of deleting only those auxiliary files that are relevant to the file in focus.
1. Switching auxiliary files to be deleted by adding new configurations. Like below:
    ```json
      "latex.workshop.latex.clean.target": ""
    ```
    | value       | description                                                      | latexmk example       |
    | ----------- | ---------------------------------------------------------------- | --------------------- |
    | `"all"`     | Delete all the auxiliary files in directory                      | `latexmk -c`          |
    | `"focused"` | Delete only the auxiliary files associated with the focused file | `latexmk -c file.tex` |

James-Yu:
This PR resolves #3849 

This PR introduces two new cleaning methods to `latex.clean.method`, namely,
- `globActive`: Glob files. The cleaned file will have the same as the current active LaTeX file. If the active one is not LaTeX, this config falls back to `globRoot`.
- `globRoot`: The cleaned file will have the same as the jobname or root file. If the root file is not defined, the cleaner will do nothing.

Two new tests are included to test these two cases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1308,6 +1308,12 @@
           ],
           "markdownDescription": "Files to clean. This property must be an array of strings. File globs such as `*.removeme`, `something?.aux` can be used. Users can also specify glob patterns like `emptyfolder*/` to remove empty folders. Non-empty folders will be ignored. The folder globs must end with a slash and the last path component must not contain the globstar `**`, otherwise the folders will not be removed. The following globs patterns are correct `['abc/', 'abc*/', '**/abc*/', 'abc/**/def/']` but these are not ['**', '**/', 'abc/**', 'abc/**/', 'abc/def**/', 'abc/d**ef/']`."
         },
+        "latex-workshop.latex.clean.jobname": {
+          "scope": "resource",
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "Defines whether the globed files by `#latex-workshop.latex.clean.fileTypes#` should have the same file name as the current job name to be cleaned."
+        },
         "latex-workshop.latex.clean.command": {
           "scope": "resource",
           "type": "string",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1332,10 +1332,14 @@
...
           "enum": [
             "glob",
+            "globActive",
+            "globRoot",
             "command"
           ],
           "markdownEnumDescriptions": [
             "Clean all the files located in `#latex-workshop.latex.outDir#` and matching the glob patterns listed in `#latex-workshop.latex.clean.fileTypes#`.",
+            "Clean all the files located in `#latex-workshop.latex.outDir#` and matching the glob patterns listed in `#latex-workshop.latex.clean.fileTypes#`. The cleaned file will have the same as the current active LaTeX file. If the active one is not LaTeX, this config falls back to `globRoot`.",
+            "Clean all the files located in `#latex-workshop.latex.outDir#` and matching the glob patterns listed in `#latex-workshop.latex.clean.fileTypes#`. The cleaned file will have the same as the jobname or root file. If the root file is not defined, the cleaner will do nothing.",
```
jlelong, line 1342: I guess the word _name_ or _base name_ is missing in 

- The cleaned file will have the same as the current active LaTeX file
- The cleaned file will have the same as the jobname or root file

jlelong (commented):
I am wondering about the use cases of `globActive`.

I think `globRoot` should check if the base names of the files to be deleted match files included in the LaTeX project. This would be a behaviour similar to `latexmk -c rootFile`.

I am wondering if the default configuration should not be `globRoot` in the sense defined above. Also what about changing `clean.args` to include `%TEX%` such that the default behaviour is to not remove any file no related to the project? I think this more conservative approach would be safer.

James-Yu:
> I am wondering if the default configuration should not be `globRoot` in the sense defined above. Also what about changing `clean.args` to include `%TEX%` such that the default behaviour is to not remove any file no related to the project? I think this more conservative approach would be safer.

I agree that maybe include LW placeholders in the glob is good enough to support the action, instead of over-complicating the cleaning method. I will create a new PR on this.

PR #3855, opened 2023-04-22
`latex.clean.fileTypes` support placeholders

### LINKED ISSUE #3849: Delete only the auxiliary files associated with the focused file

James-Yu:
This is a continuation of #3850 and resolves #3849.

The `latex.clean.fileTypes` config item now supports LaTeX Workshop placeholders for better customization.

Also, in order to align the effect of default `glob` and `command` cleaning, the default values of `latex.clean.fileTypes` use `%DOCFILE%` now instead of the previous wildcard `*`.


jlelong:
I do not think this is as simple as using `%DOCFILE%` instead of the previous wildcard `*`. Consider a multi file project : `main.tex` and `in.tex`. Calling `latexmk -c main` will clean all ancillary files related to the project such as `main.aux` and `in.aux`. Using `%DOCFILE%.aux` will not clean `in.aux`.

I think it would be good to add `%DOCFILE%` to `latex.clean.command.args`.

James-Yu:
> I do not think this is as simple as using `%DOCFILE%` instead of the previous wildcard `*`. Consider a multi file project : `main.tex` and `in.tex`. Calling `latexmk -c main` will clean all ancillary files related to the project such as `main.aux` and `in.aux`. Using `%DOCFILE%.aux` will not clean `in.aux`.

That's the issue. I intended to make `glob` logic simple, because of the name `glob`. It is possible to add project file "tracing" in cleaning, but the implementation will be yet-another-latexmk. What do you think? (We may revert the default config change back to `*`, just introduce placeholder availability.)

> I think it would be good to add `%DOCFILE%` to `latex.clean.command.args`.

Currently it is `-c` and `%TEX%`, where the latter is a cleaner specific placeholder for root or subfile root. Do you mean that we replace the TEX with DOCFILE?

jlelong:
I did not pay enough attention to the actual configuration of `latex.clean.command.args`. I think the current configuration is fine.

Concerning the glob approach, accepting placeholders cannot harm. I agree with that reimplementing `latexmk -c rootFile` inside LW is not a good idea. As the default recipe already relies on `latexmk`, I am wondering if we could not set the default cleaning approach to `command` instead of `glob`. 

Using `%DOCFILE%` instead of `*` in the glob approach is more conservative which may be good as far as removing files is concerned. There does not seem to be a good and simple solution here. I leave you the final call.

jlelong (approved):

Issue #3804, opened 2023-03-27
Root file detection is based on `\documentclass`


jlelong:
Before version `9.8.0`, the root file detection was looking for `\begin{document}`. Since `9.8.0`, it has been relying on the presence of `\documentclass`. 

See https://github.com/James-Yu/LaTeX-Workshop/wiki/Compile#multi-file-projects

jlelong:
This issue only aims at announcing the change and is not the place for discussing the why. 
If you want to do so, please open a separate discussion or contribute to #3614. Juste for the record, relying on `\documentclass` to detect the main file is what overleaf does. 

I will think about how to add a manual setting to bypass the root file detection.

PR #3856, opened 2023-04-23
Add a configuration item to choose the root file indicator.

shinyypig:
User can choose the root file indicator between `\documentclass` and `\begin{document}`.

Related to https://github.com/James-Yu/LaTeX-Workshop/issues/3804.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,2606 +1,2606 @@
 {
-  "name": "latex-workshop",
-  "displayName": "LaTeX Workshop",
-  "description": "Boost LaTeX typesetting efficiency with preview, compile, autocomplete, colorize, and more.",
-  "icon": "icons/icon.png",
-  "version": "9.8.2",
-  "publisher": "James-Yu",
-  "license": "MIT",
-  "homepage": "https://github.com/James-Yu/LaTeX-Workshop",
-  "repository": {
-    "type": "git",
-    "url": "https://github.com/James-Yu/LaTeX-Workshop.git"
-  },
-  "engines": {
-    "vscode": "^1.74.0"
-  },
-  "categories": [
-    "Programming Languages",
-    "Snippets",
-    "Linters",
-    "Formatters"
-  ],
-  "keywords": [
-    "latex",
-    "tex",
-    "compile",
-    "preview",
-    "hint"
-  ],
-  "activationEvents": [
-    "onCommand:latex-workshop.activate",
-    "onWebviewPanel:latex-workshop-pdf"
-  ],
-  "main": "./out/src/main.js",
-  "capabilities": {
-    "virtualWorkspaces": {
-      "supported": "limited",
-      "description": "Only a few features are supported."
+    "name": "latex-workshop",
+    "displayName": "LaTeX Workshop",
+    "description": "Boost LaTeX typesetting efficiency with preview, compile, autocomplete, colorize, and more.",
+    "icon": "icons/icon.png",
+    "version": "9.8.2",
+    "publisher": "James-Yu",
+    "license": "MIT",
+    "homepage": "https://github.com/James-Yu/LaTeX-Workshop",
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/James-Yu/LaTeX-Workshop.git"
     },
-    "untrustedWorkspaces": {
-      "supported": false
-    }
-  },
-  "contributes": {
-    "languages": [
-      {
-        "id": "tex",
-        "aliases": [
-          "TeX",
-          "tex"
-        ],
-        "extensions": [
-          ".sty",
-          ".cls",
-          ".bbx",
-          ".cbx"
-        ],
-        "configuration": "./syntax/latex-language-configuration.json"
-      },
-      {
-        "id": "doctex",
-        "aliases": [
-          "DocTeX",
-          "doctex"
-        ],
-        "extensions": [
-          ".dtx"
-        ],
-        "configuration": "./syntax/doctex-language-configuration.json"
-      },
-      {
-        "id": "latex",
-        "aliases": [
-          "LaTeX",
-          "latex"
-        ],
-        "extensions": [
-          ".tex",
-          ".ltx",
-          ".ctx"
-        ],
-        "configuration": "./syntax/latex-language-configuration.json"
-      },
-      {
-        "id": "bibtex",
-        "aliases": [
-          "BibTeX",
-          "bibtex"
-        ],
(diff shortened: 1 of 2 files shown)


James-Yu:
This PR needs discussion on its necessity before looking into the details. I don't think there is a use case where detecting `documentclass` won't work. The referenced issue need only to move `documentclass` to the said `main` file.

shinyypig:
I believe that many users, including myself, use a "settings.tex" file to save the document format and packages used when writing multiple LaTeX projects. If we could use "\begin{document}" to detect the root file, then when we need to modify the style of these LaTeX projects, we only need to synchronize all "settings.tex" files. On the other hand, if we use "\documentclass" to detect the root file, we not only need to synchronize "settings.tex", but also need to modify the "\documentclass" of each LaTeX project.

James-Yu:
I personally am not convinced by the use case. May I have your thoughts? @jlelong

shinyypig:
This is a commonly used means of synchronizing format when writing papers with independent appendices and supplementary files.

mihaiconstantin:
> I believe that many users, including myself, use a "settings.tex" file to save the document format and packages used when writing multiple LaTeX projects. If we could use "\begin{document}" to detect the root file, then when we need to modify the style of these LaTeX projects, we only need to synchronize all "settings.tex" files. On the other hand, if we use "\documentclass" to detect the root file, we not only need to synchronize "settings.tex", but also need to modify the "\documentclass" of each LaTeX project.

@shinyypig, I very much agree with this. Literally all my workflows follow the setup you indicated.

> I personally am not convinced by the use case. May I have your thoughts? @jlelong

@James-Yu, I am personally disappointed to read your comment. At the very least, please allow users to choose rather than forcing them into a use case.

mihaiconstantin:
> I believe that many users, including myself, use a "settings.tex" file to save the document format and packages used when writing multiple LaTeX projects. If we could use "\begin{document}" to detect the root file, then when we need to modify the style of these LaTeX projects, we only need to synchronize all "settings.tex" files. On the other hand, if we use "\documentclass" to detect the root file, we not only need to synchronize "settings.tex", but also need to modify the "\documentclass" of each LaTeX project.

@shinyypig, I very much agree with this. Literally all my workflows follow the setup you indicated.

> I personally am not convinced by the use case. May I have your thoughts? @jlelong

@James-Yu, Since this change is likely to impact users' workflows, I think it would be greatly appreciated if you allowed users to choose rather than being force into a use case. What are your concerns regarding the ability to choose the root file?

jlelong:
I have looked at various LaTeX editors and it clearly appears that setting the root file remains really challenging. When the editor tries to discover the root file by some internal mechanism, relying on the presence of `documentclass` seems much more common. This is how Overleaf proceeds for instance.

Even though I am not fully convinced by the use case,  I can understand that some projects are organised in a way that detecting `begin{document}` may be simpler. I have the impression that letting the user decide which string to detect does not harm.

package.json
```diff
@@ -1151,6 +1151,16 @@
...
         },
+        "latex-workshop.latex.rootFile.indicator": {
+            "scope": "resource",
+            "type": "string",
+            "default": "\\documentclass",
+            "enum": [
+                "\\documentclass",
+                "\\begin{document}"
+            ],
+            "markdownDescription": "Determine the root file based on the specific indicator. If `\\documentclass`, the root file is the file containing `\\documentclass`. If `\\begin{document}`, the root file is the file containing `\\begin{document}`."
```
James-Yu, line 1162: The description can be more concise, e.g., Determine if the root file is detected based on the presence of `\\documentclass` or `\\begin{document}`.

James-Yu: Also, the detection is not just `\\documentclass`, but `\\documentclass[]{}`.

src/components/manager.ts
```diff
@@ -87,6 +85,27 @@ export class Manager {
...
+        const configuration = vscode.workspace.getConfiguration('latex-workshop')
+        const indicator = configuration.get('latex.rootFile.indicator')
+        let ROOT_INDICATOR = null
+        switch (indicator) {
+            case '\\documentclass':
+                ROOT_INDICATOR = /\\documentclass(?:\s*\[.*\])?\s*\{.*\}/ms
+                break
+            case '\\begin{document}':
+                ROOT_INDICATOR = /\\begin{document}/m
+                break
```
James-Yu, line 101: Do not use a CONSTANT variable in function. For this implementation, returning the corresponding regex is sufficient.

src/components/manager.ts
```diff
@@ -87,6 +85,27 @@ export class Manager {
...
+            default:
+                ROOT_INDICATOR = /\\documentclass(?:\s*\[.*\])?\s*\{.*\}/ms
+                break
```
James-Yu, line 104: The default case should also log as an error.

src/components/manager.ts
```diff
@@ -87,6 +85,27 @@ export class Manager {
         return path.normalize(out).split(path.sep).join('/')
     }
 
+    /**
+     * @returns the indicator if the root file
+     */
+    getRootIndictor() {
```
James-Yu, line 91: Typo. Also make this a proper [Typescript getter](https://www.typescripttutorial.net/typescript-tutorial/typescript-getters-setters/).

James-Yu (changes requested):

src/components/manager.ts
shinyypig: image

What should I use? Const or let?

src/components/manager.ts
```diff
@@ -87,6 +85,23 @@ export class Manager {
...
+     * The indicator of the root file.
+     */
+    get rootIndictor() {
+        const configuration = vscode.workspace.getConfiguration('latex-workshop')
+        const indicator = configuration.get('latex.rootFile.indicator')
+        switch (indicator) {
+            case '\\documentclass[]{}':
+                return /\\documentclass(?:\s*\[.*\])?\s*\{.*\}/ms
+            case '\\begin{document}':
+                return /\\begin{document}/m
```
James-Yu, line 98: It's possible to have spaces or newlines before the left brace `{`. This case was neglected in previous versions. Please add this.

shinyypig: Sorry, I am not familiar with regex. These are directly copied from the source code. I don't know how to modify them. Could you please provide  me the required regex?

James-Yu: You may try /\\begin\s*{document}/m

Issue #3866, opened 2023-05-01
Outline view not working when containing some text in .tex file


epsonlq138:
> _Please fill the following fields with a star (*) and provide as much related information as possible._

## Pre-checks*
> _Please change the following `[ ]` to `[x]` for confirmation._
- [ x] The issue has not been reported in this repository.
- [ x] The issue remains after disabling all other extensions and restarting Visual Studio Code.
- [ x] The [FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) cannot address the issue.
- [ x] The issue is not related to compiling a document, or the document can be successfully compiled in the OS terminal but not in Visual Studio Code with this extension.

## Environment*
> _Please write exact version numbers instead of descriptors such as `latest`._
- Operating System: Windows 10
- Visual Studio Code Version: 1.77.3 (user setup)
- LaTeX Workshop Version: v9.9.0
- TeX Distribution Version: TeX 3.141592653 (TeX Live 2022)
- 
> _Please list the environment and version number if you are using VSCodium, Snap or Flatpack versions of Visual Studio Code, and/or Visual Studio Code Remote Containers/SSH/WSL._
- I am using XXX whose version is YY.ZZ

## The Issue*
> _Please briefly describe the issue you come across._
- Outline view not working when containing some text in .tex file

### Reproduction Steps
> _Please list out the steps to reproduce your bug. Include relevant environmental variables or any other configuration._
0. Disable all extension except LaTex-Workshop.
1. Open test.tex file from a folder in VSCode. 
2. Insert the following latex code and save
```
\begin{document}
\section{123}
$\left[ e^{ j\pi \left( N_t-1 \right) } \right]$
\end{document}
```
3.  Check the outline view shows that "No symbols found in document 'test.tex'".
![image](https://user-images.githubusercontent.com/12795429/235455865-d7a44cf1-da1e-4d4f-956c-ed6c0c02e41c.png)

### Expected Behavior
> _What were you expecting to see? Include any relevant examples or documentation links._
- If replace latex code as following, it will be normal.
```
\begin{document}
\section{123}
$\left[ e^{ j\pi \left( N_t-1 \right) } \right]$
\end{document}
```
- 
![image](https://user-images.githubusercontent.com/12795429/235455943-44ecc470-45e5-4a32-a22f-81d258a130a5.png)

### LaTeX Workshop Output*
[...]

PR #3817, opened 2023-04-03
Let ) may also close [ when stripping text

James-Yu:
This PR partially resolves #3813.

Previously, stripping text before AST considers `[` can be only closed by `]`. This is causing a lot of trouble for range definitions like `[1, 2)`, as the next `]` at a random position is used to close the brace.

This PR makes the `getLongestBalancedString` function take `)` to close `[`. I've tested some use cases and this change indeed creates more authentic stripped tex code.

Nonetheless, the current `utensils` parser is still sometime problematic, especially there are errors in brace pairs. I will look into the issue and work on `parser.ts` to see if there's anything can be done from the extension side.

PR #3867, opened 2023-05-01
Use bracket stack to `getLongestBalancedString`

James-Yu:
This PR is a rewrite of the `getLongestBalancedString` function, aiming at resolving #3866 

In the previous implementation, a universal counter is used to count the nests. This should work as intended if curlies are matched with curlies, and squares match squares.

However, in order to make `\in[1,2)` work, we previously added that `)` can also close `[`. This is causing trouble with `\left[ e^{ j\pi \left( N_t-1 \right) } \right]`, as the `)` is instead of closing the `(`, but traced back to the first `[` and making the stripped string invalid.

In this PR, we use a stack to hold the sequence of brackets `{ [ (` in the context, and match only those that can match:
- `}` closes the previous `{`, no matter what's in between.
- `]` closes the previous `[`, no matter what's in between.
- `)` closes the `(` or `[` if these two are immediately preceeding the `)`, otherwise this `)` is discarded.

It works well.

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -121,39 +121,34 @@ export function trimMultiLineString(text: string): string {
  * @param s A string to be searched.
  */
 function getLongestBalancedString(s: string, bracket: 'curly' | 'square'='curly'): string | undefined {
-    const openner = bracket === 'curly' ? '{' : '['
-    const closer = bracket === 'curly' ? '}' : ']'
-    let nested = s[0] === openner ? 0 : 1
-    let i = 0
-    for (i = 0; i < s.length; i++) {
-        switch (s[i]) {
-            case openner:
-                nested++
-                break
-            case closer:
-                nested--
-                break
-            case '\\':
-                // skip an escaped character
-                i++
-                break
-            case ')':
-                // [1, 2)
-                if (openner === '[') {
-                    nested--
-                }
-                break
-            default:
+    const bracketStack: ('{' | '[' | '(')[] = []
+
+    const opener = bracket === 'curly' ? '{' : '['
+    if (s[0] !== opener) {
+        bracketStack.push(opener)
+    }
+
+    for (let i = 0; i < s.length; ++i) {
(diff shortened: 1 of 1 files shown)


src/utils/utils.ts
```diff
@@ -121,39 +121,34 @@ export function trimMultiLineString(text: string): string {
...
+
+    for (let i = 0; i < s.length; ++i) {
+        const char = s[i]
+        if (char === '{' || char === '[' || char === '(') {
+            bracketStack.push(char)
+        } else if (char === '}' || char === ']') {
+            const openPos = bracketStack.lastIndexOf(char === '}' ? '{' : '[')
+            if (openPos > -1) {
+                bracketStack.splice(openPos, 1)
+            }
```
jlelong, line 139: Maybe, splitting the cases `}` and `]` will make maintenance and understanding easier.

jlelong:
I am not sure to properly understand 

> `)` closes the `(` or `[` if these two are immediately preceding the `)`, otherwise this `)` is discarded

More importantly, I am not able to find this condition in the implementation. Am I missing something?

James-Yu:
> I am not sure to properly understand
> 
> > `)` closes the `(` or `[` if these two are immediately preceding the `)`, otherwise this `)` is discarded

The original use case was recorded in #3817. The rule above is an abstraction to the PR:
- `[ )` closes (`\in[1,2)`).
- `( )` closes.
- `[ { )` should not close the first square because the `)` is within the context of `{`.

Hope this may remove some of the mists on this rule.

> More importantly, I am not able to find this condition in the implementation. Am I missing something?

The implementation is in L140 to L145: if the current char is `)`, then we get the last bracket in the stack. If `[` or `(`, pop it (close the bracket), otherwise, ignore the `)`.

src/utils/utils.ts
James-Yu: Here it goes.

jlelong (approved):

Issue #3877, opened 2023-05-10
Is removing dependency to `latex-utensils` a good idea?


James-Yu:
This issue is a dev discussion.

@jlelong In the past few days, I've been thinking on the pros and cons of `latex-utensils` dependency. I am currently with a thought that while LU grants lazy parsing, the limitations are significant enough such that reverting back to an enhanced regex-based parsing might indeed improve the long-term maintanence experience. The incapability when a minor syntax error appeared in editor, the (relatively) long parsing time on a long document, the still-problematic `stripText`, the less maintained and related repo, and the fact that indeed we may not need a full-fledged AST for LaTeX.

I am aware that this can be a quite big change, and a very notable harm to what we have been sticking to in the past year. If you also share a similar thought, I may start on something small, e.g., the structure (as I have been struggling with it for quite some time). The change may rollout very slowly and gradually, so that the flood of bug reports can be attenuated.

What do you think @jlelong ? I am happy to take the tasks if you think it may worth a try.

James-Yu:
Thanks for the input. Indeed, the major drawback for AST-based parsing, afaik:

X. Relatively slow parsing speed, which in fact 1) is not a big deal if we have a perfect `stripText` (somehow hard) or 2) is notable if the document is too literal.
X. Still needs to fallback to regex if the LaTeX document has syntax errors, even if they are not related to a particular feature.

Regex may contribute to both, but the maintainence is indeed very challenging. I totally agree.

Let me take a look at `unified-latex` and see if it is a good in-place substitution. It would be awesome if both problems can be resolved.

Let me also shake off my thought on turning back to regex 🤣

James-Yu:
Tiny test:

- `unified-latex` is 3x slower than `latex-utensils`. However, the excessively long document in #3813 (7400+ lines, 300k+ chars) renders ~400ms and ~166ms for either to parse, which is beyond my expectation. Maybe AST-based parsing is not as bad as I thought in terms of speed.
- `unified-latex` can properly parse LaTeX with syntax errors! It just follows the catcode definition of LaTeX.
- `unified-latex` does not have too many and too detailed types for LaTeX syntax, which to me is indeed quite promising.

It looks like `unified-latex` is a good candidate to switch. However,

- `unified-latex` does not have a `bibtex` parser. At least, yet.

[...]

PR #3879, opened 2023-05-13
`unified-latex`-based LaTeX and DocTeX structuring

James-Yu:
This PR is a follow-up of the discussion in #3877 and a first attempt of migrating to `unified-latex`.

The logic of constructing structure is almost retained, with only proper adaptation to `unified-latex`.

This PR also serves a demo on how `unified-latex` can be integrated into this extension.


James-Yu:
NB: The refactor passes previous structure tests in 991d3f5, before the test refactor in fb1b3ce.

jlelong (commented):
It is not easy to review the PR in detail but it looks good to me. It seems to be a quite straightforward drop-in replacement. 

Il would interesting to investigate if `unified-latex` could also be used fro completion purposes. It would make it a true alternative to LU.

James-Yu:
> It is not easy to review the PR in detail but it looks good to me. It seems to be a quite straightforward drop-in replacement.

I fully agree. I do plan to release this UL-based structure in few days as sort of preview into a fully UL-based LW. This may limit the possible negative impact within the structurer scope.

> Il would interesting to investigate if `unified-latex` could also be used fro completion purposes. It would make it a true alternative to LU.

Indeed, it is quite possible and straightforward to use UL for completion and almost all TeX parsing parts. As to `bibtex`, I do have a plan to implement our own simple parser based on existing ones. It is far more simple than a LaTeX one.

Let me merge this and cross the finger.

James-Yu:
It's just too unfortunate that we are unlikely to integrate `unified-latex` into LW. This package is built upon `unified`, which since `v10` only publish ESM distributions. However, either vscode or electron only supports CommonJS, no ESM for sure.

I have tried 10+ different solutions to circumvent the issue, to no vail. I will keep an eye on https://github.com/prettier/prettier-vscode/issues/3007, which should be facing the same problem. (`prettierjs` is also built on `unified`).

PR #3934, opened 2023-06-16
Unified-latex based cache

James-Yu:
This is a mega-PR aiming at using `unified-latex` to replace `latex-utensils` in Cache, mainly on auto-completions.


James-Yu:
Due to the difficulty of reviewing a big PR like this, I will directly merge after vetting the code. Let the bullet fly.

Issue #3722, opened 2023-02-15
Invert preview colors directly from command palette


kszenes:
> _Please fill the following fields with a star (*) and provide as much related information as possible._

## Pre-checks*
> _Please change the following `[ ]` to `[x]` for confirmation._
- [x ] The feature request has not been suggested in this repository.

## The Missed*
> _Is your feature request related to a problem? Please provide a clear and concise description of what the problem is._
- Currently, to invert the colors in the pdf previewer, the invert filter needs to be modified from the settings pane and the previewer needs to be relaunched. I would like to have a quicker method for toggling the CSS filter and potentially also reopening the previewer.

## The Solution*
> _Please provide a solution you would like to have._
- I would suggest the addition of a command to the command palette that toggles the filter on and off and potentially also reopens the previewer. I am aware that the CSS filter can take values between 0.0 and 1.0 but I think it would be quite useful to have an option to simply toggle it between the two extrema.

## Anything Else?
> _Add any other context about the feature request below._
- If you think this would be a helpful addition, I would be happy to implement this if you give me a few pointers.

Thanks for developing this great extension!

PR #3935, opened 2023-06-16
Add a command for inverting PDF colours.

kamilest:
This adds an option for inverting PDF colours directly from the command palette.

This avoids the need to go to settings every time the colours need to be inverted—which I think was also the request in #3722.

Let me know if this makes sense or if anything else is needed.

https://github.com/James-Yu/LaTeX-Workshop/assets/33470490/567331b9-16ef-4614-aef0-dffba5f22a98

### DIFF
--- a/package.json
+++ b/package.json
@@ -497,6 +497,11 @@
         "command": "latex-workshop.toggleMathPreviewPanel",
         "title": "Toggle Math Preview Panel",
         "category": "LaTeX Workshop"
+      },
+      {
+        "command": "latex-workshop.togglePdfInvertFilter",
+        "title": "Toggle PDF invert filter",
+        "category": "LaTeX Workshop"
       }
     ],
     "keybindings": [
--- a/src/commander.ts
+++ b/src/commander.ts
@@ -493,6 +493,12 @@ export function toggleMathPreviewPanel() {
     lw.mathPreviewPanel.toggle()
 }
 
+export function togglePdfInvertFilter() {
+    const configuration = vscode.workspace.getConfiguration('latex-workshop')
+    const invert = configuration.get('view.pdf.invert') as boolean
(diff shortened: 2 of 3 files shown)


James-Yu:
Thank you for this proposal. However, I do not think this additional command adds more convenience over changing a value in the settings pane. #3722 complaints not on the difficulty of changing the value, but the need of close-reopen viewer after changing the value, which has been addressed. Therefore, this PR won't be merged.

kamilest:
Thank you for the quick response, @James-Yu! 

Would you be able to reconsider? I, for one, think this feature would be very useful, as I use the dark mode for code, but prefer seeing the PDF file in its true colours _unless_ at night—therefore I need to change the value in the settings on a daily basis.

#3722 specifically mentions the "command palette", which is the one that is accessible through `⇧⌘P`—precisely what this PR implements. @kszenes also mentions they "would like a _quicker_ method for _toggling_ the CSS filter" and only "_potentially_" reopening the previewer. 

Regarding the values between `0` and `1`—I would also agree with the user above that toggling between the two extrema could still be useful, though I, too, am aware this does not restore the original value if one previously had some custom setting. If you have any suggestions what would be a good way save the custom value and toggle between 0 and that (including perhaps through using `invertMode.enabled` rather than `invert`, or adding a new setting for toggling options), I'd be happy to help implement this.

James-Yu:
You may consider use the approaches in https://stackoverflow.com/questions/57470525/how-can-i-run-a-vscode-command-as-a-task and or https://stackoverflow.com/questions/73867739/how-to-change-vs-code-settings-from-terminal-bash to work out a local solution.

As the frequency of change in your use case is a few times per day, I won't consider adding a new command is appropriate as the latter incurs much maintainence burden (though now it seems to be trivial).

PR #3942, opened 2023-06-20
Fix misleading checkmark label

japrozs:


James-Yu:
Adding the word will horizontally expand the dropdown menu. Can you please instead change the default status of the input, and toggle its behavior in the script? Thanks.

James-Yu:
Close for inactivity.

PR #3961, opened 2023-07-09
Custom Editor API

ManuelHentschel:
Implements the changes discussed in #3951

I added a new viewmode instead of replacing the "singleton" mode, in case some users want to preserve the specific behavior of that mode, but of course that can be changed easily.

I had to add the option `"latex-workshop.view.pdf.tab.viewColumn"`, since the current option `"latex-workshop.view.pdf.tab.editorGroup"` is not compatible with the editor-open API of vscode, and merging the two felt rather confusing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1454,13 +1454,10 @@
         },
         "latex-workshop.view.pdf.tab.openDelay": {
           "scope": "window",
-          "type": "object",
-          "additionalProperties": {
-            "type": "string"
-          },
-          "default": {},
-          "deprecationMessage": "Deprecated: This config has been removed.",
-          "markdownDeprecationMessage": "**Deprecated**: This config has been removed."
+          "type": "number",
+          "default": 1000,
+          "deprecationMessage": "Deprecated: This config has no use and has been removed.",
+          "markdownDeprecationMessage": "**Deprecated**: This config has no use and has been removed."
         },
         "latex-workshop.view.pdf.ref.viewer": {
           "type": "string",
--- a/package.json
+++ b/package.json
@@ -1420,14 +1420,16 @@
             "browser",
             "tab",
             "singleton",
-            "external"
+            "external",
+            "customEditor"
           ],
           "markdownDescription": "The default PDF viewer.",
           "enumDescriptions": [
             "Open PDF with the default web browser.",
             "Open PDF with the built-in tab viewer.",
             "Open PDF with the built-in tab viewer, reveal existing one if possible.",
-            "[Experimental] Open PDF with the external viewer set in \"View > Pdf > External: command\"."
+            "[Experimental] Open PDF with the external viewer set in \"View > Pdf > External: command\".",
+            "Open PDF with the built-in tab viewer, treating it as a custom editor."
           ]
         },
         "latex-workshop.view.pdf.tab.editorGroup": {
@@ -1450,11 +1452,32 @@
             "Put the viewer tab in a new group below the current one"
           ]
         },
+        "latex-workshop.view.pdf.tab.viewColumn": {
+          "scope": "window",
+          "type": "string",
+          "default": "Beside",
+          "markdownDescription": "The view column in which to open the tab viewer, if treating it as a custom editor.",
+          "enum": [
+            "Active",
+            "Beside",
+            "One",
+            "Two",
+            "Three"
+          ],
+          "enumDescriptions": [
+            "Use the currently active editor group",
+            "Use the column next to the current editor",
+            "Use the first editor column",
+            "Use the second editor column",
+            "Use the third editor column"
+          ]
+        },
         "latex-workshop.view.pdf.tab.openDelay": {
           "scope": "window",
           "type": "number",
           "default": 1000,
-          "markdownDescription": "Defines the delay in milliseconds to wait for a tab opening. Please increase the value if you encounter a focus issue after opening a tab."
+          "markdownDescription": "Defines the delay in milliseconds to wait for a tab opening. Please increase the value if you encounter a focus issue after opening a tab.",
+          "markdownDeprecationMessage": "Deprecated, does not have any effect."
         },
         "latex-workshop.view.pdf.ref.viewer": {
           "type": "string",
--- a/package.json
+++ b/package.json
@@ -1429,7 +1429,7 @@
             "Open PDF with the built-in tab viewer.",
             "Open PDF with the built-in tab viewer, reveal existing one if possible.",
             "[Experimental] Open PDF with the external viewer set in \"View > Pdf > External: command\".",
-            "Open PDF with the built-in tab viewer, treating it as a custom editor."
+            "[Experimental] Open PDF with the built-in tab viewer, using the vscode custom editor API."
           ]
         },
         "latex-workshop.view.pdf.tab.editorGroup": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -1420,14 +1420,16 @@
...
+            "customEditor"
           ],
           "markdownDescription": "The default PDF viewer.",
           "enumDescriptions": [
             "Open PDF with the default web browser.",
             "Open PDF with the built-in tab viewer.",
             "Open PDF with the built-in tab viewer, reveal existing one if possible.",
-            "[Experimental] Open PDF with the external viewer set in \"View > Pdf > External: command\"."
+            "[Experimental] Open PDF with the external viewer set in \"View > Pdf > External: command\".",
+            "Open PDF with the built-in tab viewer, treating it as a custom editor."
```
James-Yu, line 1432: General users may not know what is a custom editor. I suggest rephrase it to something like "use vscode custom editor api" or so.

James-Yu: Also, I'd suggest to add an "experimental" flag.

package.json
```diff
@@ -1450,11 +1452,32 @@
...
+            "Three"
+          ],
+          "enumDescriptions": [
+            "Use the currently active editor group",
+            "Use the column next to the current editor",
+            "Use the first editor column",
+            "Use the second editor column",
+            "Use the third editor column"
+          ]
+        },
```
James-Yu, line 1474: At the beginning of LW development, I used a `view.pdf.tab.editorGroup` to control the viewer position. I'm not sure which one is better, but we should be consistent. I'd suggest use `view.pdf.tab.editorGroup` for now, and we may have another discussion on this after merged.

package.json
```diff
@@ -1450,11 +1452,32 @@
...
         "latex-workshop.view.pdf.tab.openDelay": {
           "scope": "window",
           "type": "number",
           "default": 1000,
-          "markdownDescription": "Defines the delay in milliseconds to wait for a tab opening. Please increase the value if you encounter a focus issue after opening a tab."
+          "markdownDescription": "Defines the delay in milliseconds to wait for a tab opening. Please increase the value if you encounter a focus issue after opening a tab.",
+          "markdownDeprecationMessage": "Deprecated, does not have any effect."
```
James-Yu, line 1480: Deprecated configs should be changed to what we did for e.g. `intellisense.atSuggestionJSON.replace`

src/components/viewer.ts
```diff
@@ -16,6 +16,8 @@ import { moveActiveEditor } from '../utils/webview'
 
 const logger = getLogger('Viewer')
 
+export type ViewerLocation = 'tab' | 'browser' | 'singleton' | 'external' | 'customEditor'
```
James-Yu, line 19: `ViewerMode`? Location may be mis-interpreted as left, right or so.

src/components/viewer.ts
```diff
@@ -97,13 +99,19 @@ export class Viewer {
     async open(pdfFile: string, mode?: string): Promise<void> {
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
         const tabEditorGroup = configuration.get('view.pdf.tab.editorGroup') as string
-        const viewer = mode ?? configuration.get<'tab' | 'browser' | 'singleton' | 'external'>('view.pdf.viewer', 'tab')
+        const viewerLocation = configuration.get<ViewerLocation>('view.pdf.viewer', 'tab')
+        if(mode === 'tab' && (viewerLocation === 'customEditor' || viewerLocation === 'singleton')) {
```
James-Yu, line 103: Instead of using two variables here, `mode` and `viewerLocation` should be unified. This only involves changing the function at [here](https://github.com/James-Yu/LaTeX-Workshop/blob/40a065513c41972327a0c5d757253a3df4fb48a0/src/commander.ts#L82)

src/components/viewerlib/pdfviewerhook.ts
```diff
@@ -9,12 +12,23 @@ class PdfViewerHookProvider implements vscode.CustomReadonlyEditorProvider {
...
+        const configuration = vscode.workspace.getConfiguration('latex-workshop')
+        const viewerLocation = configuration.get<ViewerLocation>('view.pdf.viewer', 'tab')
+        if(viewerLocation === 'customEditor'){
+            webviewPanel.webview.options = {
+                ...webviewPanel.webview.options,
+                enableScripts: true
+            }
+            const pdfPanel = await populatePdfViewerPanel(document.uri, webviewPanel)
+            void viewerManager.initiatePdfViewerPanel(pdfPanel)
+        } else{
```
James-Yu, line 25: There are some trivial formatting issues here and there.

src/components/viewerlib/pdfviewerpanel.ts
```diff
@@ -85,16 +85,23 @@ async function patchCodespaces(url: vscode.Uri) {
...
-    const htmlContent = await getPDFViewerContent(pdfUri)
     const panel = vscode.window.createWebviewPanel('latex-workshop-pdf', path.basename(pdfUri.path), {
         viewColumn: vscode.ViewColumn.Active,
         preserveFocus
     }, {
         enableScripts: true,
         retainContextWhenHidden: true
     })
+    const pdfPanel = await populatePdfViewerPanel(pdfUri, panel)
+    return pdfPanel
```
James-Yu, line 98: You can just `return await` here.

James-Yu (changes requested):

package.json
ManuelHentschel: I added the flag and modified the description. Feel free to change the description again, if you have something specific in mind.

package.json
ManuelHentschel: I added a rough "translation" of `editorGroup` to `ViewColum` for now.

James-Yu:
Can you take a look at 8c3df83 and see if things work as you intended?

ManuelHentschel:
> Can you take a look at [8c3df83](https://github.com/James-Yu/LaTeX-Workshop/commit/8c3df835284d49066138a36d6bf660f1e5f2821d) and see if things work as you intended?

With this change, the command `latex-workshop.tab` opens the (non-custom-editor) viewer, even if otherwise the custom editor API is specified in the settings. In line with your earlier [comment](https://github.com/James-Yu/LaTeX-Workshop/discussions/3951#discussioncomment-6335537), I think it would make sense to adjust the behavior of this command to choose the correct internal viewer. This is why I used the `'internal'` option in some places. To avoid that we can also put more logic into [`main.ts`](https://github.com/James-Yu/LaTeX-Workshop/blob/8c3df835284d49066138a36d6bf660f1e5f2821d/src/main.ts#L124) if you prefer that over using `"internal"`.

Also, omitting the default value in `Viewer.open` also causes the command `latex-workshop.view` to stop working.

James-Yu:
I suppose f8c7c3a fixes both?

ManuelHentschel:
> I suppose [f8c7c3a](https://github.com/James-Yu/LaTeX-Workshop/commit/f8c7c3ab274bcc8bf8e613a49373514d12dd68e5) fixes both?

The latest version seems to work as intended

James-Yu:
Before merge, please add related tests. A minimum one is to try setting the config and view and see if a viewer is fired. You may refer to other tests for the implementation. I do encourage to add more as you may think of (I am not quite sufficient in thinking of different test cases.)

ManuelHentschel:
Regarding the recent test run: the only failed test is `"04_intellisense: command intellisense with cmds provided by \usepackage and its argument"` on Mac. I did not modify the intellisense related files and cannot reproduce the failing test locally (I'm on windows and the windows and Linux tests did not fail anyways), so I don't really know what to do here.

In other repos, I sometimes had issues with MacOS test runners that resolved themselves by simply rerunning the test, not sure if that makes a difference here.

James-Yu:
Thanks for your contribution!

James-Yu:
It seems that the custom editors can be duplicated by the "splitting" button, which I think is acceptable. Therefore, I'd propose to switch the options of `view.mode` to
- `tab` uses custom editor
- `singleton` uses old logic (primary unchanged), marked with "deprecated"
- Remove `customEditor`
- Add `legacy`, which uses old logic

What do you think? @ManuelHentschel

ManuelHentschel:
> It seems that the custom editors can be duplicated by the "splitting" button, which I think is acceptable. Therefore, I'd propose to switch the options of `view.mode` to
> 
>     * `tab` uses custom editor
> 
>     * `singleton` uses old logic (primary unchanged), marked with "deprecated"
> 
>     * Remove `customEditor`
> 
>     * Add `legacy`, which uses old logic
> 
> 
> What do you think? @ManuelHentschel

Sounds good to me!

PR #3966, opened 2023-07-13
Add option to include filename in status bar

ManuelHentschel:
This PR adds an option to include the name of the root file in the status bar while compiling/building.

I find this very useful when working in directories with multiple latex files that take a long time to compile. Scenarios are e.g.:
- `fileA.tex` is being built/just got done being built, and I save `fileB.tex`. Now I can see whether `fileA.tex` is still being built or `fileB.tex` is already being built.
- `fileA.tex` is included by two root files, e.g. `examQuestions.tex` and `examSolutions.tex`, and I save it. Now I can see which of the two root files was chosen to be built by the root-finding mechanism.

This does not solve the issues raised e.g. in #3614 and #3172, but helps coping with the "blackbox" root-finding mechanism.

I set the option default to `false` to not add noise to the status bar for users who are not interested in this info, but that can of course be changed.

I am already using this modification on my personal fork, so I directly opened a PR. If you prefer, I can also close this PR and open a discussion topic first.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1074,6 +1074,12 @@
           "default": "",
           "markdownDescription": "The jobname argument of the compiling tool, which is used by the extension to find project files (e.g., PDF and SyncTeX files). This config should be set identical to the value provided to the `-jobname=` argument, and should not have placeholders. Leave the config empty to ignore jobname and keep the default behavior."
         },
+        "latex-workshop.latex.build.rootfileInStatus": {
+          "scope": "resource",
+          "type": "boolean",
+          "default": false,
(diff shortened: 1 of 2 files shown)


James-Yu:
Thanks for the PR. I do have an internal plan to overhaul the status bar display to include more status updates. This PR may fit into that. I'll leave this PR open, and come back when the plan is finished.

ManuelHentschel:
Any update on this? If the status bar overhaul is taking longer, could it be considered to merge this PR in the meanwhile?

PR #4061, opened 2023-11-16
Re-structure project source directory

James-Yu:
This PR aims at creating a more straightforward source code directory under `src/`.

The original one was used for seven years, seeing the project to grow from a small toy with a few source files to a large one as is. The old component-provider structure no longer being meaningful with the growth of features.

The new directory structure re-organize all source files according to their nature in LaTeX-Workshop. As all file moves are performed with `git mv`, the file editing history and `git blame` are preserved. Hopefully this won't be a big problem for later development.

I consider this change to be necessary in order to maintain the project for the upcoming years and make the project less messy to new and potential contributors and maintainers.

The next steps after this PR will be 1) extract interfaces and types, 2) decouple the source codes, which I had been doing terribly, 3) re-visit the math preview files for a possible refactor with `unified-latex`.


James-Yu:
@jlelong May I have your insights on it? I'm sorry that the file changes may look scary, yet there's only two things: 1) move files and 2) fix imports. So everything works as it was 😆

James-Yu:
After some test, `git mv` turned out not a good way of preserving file history. I will look for other solutions.

PR #4097, opened 2023-12-15
Support filename conversion for error messages

westpaddy:
This RP adds a function that can convert the filename of an error message before outputting it to the log panel.

The primary purpose of this extension is to enhance support for TeX writing, especially in cases where TeX files undergo preprocessing before typesetting. In such scenarios, the files compiled by a TeX processor may differ from the ones the author originally worked on. This discrepancy often leads to difficulties in locating error positions when using VSCode support.

It's important to note that this extension may not perform optimally with general preprocessing tasks, such as altering the line positions of text before and after preprocessing. However, it is well-suited for lightweight preprocessing, where line positions remain unchanged. Therefore, this extension is particularly beneficial for authors employing such a workflow.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2091,6 +2091,18 @@
           "default": true,
           "markdownDescription": "Convert the encoding of filenames if necessary when displaying them in the problems panel."
         },
+        "latex-workshop.message.filenameConverter.matcher": {
+          "scope": "window",
+          "type": "string",
+          "default": "",
+          "markdownDescription": "A regular expression to match the filename of an error message. The matched part of the filename is replaced by `#latex-workshop.message.filenameConverter.replacer#` before sending to the log panel. The syntax of regular expressions and replacers is one for `String.replace()` in JavaScript."
+        },
+        "latex-workshop.message.filenameConverter.replacer": {
+          "scope": "window",
+          "type": "string",
+          "default": "",
+          "markdownDescription": "A string specifying the replacement of the filename converting function. See `#latex-workshop.message.filenameConverter.matcher#` for the details."
+        },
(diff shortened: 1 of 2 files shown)


James-Yu:
Can you please provide some use case examples to demonstrate this PR's necessity? Thank you.

westpaddy:
As a concrete workflow, I (and some people) use the ott tool (https://github.com/ott-lang/ott) to write a paper. The tool can translate strings enclosed by `[[` and `]]` in a given text file into the corresponding strings defined by user-given rules. The details of the tool are not important here.

In the workflow, for instance, I write the following latex file, named `example.otex`.
```
\documentclass{article}
\begin{document}
$[[ (\x.M) V --> M[V/x] ]]$
\FOO{}
\end{document}
```
Then, I preprocess the file and obtain the following intermediate file, named `example.tex`.
```
\documentclass{article}
\begin{document}
$(\lambda x . M) \longrightarrow M[V/x]$
\FOO{}
\end{document}
```
Finally, I try to compile the obtained file, but an error happens since there is an undefined macro `\FOO`.
(The process is automated by Makefile.)

The problem is error log says the error is in `example.tex`.
That is correct behavior, but it results in inconveniences such as clicking an error log jump to `example.tex`, and error lens extension (https://github.com/usernamehw/vscode-error-lens) does not work on the original `example.otex`.

James-Yu:
Unfortunately, this feature won't be integrated in LW as its introduced maintenance outweigh its potential user base. You may consider submitting the feature to other utility extensions such as https://github.com/tecosaur/LaTeX-Utilities .

PR #4119, opened 2023-12-31
Override pdf.js to enable VIM-like scrolling by j, k, J, K, H, L

taquangtrung:
Hi,

I override some default keys of pdf.js to enable a VIM-like scrolling experience:
- `j, k` for a small vertical scrolls like VIM (by default in pdf.js, `j, k` will scroll pages)
- `J, K` now scroll pages
- `H, L` for small horizontal scrolls (I didn't assign to `h` since it is used by pdf.js to select the hand tool).

Can you consider merging this PR?

Thanks!


James-Yu:
I've modified your PR a bit:

- Instead of changing the default behavior of `j` and `k`, I'd rather suggest adding new shortcuts `shift+jkhl` for small navigations to avoid changing current behavior.
- The previous implementation is verbose.

PR #4122, opened 2024-01-02
use auto scroll instead of smooth to avoid lagging in repeating scroll

taquangtrung:
Hi,

I made a small change to the previous merged PR.

In your modified code, if the behavior for repeating scroll is `smooth`, which can cause lagging when scroll. 

I don't know what the reason of that, but changing that behavior to `auto` will fix the issue.

### DIFF
--- a/viewer/latexworkshop.ts
+++ b/viewer/latexworkshop.ts
@@ -647,14 +647,15 @@ class LateXWorkshopPdfViewer implements ILatexWorkshopPdfViewer {
                 const container = document.getElementById('viewerContainer') as HTMLElement
 
                 const configMap: {[key: string]: ScrollToOptions} = {
-                    'J': { top: evt.repeat ? 20 : 40 },
-                    'K': { top: evt.repeat ? -20 : -40 },
-                    'H': { left: evt.repeat ? -20 : -40 },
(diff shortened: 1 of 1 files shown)


James-Yu:
I don't get the change of `20` -> `30`. Indeed, why a different value is required?

taquangtrung:
> I don't get the change of `20` -> `30`. Indeed, why a different value is required?

I use the value 30 to make the scroll a bit faster. I tested when the value of 20 and feel it's a bit slow. 

The value `40` is to make the scroll amount exactly like when pressing the up/down arrows.
When using up/down arrow, somehow the smooth scrolling feature works well without lagging.

For the implementation in the PR, using smooth scroll for repeating scrolling will cause lagging. So, I changed to use `auto` (actually it's `instant` scrolling), I used a smaller scroll step (20 or 30) to make the experience smoother than 40.

James-Yu:
I'm not sure whether the "lagging" you quoted was the same thing I observed, which is the "acceleration" of scrolling before finishing the previous scroll when holding the keys. If so, that is indeed normal in electron webview and should be preserved. This can be quite opinionated, and I tend to stick to what the upstream (electron/vscode) provides.

PR #4148, opened 2024-01-28
update expl3.json with new macros and remove deprecated macros

mbertucci47:
This PR updates the expl3.json completion file with new macros provided by l3kernel and removes deprecated macros, including many of the `x`-type variants. It also puts the macros in alphabetical order which makes the diff difficult to analyze, so apologies for that.


James-Yu:
May I know how this file is generated in this PR? As we essentially use `dev/latex3command.py` to generate this file, it would be very helpful if you may also update the script.

mbertucci47:
Ah I see, now it makes sense why the commands were in seemingly random order. I just updated the json file by hand and haven't taken the time to figure out what the python script is doing. If the script generates an up-to-date json file and simply hasn't been run in a while, then feel free to close this and update the usual way

James-Yu:
The script indeed hasn't been run for some time and may be outdated. Did you use some sort of scripts to generate this file? If so, it would be great for maintenance.

mbertucci47:
No, it was very low tech. I literally just alphabetized the existing file with my editor, removed the commands found in l3deprecation.dtx, ran `\cs_meaning:N` on the remaining ones to see which were undefined and removed those, then added whatever was missing from a quick scan of interface3.

James-Yu:
I'm a bit torn on whether this should be merged. Definitely an improvement but not likely maintainable. What do you think @jlelong ?

mbertucci47:
I was able to run latex3command.py after updating the path in `dtx_files` for texlive 2023 (soon 2024) and changing `expl3.cmds` to `expl3.macros`. It gave the correct list, removing deprecated commands and adding new ones. So I would close this PR and keep the script. I no next to no Python but I'd assume alphabetizing the list wouldn't be that hard?

jlelong:
I have updated the script and added some sorting too. The new completion list has been generated too.
See #4151

PR #4151, opened 2024-01-30
Update LaTeX3 completion

jlelong:
Related to #4148 

The script is updated and the completion list sorted.
@mbertucci47 What do you think of the list?


mbertucci47:
Looks good to me!

PR #4175, opened 2024-03-01
Update leading string for doctex line comments

huangyxi:
I changed the leading string for line comments in doctex files to `%^^A`.
This would cause the line to be ignored in all scenarios: DocStrip by `ins` file, TeX, and LaTeX.

https://tex.stackexchange.com/questions/415738/is-it-possible-to-make-comments-in-dtx-file#comment1040003_415744

### DIFF
--- a/syntax/doctex-language-configuration.json
+++ b/syntax/doctex-language-configuration.json
@@ -1,6 +1,6 @@
 {
 	"comments": {
-		"lineComment": "^^A"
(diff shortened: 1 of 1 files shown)


James-Yu:
We have outsourced our LaTeX syntax highlighting grammar to [vscode-latex-basics](https://github.com/jlelong/vscode-latex-basics). Please open an issue there.

Issue #4168, opened 2024-02-26
SyncTex forward indicator not at the right place in web brower


pyk1998:
> _Please fill the following fields with a star (*) and provide as much related information as possible._

## Pre-checks*
> _Please change the following `[ ]` to `[x]` for confirmation._
- [x] The issue has not been reported in this repository.
- [x] The issue remains after disabling all other extensions and restarting Visual Studio Code.
- [x] The [FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) cannot address the issue.
- [x] The issue is not related to compiling a document, or the document can be successfully compiled in the OS terminal but not in Visual Studio Code with this extension.

## Environment*
> _Please write exact version numbers instead of descriptors such as `latest`._
- Operating System: Windows 10
- Visual Studio Code Version: 1.86.2
- LaTeX Workshop Version: 9.18.2
- TeX Distribution Version: TeX Live 2021

> _Please list the environment and version number if you are using VSCodium, Snap or Flatpack versions of Visual Studio Code, and/or Visual Studio Code Remote Containers/SSH/WSL._
- None

## The Issue*
> _Please briefly describe the issue you come across._
- The synctex forward indicator shows a random location of the sentence in the web browser.

### Reproduction Steps
> _Please list out the steps to reproduce your bug. Include relevant environmental variables or any other configuration._
1. select a sentence
2. select "SyncTex from cursor"
3. View the location of the indicator in the web browser

### Expected Behavior
> _What were you expecting to see? Include any relevant examples or documentation links._
-  The indicator should appear at the beginning of the sentence or the exact word.
-  Demo GIF in documentation: https://github.com/James-Yu/LaTeX-Workshop/wiki/View

### LaTeX Workshop Output*
> _Please paste the whole log messages below, not parts of ones. The log should start with `New log placeholder %WS1% registered`._
```
[16:26:03.942][Logger] New log placeholder %WS1% registered for e:\OneDrive - ****\Work2\writing .
[16:26:03.942][Extension] Initializing LaTeX Workshop.
[16:26:04.014][Build][Recipe] Set $LATEXWORKSHOP_DOCKER_LATEX: ""
[16:26:04.064][Server] Creating LaTeX Workshop http and websocket server.
[...]

Issue #4186, opened 2024-03-02
SyncTeX binary precedence over synctex.js?


James-Yu:
[Internal Discussion]

As per https://github.com/James-Yu/LaTeX-Workshop/issues/4181#issuecomment-1974758362 and #4168, I did not find the column-locating code for synctexjs we have been using. I also backtrace a few years code in the past and did not find that. Strangely both you @jlelong and I seem to have an impression on the feature.

I also checked (roughly) the code on invoking SyncTeX binary and figured that the binary indeed returns with a column index:
https://github.com/James-Yu/LaTeX-Workshop/blob/6c63cc3e64f782a6dbba0d2e98ec5ebe5b927878/src/locate/synctex.ts#L98-L101
instead of the 0-column of synctexjs:
https://github.com/James-Yu/LaTeX-Workshop/blob/6c63cc3e64f782a6dbba0d2e98ec5ebe5b927878/src/locate/synctex/worker.ts#L236
But now we have `latex-workshop.synctex.synctexjs.enabled` defaults to `true`:
https://github.com/James-Yu/LaTeX-Workshop/blob/6c63cc3e64f782a6dbba0d2e98ec5ebe5b927878/package.json#L1682-L1687

What about we return to first try the binary, if anything went wrong (not found, non-zero return), we fallback to the javascript implementation to balance the utility and availability? We may also remove this config if so, or change the purpose of it to be overriding this new logic. May I have your thoughts? @jlelong

jlelong:
I have good news: we are not crazy yet! I have managed to trace back to the column-locating code. It was introduced in #1161. It is still alive

https://github.com/James-Yu/LaTeX-Workshop/blob/ee13477a5a2bf94ed107f3de2572aa55f96423c0/src/locate/synctex.ts#L471-L492

Ctrl-clicking inside the pdf calls https://github.com/James-Yu/LaTeX-Workshop/blob/614eec118aa95a390237d0ec60fd15d7edb2b18d/src/preview/viewer.ts#L300

If I understand correctly, synctex does provide a column number but it can be quite inaccurate. The function `getRowAndColumn` tries to fix it.

James-Yu:
Thats awesome! I will look into the implementation c.f. #4181 #4168

PR #4194, opened 2024-03-08
Fix forward SyncTex accuracy in internal browser by providing a range location indication alternative

pyk1998:
This proposed solution for issue #4168  involves implementing a range indicator in the PDF viewer based on the line position in the editor. The solution includes several major modifications:

- Adding a configuration option to choose the indicator type between indicating a single location or a range.
- Ignoring the column parameter when invoking the SyncTex binary when indicating a range.
- Parsing all the SyncTex binary records and extracting the box (rectangle) coordinates.
- Handling the list of records in the viewer, calculating the rectangular coordinates, and displaying them.
- Adding CSS animation to show the corresponding area in the PDF.

To use this solution, follow these steps:
- Uncheck the configuration option `latex-workshop.synctex.synctexjs.enabled`.
- Choose the `Range` option in the `latex-workshop.synctex.indicator.type` configuration.

By implementing these changes, the PDF viewer will display a range rectangular indicator that corresponds to the selected line in the editor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1686,7 +1686,7 @@
             "Indicates a possible range with a rectangular SyncTeX indicator.",
             "Indicates a single specific spot with a red circle SyncTeX indicator"
           ],
-          "default": "Range",
+          "default": "Spot",
           "markdownDescription": "Define the behavior of SyncTex indicator after a forward SyncTeX in the PDF viewer (tab or browser). Valid when use SyncTex binary (latex-workshop.synctex.synctexjs.enabled = false)."
         },
         "latex-workshop.synctex.afterBuild.enabled": {
--- a/package.json
+++ b/package.json
@@ -1672,16 +1672,16 @@
           "scope": "window",
           "type": "string",
           "enum": [
-            "disabled",
-            "spot",
-            "range"
+            "none",
+            "circle",
+            "rectangle"
           ],
           "enumDescriptions": [
-            "Disable indicator.",
-            "Indicates a single specific spot with a red circle SyncTeX indicator.",
-            "Indicates a possible range with a blue rectangular SyncTeX indicator. (Valid when not using synctex.js)"
+            "Hide the indicator.",
+            "Indicates a possible location with a red circular SyncTeX indicator.",
+            "Indicates the whole line selected in the Tex file with a red rectangular SyncTeX indicator. (Valid when not using synctex.js)"
           ],
-          "default": "spot",
+          "default": "circle",
           "markdownDescription": "Define the visibility and style of SyncTeX indicator after a forward SyncTeX in the PDF viewer."
         },
         "latex-workshop.synctex.afterBuild.enabled": {
--- a/package.json
+++ b/package.json
@@ -1672,7 +1672,22 @@
           "scope": "window",
           "type": "boolean",
           "default": true,
-          "markdownDescription": "Define the visibility of SyncTeX indicator (a red highlighting circle) after a forward SyncTeX in the PDF viewer."
+          "markdownDescription": "Define the visibility of SyncTeX indicator after a forward SyncTeX in the PDF viewer."
+        },
+        "latex-workshop.synctex.indicator.type": {
+          "scope": "window",
+          "type": "string",
+          "when": "!latex-workshop.synctex.synctexjs.enabled",
+          "enum": [
+            "Range",
+            "Spot"
+          ],
+          "enumDescriptions": [
+            "Indicates a possible range with a rectangular SyncTeX indicator.",
+            "Indicates a single specific spot with a red circle SyncTeX indicator"
+          ],
+          "default": "Range",
+          "markdownDescription": "Define the behavior of SyncTex indicator after a forward SyncTeX in the PDF viewer (tab or browser). Valid when use SyncTex binary (latex-workshop.synctex.synctexjs.enabled = false)."
         },
         "latex-workshop.synctex.afterBuild.enabled": {
           "scope": "window",
--- a/package.json
+++ b/package.json
@@ -1670,24 +1670,19 @@
         },
         "latex-workshop.synctex.indicator.enabled": {
           "scope": "window",
-          "type": "boolean",
-          "default": true,
-          "markdownDescription": "Define the visibility of SyncTeX indicator after a forward SyncTeX in the PDF viewer."
-        },
-        "latex-workshop.synctex.indicator.type": {
-          "scope": "window",
           "type": "string",
-          "when": "!latex-workshop.synctex.synctexjs.enabled",
           "enum": [
-            "Range",
-            "Spot"
+            "disabled",
+            "spot",
+            "range"
           ],
           "enumDescriptions": [
-            "Indicates a possible range with a rectangular SyncTeX indicator.",
-            "Indicates a single specific spot with a red circle SyncTeX indicator"
+            "Disable indicator.",
+            "Indicates a single specific spot with a red circle SyncTeX indicator.",
+            "Indicates a possible range with a blue rectangular SyncTeX indicator. (Valid when not using synctex.js)"
           ],
-          "default": "Spot",
-          "markdownDescription": "Define the behavior of SyncTex indicator after a forward SyncTeX in the PDF viewer (tab or browser). Valid when use SyncTex binary (latex-workshop.synctex.synctexjs.enabled = false)."
+          "default": "spot",
+          "markdownDescription": "Define the visibility and style of SyncTeX indicator after a forward SyncTeX in the PDF viewer."
         },
         "latex-workshop.synctex.afterBuild.enabled": {
           "scope": "window",
--- a/src/locate/synctex.ts
+++ b/src/locate/synctex.ts
@@ -242,7 +242,7 @@ function toPDF(args?: {line: number, filePath: string}, forcedViewer: 'auto' | '
             if (!record) {
                 return
             }
-            void lw.viewer.locate(pdfFile, record, 'spot')
+            void lw.viewer.locate(pdfFile, record)
         } catch (e) {
             logger.logError('Forward SyncTeX failed.', e)
         }
@@ -267,7 +267,7 @@ function toPDF(args?: {line: number, filePath: string}, forcedViewer: 'auto' | '
 
         void callSyncTeXToPDF(line, character, filePath, pdfFile, indicatorType).then( (record) => {
             if (pdfFile) {
-                    void lw.viewer.locate(pdfFile, record, indicatorType)
+                    void lw.viewer.locate(pdfFile, record)
             }
         })
 
--- a/src/locate/synctex.ts
+++ b/src/locate/synctex.ts
@@ -246,7 +246,7 @@ function toPDF(args?: {line: number, filePath: string}, forcedViewer: 'auto' | '
         const indicatorType = configuration.get('synctex.indicator.type')
 
         if (indicatorType === 'Range') {
-            void callSyncTeXToPDFRange(line, character, filePath, pdfFile).then( (recordList) => {
+            void callSyncTeXToPDFRange(line, filePath, pdfFile).then( (recordList) => {
                 if (pdfFile) {
                     void lw.viewer.locateRange(pdfFile, recordList)
                 }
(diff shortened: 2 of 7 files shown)


James-Yu:
Thanks for the contribution. Before looking into the implementation (which is good, just a few function merges to comply with the practice), may I know if this feature is portable to synctex.js?

@jlelong It looks like the binary is still providing a bit more info about the location over synctex.js, making #4186 valid. What do you think of adopting a priority of execution (binary > javascript) to balance utility (binary) and availability (js as fallback)?

pyk1998:
@James-Yu Thank you for the review. If there's any modification needed, please let me know. I'm very happy to have a chance learning from the community.

I've checked the related code, it looks also possible to apply the same logic to synctex.js.

By default, the forward search in synctex.js also returns a single location:
https://github.com/James-Yu/LaTeX-Workshop/blob/2128e1ea0599bf7c9dee0365469de2f859532186/src/locate/synctex/worker.ts#L75-L90 

However, it constains a lot of low-level code handling blocks in synctex file:
https://github.com/James-Yu/LaTeX-Workshop/blob/2128e1ea0599bf7c9dee0365469de2f859532186/src/locate/synctex/worker.ts#L156-L176

To me, it is not as intuitive to modify them as simply invoking the binary. I will try to analyze them.

pyk1998:
Having tried to modifed synctex.js with no luck.
Maybe the low-level synctex block handling code seems to be incomplete?

According to the comment, the synctex.js seems not handling all the blocks:
https://github.com/James-Yu/LaTeX-Workshop/blob/2128e1ea0599bf7c9dee0365469de2f859532186/src/locate/synctex/worker.ts#L75-L77

A similar but more complete parser is the one used in SumatraPDF:
https://github.com/sumatrapdfreader/sumatrapdf/blob/85a0fa0f09d729e06c7f6e14bf55445723fe0304/ext/synctex/synctex_parser.c#L3310-L3429

 (Note that the latest parser used in https://github.com/jlaurens/synctex/blob/main/synctex_parser.c have changed from these two greatly.)

This caused the record returned from the synctex.js is not enough for viewer showing the range of the forward syntex location.
From my simple opinon, using the binary directly is more convenient and easy to maintenance.

jlelong:
> @jlelong It looks like the binary is still providing a bit more info about the location over synctex.js, making #4186 valid. What do you think of adopting a priority of execution (binary > javascript) to balance utility (binary) and availability (js as fallback)?

I agree that we can change the order and put the binary first and only keep the js implementation as a fallback.

package.json
```diff
@@ -1672,7 +1672,22 @@
...
+          "enum": [
+            "Range",
+            "Spot"
+          ],
+          "enumDescriptions": [
+            "Indicates a possible range with a rectangular SyncTeX indicator.",
+            "Indicates a single specific spot with a red circle SyncTeX indicator"
+          ],
+          "default": "Spot",
+          "markdownDescription": "Define the behavior of SyncTex indicator after a forward SyncTeX in the PDF viewer (tab or browser). Valid when use SyncTex binary (latex-workshop.synctex.synctexjs.enabled = false)."
```
James-Yu, line 1690: Instead of creating a new config, please consider extending the existing boolean to be an enum, and use fallback logic if a boolean value is provided. Kindly add a tiny comment on the fallback processing.

src/locate/synctex.ts
```diff
@@ -247,6 +319,65 @@ function callSyncTeXToPDF(line: number, col: number, filePath: string, pdfFile:
...
+ * This function calls the SyncTeX binary to retrieve PDF information for a
+ * given line as a range in a TeX file. It returns a promise
+ * resolving to a SyncTeXRecordToPDFAllList object.
+ *
+ * @param line - The line number in the TeX file.
+ * @param filePath - The path of the TeX file.
+ * @param pdfFile - The path of the PDF file.
+ * @returns A promise resolving to a SyncTeXRecordToPDFAllList object.
+ */
+function callSyncTeXToPDFRange(line: number, filePath: string, pdfFile: string): Thenable<SyncTeXRecordToPDFAllList> {
```
James-Yu, line 334: This function is largely similar to the existing `callSyncTeXToPDF`. Please instead merge the two functions and use a toggle parameter to distinguish the logic differences.

src/locate/synctex.ts
```diff
@@ -181,11 +243,21 @@ function toPDF(args?: {line: number, filePath: string}, forcedViewer: 'auto' | '
...
+                    void lw.viewer.locateRange(pdfFile, recordList)
+                }
+            })
+        } else {
+            void callSyncTeXToPDF(line, character, filePath, pdfFile).then( (record) => {
+                if (pdfFile) {
+                    void lw.viewer.locate(pdfFile, record)
+                }
+            })
+        }
```
James-Yu, line 260: If they are merged, simplify here.

src/preview/viewer.ts
```diff
@@ -414,6 +415,38 @@ async function locate(pdfFile: string, record: SyncTeXRecordToPDF): Promise<void
     }
 }
 
+/**
+ * Reveals the position range of `records` on the internal PDF viewers with a Rectangle indicator.
+ *
+ * @param pdfFile The path of a PDF file.
+ * @param recordList The list of positions to be revealed.
+ */
+async function locateRange(pdfFile: string, recordList: SyncTeXRecordToPDFAllList): Promise<void> {
```
James-Yu, line 424: Merge with `locate`.

src/types.ts
```diff
@@ -108,6 +108,20 @@ export type SyncTeXRecordToTeX = {
...
+    h: number;
+    v: number;
+    W: number;
+    H: number;
+}
+
+export type SyncTeXRecordToPDFAllList = {
+    records: SyncTeXRecordToPDFAll[];
+}
+
```
James-Yu, line 124: I don't think we need to explicitly define this type. What are the cases when `SyncTeXRecordToPDFAll[]` won't work?

src/types.ts
```diff
@@ -108,6 +108,20 @@ export type SyncTeXRecordToTeX = {
...
 
+export type SyncTeXRecordToPDFAll = {
+    Page: number;
+    x: number;
+    y: number;
+    h: number;
+    v: number;
+    W: number;
+    H: number;
+}
```
James-Yu, line 119: This type should be extended from `SyncTeXRecordToPDF`. If the original `indicator` has no use, either make it optional in parent and add corresponding logic, or ignore the value with proper comment annotation.

types/latex-workshop-protocol-types/index.d.ts
```diff
@@ -1,3 +1,14 @@
+
+type SyncTeXRecordToPDFAll = {
+    Page: number;
+    x: number;
+    y: number;
+    h: number;
+    v: number;
+    W: number;
+    H: number;
+}
```
James-Yu, line 10: Ditto.

viewer/latexworkshop.ts
```diff
@@ -346,6 +346,65 @@ class LateXWorkshopPdfViewer implements ILatexWorkshopPdfViewer {
         // }, 1000)
     }
 
+    private forwardSynctexRange(position: { records: SyncTeXRecordToPDFAll[], indicator: boolean }) {
```
James-Yu, line 349: Merge with `forwardSynctex`. Reuse common codes.

viewer/latexworkshop.ts
```diff
@@ -482,6 +541,10 @@ class LateXWorkshopPdfViewer implements ILatexWorkshopPdfViewer {
                     this.forwardSynctex(data.data)
                     break
                 }
+                case 'synctexRange': {
+                    this.forwardSynctexRange(data.data)
+                    break
+                }
```
James-Yu, line 547: Should be merable with `'synctex'` case with previous changes?

James-Yu (changes requested):
Thank you very much for this contribution. Please consider the general requests on code structure.

pyk1998:
@James-Yu Thank you so much for the detailed and kind advice. I have tried my best to merge the functions and types. However, I found that the `forwardSynctex` seems a lot different from `forwardSynctexRange` from both parameters and internal logic. I think we can jut let them separate?

pyk1998:
@James-Yu I have also tried to merge `forwardSynctexRange` with `forwardSynctex` in viewer in the latest commit. Please kindly review all the refactors.

James-Yu:
Thanks for the changes. Will review soon (but not too soon).

src/locate/synctex.ts
```diff
@@ -181,11 +247,30 @@ function toPDF(args?: {line: number, filePath: string}, forcedViewer: 'auto' | '
...
+
+        const indicatorConfig = configuration.get('synctex.indicator.enabled')
+        
+        if (typeof indicatorConfig === "boolean") {
+            // if configuration is boolean in previous version, then use fallback logic.
+            if (indicatorConfig) {
+                indicatorType = "spot"
+            } else {
+                indicatorType = "disabled"
+            }
```
James-Yu, line 260: `indicatorType = indicatorConfig ? 'spot' : 'disabled'`

package.json
```diff
@@ -1670,9 +1670,19 @@
         },
         "latex-workshop.synctex.indicator.enabled": {
           "scope": "window",
-          "type": "boolean",
-          "default": true,
-          "markdownDescription": "Define the visibility of SyncTeX indicator (a red highlighting circle) after a forward SyncTeX in the PDF viewer."
+          "type": "string",
+          "enum": [
+            "disabled",
```
James-Yu, line 1675: "none"

package.json
```diff
@@ -1670,9 +1670,19 @@
...
+            "spot",
```
James-Yu, line 1676: "circle"

package.json
```diff
@@ -1670,9 +1670,19 @@
...
+            "range"
```
James-Yu, line 1677: "rectangle"

package.json
```diff
@@ -1670,9 +1670,19 @@
...
+          ],
+          "enumDescriptions": [
+            "Disable indicator.",
+            "Indicates a single specific spot with a red circle SyncTeX indicator.",
+            "Indicates a possible range with a blue rectangular SyncTeX indicator. (Valid when not using synctex.js)"
```
James-Yu, line 1682: Change accordingly.

James-Yu: By the way, why blue rectangle instead of red?

src/locate/synctex.ts
```diff
@@ -200,12 +285,21 @@ function toPDF(args?: {line: number, filePath: string}, forcedViewer: 'auto' | '
...
     const configuration = vscode.workspace.getConfiguration('latex-workshop')
     const docker = configuration.get('docker.enabled')
-    const args = ['view', '-i', `${line}:${col + 1}:${docker ? path.basename(filePath): filePath}`, '-o', docker ? path.basename(pdfFile): pdfFile]
+
+    let args: string[]
+    if (indicatorType === 'range') {
+        args = ['view', '-i', `${line}:0:${docker ? path.basename(filePath): filePath}`, '-o', docker ? path.basename(pdfFile): pdfFile]
+    } else {
+        args = ['view', '-i', `${line}:${col + 1}:${docker ? path.basename(filePath): filePath}`, '-o', docker ? path.basename(pdfFile): pdfFile]
+    }
```
James-Yu, line 302:
```typescript
const args = ['view', '-i'].concat([
    `${line}${indicatorType === 'range' ? ':0' : `:${col + 1}`}:${docker ? path.basename(filePath) : filePath}`,
    '-o',
    docker ? path.basename(pdfFile) : pdfFile
])
```

src/locate/synctex.ts
```diff
@@ -236,15 +330,20 @@ function callSyncTeXToPDF(line: number, col: number, filePath: string, pdfFile:
...
         proc.on('exit', exitCode => {
             if (exitCode !== 0) {
                 logger.logError(`(${logTag}) Forward SyncTeX failed.`, exitCode, stderr)
             } else {
-                resolve(parseToPDF(stdout))
+                if (indicatorType === 'range') {
+                    resolve(parseToPDFList(stdout))
+                } else if (indicatorType === 'spot') {
+                    resolve(parseToPDF(stdout))
+                }
```
James-Yu, line 342: resolve(indicatorType === 'range' ? parseToPDFList(stdout) : parseToPDF(stdout))

viewer/latexworkshop.ts
```diff
@@ -304,46 +304,100 @@ class LateXWorkshopPdfViewer implements ILatexWorkshopPdfViewer {
...
+
+        if (Array.isArray(data)){
+            const indicatorTemplate = document.getElementById('synctex-indicator') as HTMLElement
+            const indicatorParent = indicatorTemplate.parentNode as HTMLElement
+
+            for (const record of data) {
+                const pos_left_top = PDFViewerApplication.pdfViewer._pages[record.page - 1].viewport.convertToViewportPoint(record.h, record.v - record.H)
+                const pos_right_down = PDFViewerApplication.pdfViewer._pages[record.page - 1].viewport.convertToViewportPoint(record.h + record.W, record.v)
+    
+                const width_px = pos_right_down[0] - pos_left_top[0]
```
James-Yu, line 323: This new implementation and below until end of function is a bit long. I'd advice add key comments to if and for conditions at least, to improve readability.

James-Yu (changes requested):

pyk1998:
@James-Yu I think I have changed the code according to the comments. Any further updates?

James-Yu:
Not having time to touch this. There are indeed a few issues with code style, yet I will directly commit to the PR shortly.

pyk1998:
@James-Yu Thank you so much for the abundant and kind work on the code style. I've really got so many lessons to learn from the followed up commits. Thank you again for the maintanence of such great extension, and I am so happy having a chance to participate. Wish you all the best.

PR #4217, opened 2024-03-28
Add padding to pdf viewer

weisJ:
Makes it easer to closely look at the top/bottom of the first/last page of the document.

### DIFF
--- a/viewer/latexworkshop.css
+++ b/viewer/latexworkshop.css
@@ -112,6 +112,10 @@ html[dir='rtl'] .findbar {
     top: 0;
 }
 
+#viewer.pdfViewer {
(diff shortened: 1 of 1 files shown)


weisJ:
Without padding: Top of document is hidden by the toolbar
SCR-20240402-ttew

With padding: Toolbar doesn't hide the top of the document. The toolbar not shown as it is 32px = 2em high.
SCR-20240402-ttzo

James-Yu:
Unfortunately, this PR cannot be merged as it breaks the restoration of viewer state, which is computed based on the absolute vertical distance from the bounding box. There is no easy fix afaik.

PR #4220, opened 2024-04-02
Bugfix: Synctex does not work while using vscode in browser

wasalm:
Although [code-server](https://github.com/coder/code-server) is not officially supported, hereby a bugfix making (ctrl) + click work in the browser.

Namely, before synctex is enabled, it waits before PDF.js is loaded. PDF.js signals this by sending a custom "webviewloaded" event. It first tries to send it to the parent window and if this fails, it send it to its own window.
in code-server, this parent _exists_ as a iframe, hence this event is where there are no listeners. Issue is fixed by added a listener to the parent window if possible.

### DIFF
--- a/viewer/latexworkshop.ts
+++ b/viewer/latexworkshop.ts
@@ -33,7 +33,7 @@ class LateXWorkshopPdfViewer implements ILatexWorkshopPdfViewer {
     private readonly webViewerLoaded: Promise<void> = new Promise((resolve) => {
         document.addEventListener('webviewerloaded', () => resolve() )
 
-        // Bugfix when using LaTeXWorkshop with web based version of vscode
+        // https://github.com/James-Yu/LaTeX-Workshop/pull/4220#issuecomment-2034520751
         try {
           parent.document.addEventListener('webviewerloaded', () => resolve() )
         } catch(err) { /* do nothing */ }
--- a/viewer/latexworkshop.ts
+++ b/viewer/latexworkshop.ts
@@ -35,7 +35,7 @@ class LateXWorkshopPdfViewer implements ILatexWorkshopPdfViewer {
 
         // Bugfix when using LaTeXWorkshop with web based version of vscode
         try {
-          parent.document.addEventListener('webviewerloaded', () => resolve() );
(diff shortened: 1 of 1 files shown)


James-Yu:
This patch will add two listeners to the event if the webpage is not in code-server. `parent.document` exists even in such environments.

wasalm:
Agree, it will create two listeners. However, I don't think this will create an issue because

1. the event is only fired once in [viewer.mjs](https://github.com/James-Yu/LaTeX-Workshop/blob/14fef41ad0f3146fa7ee0f1ccae4669d129ab5b1/viewer/viewer.mjs#L14884) between the lines 14884 and 14896. It is either fired in parent.document or document itself.
2. in the hypothetical case it would fire multiple times, the promise webViewerLoaded in latexworkshop.ts will only be resolved once, because any second call to resolve() will be ignored.
3. This event will only be called once, and hence this extra overhead will not be significant.

James-Yu:
What will be the value of `document` in a code-server env? If `undefined`, `document ?? parent.document` should work well.

wasalm:
Also it is good to understand why the origional code does work in vscode. The event "webviewerloaded" is only created in [viewer.mjs](https://github.com/James-Yu/LaTeX-Workshop/blob/14fef41ad0f3146fa7ee0f1ccae4669d129ab5b1/viewer/viewer.mjs#L14884). Here it tries to dispatch this event first in `parent.document` and if it fails it dispatches it in `document` itself.

The listener for the "webviewerloaded" event is in the file [latexworkshop.ts](https://github.com/James-Yu/LaTeX-Workshop/blob/8486201fe70b652d1c896bf13d6fc7bc6ef3352f/viewer/latexworkshop.ts#L824). It shares the same `document` with viewer.mjs, because viewer.mjs gets loaded by latexworkshop.ts at line 824. Hence in the origional version, the code will only work if it gets dispatched in `document` and not in `parent.document`. 

For the code-server the parent document and document are on the same origin and hence it is allowed to dispatch events in the parents. I expect that in vscode this will be different and hence when it tries to dispatch in parent, it will give an error.

Issue #1130, opened 2019-01-20
[Feature Request] Podman support


AdrianKoshka:
**Is your feature request related to a problem? Please describe.**
My FR isn't related to any problem.

**Describe the solution you'd like**
There already is `docker` support, I'd like there to be [`podman`](https://podman.io/) support. I'd prefer to use `podman` over `docker` because `podman` can be used to create rootless containers, and doesn't require extra privileges like `docker` does. Perhaps this could be done via a configuration option that let's you choose whether to use `docker` or `podman`.

**Describe alternatives you've considered**
I haven't considered any.

**Additional context**

`podman` github page: https://github.com/containers/libpod

James-Yu:
Unfortunately, before the referenced software became an industry-standard-like-thing, there is no plan to implement such supports from my side. PRs always welcomed.

Note that even docker only gets experimental support. I myself does not use it, and has no way to test the implementation. This is why #1069 is still open.

AdrianKoshka:
Oh okay, `podman` is OCI compatible and docker compatible fwiw. So it's more or less a drop-in replacement. Understandable though.

ebousse:
Sorry to post while the issue is marked as closed, but I simply wanted to mention that possibly one simple way to solve this would be to add a configuration option in latex-workshop to allow users to choose the path to the used docker binary. This would make it possible to point to `podman` instead, which as @AdrianKoshka mentionned is a drop-in replacement in the simpler cases (such as this one).

PR #4222, opened 2024-04-03
Customize docker (or podman) path

themagiulio:
Currently LaTeX Workshop supports only `docker` and it expects to have it in PATH. This small PR addresses this problem by creating a new config `docker.path` allowing the user to customize the location of `docker`.

This config makes also possible to use `podman`, a drop-in replacement of `docker` as explained in #1130, by setting `docker.path` to `podman`.

My commits makes this to work only in GNU/Linux and macOS as I modified only the bash scripts. Modifying the powershell scripts would make this fix available to Windows users as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2105,6 +2105,12 @@
           "default": "",
           "markdownDescription": "Define the image for `latexmk`, `synctex`, `texcount`, and `latexindent`."
         },
+        "latex-workshop.docker.path": {
+          "scope": "window",
+          "type": "string",
+          "default": "docker",
+          "markdownDescription": "Docker (or Podman) executable name or path."
+        },
         "latex-workshop.showContextMenu": {
           "scope": "window",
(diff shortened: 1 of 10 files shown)


James-Yu:
Thank you for your contribution. I vote a yes to this PR.

> My commits makes this to work only in GNU/Linux and macOS as I modified only the bash scripts.

This is the remaining concern. Generally, we'd like to avoid platform-specific configs. Any solution?

themagiulio:
Thank you for creating this great extension.

> This is the remaining concern. Generally, we'd like to avoid platform-specific configs. Any solution?

We just need to edit the bat files to invoke the program specified by the env variable `LATEXWORKSHOP_DOCKER_PATH` instead of `docker`. I am not a Windows user and I have never tinkered with bat files. The change should be simple to apply, but I don't know the syntax.

PR #4264, opened 2024-05-16
allow more relative pathes

Shumpei-Tanaka:
it allows to set relative path in below
- "latex-workshop.latex.texDirs"
- "latex-workshop.intellisense.package.dirs"

args: rootFile and tmpDir set as "".
so it  allows only WORKSPACE_FOLDER and RELATIVE_DIR

This change is useful to use .sty which user made.

And if setting is below:
"latex-workshop.intellisense.package.dirs" = %WorkspaseFolder%/styles
It also allow completions (.json) which user made.

So, this change will make user's workspace more potable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1092,7 +1092,7 @@
           "scope": "resource",
           "type": "boolean",
           "default": false,
-          "markdownDescription": "Whether to include the name of the root file being built in the status bar. You can use only 2 placeholders: `%WORKSPACE_FOLDER%`, `%RELATIVE_DIR%`. Please visit https://github.com/James-Yu/LaTeX-Workshop/wiki/Compile#placeholders for a complete list of all placeholders."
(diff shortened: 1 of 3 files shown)


James-Yu:
Thanks for your contribution. Please also amend the descriptions to these two config items and state the usage of placeholders.

Shumpei-Tanaka:
setting example is below:

```json
    "latex-workshop.latex.texDirs": "%WORKSPACE_FOLDER%/styles",
    "latex-workshop.intellisense.package.dirs": "%WORKSPACE_FOLDER%/styles",
```

`%RELATIVE_DIR%` also works.

James-Yu:
I mean the descriptions in `package.json`. They will show when users editing the config items.

Shumpei-Tanaka:
thanks for reply.
I pushed package.json which added variable info.

Issue #4281, opened 2024-06-13
Seamless and flicker-free pdf reload


quoc-ho:
## Pre-checks
- [x] The feature request has not been suggested in this repository.

## The Missed
After each compilation, the PDF viewer reloads the PDF *with a flicker*. Namely, instead of displaying the content of the new PDF seamlessly, a blank page briefly appears before being replaced by the new content.

It would be great to remove the flicker.

## The Solution*
The PDF viewer of [Overleaf-Workshop](https://github.com/iamhyc/Overleaf-Workshop) is also built on top of pdf.js but doesn't flicker when reloading the PDF. I think the main difference between what [they do](https://github.com/iamhyc/Overleaf-Workshop/blob/master/src/core/pdfViewEditorProvider.ts) and LaTeX-Workshop is that when reloading, pdf.js object is not recreated (though I'm not familiar with the code base of LaTeX-Workshop enough to say). Moreover, they also read the content of the PDF file into a byte array before handing it to pdf.js, which might have reduced the latency.

## Anything Else?
The amount of code related to PDF viewing in Overleaf-Workshop is quite small. So probably/hopefully, the amount of work needed to implement this in LaTeX-Workshop is also small. I tried to look into this myself, but I found the PDF viewing code in LaTeX-Workshop a bit complicated for me to do it quickly since I'm not very familiar with the codebase. 

Even if the maintainers of LaTeX-Workshop decide that it's not worth the trouble, having this available here might help others who want to do this themselves.

Thank you!

quoc-ho:
I wanted to add that the popular [vscode-pdf](https://marketplace.visualstudio.com/items?itemName=tomoki1207.pdf) also does not flicker when the pdf reloads.

James-Yu:
Both are using `PDFViewerApplication.load(doc)`, while we are using `PDFViewerApplication.open(url)`.

However, I have been playing with the API for a while. Switching to `load()` seems not resolving the flicking in this extension. As this request is primarily aesthetic and seems need indeed much work. I will leave it here and hopefully there will be community contribution.

quoc-ho:
Very interesting. And thanks for leaving it here.

Is there a particular reason why `PDFViewerApplication.open(url)` rather than `PDFViewerApplication.load(doc)` is used in LaTeX-Workshop? Also, the snippet I copied above does a little more than just calling `PDFViewerApplication.load(doc)`. However, you still think that it wouldn't do anything for our extension?

[...]

PR #3754, opened 2023-03-02
CSS-based page trimmer

James-Yu:
The previous page trimming logic is a mess. It was based on not-easy-to-read calculations in javascript, where the performance is also a problem.

This PR re-write the feature by switching to a CSS-styling solution to perform trimming.

Merging this PR will also fix #3749 .

@bwegge I invite you to have a try on the artifact at https://github.com/James-Yu/LaTeX-Workshop/actions/runs/4311627924/jobs/7521222688 and see if the horizontal scrolling issue is resolved and the general trimming functions as intended.

PR #4292, opened 2024-06-20
Better viewer trimming

James-Yu:
This PR provides a new implementation to replace the original PDF viewer trimming logic.

The old one heavily relies on Javascript computation on the size/margin of viewer, and the implementation is quite inter-twined.

This new implementation instead is CSS-based. We create a new css-var called `--trim-factor`, which will be multiplexed onto the `--scale-factor` of PDF.js for high-resolution rendering of PDF.

Further, the viewer relies on `--trim-factor` to calculate the width/height of each page, then set a negative margin to the top and left to centralize each page, effectively trimming the margins.

This PR also moves the trimming selector from main toolbar to the dropdown menu: this decision is based on the small size of empty space in the main toolbar. The previous place often caused layout glitches.

Further, the trimming selector is improved to be a numerical input, such that users have a more precise control over the trimming percentage.

Further +1, the structure of `latexworkshop.ts` is changed by extracting features in other `component` files. Clean up a bit.


James-Yu:
The someday in #3754 has come. @bwegge hello! I'm not sure if you are still using this extension. If so, may I invite you to have a test on this PR? The latest artifact (as of now) can be found at https://github.com/James-Yu/LaTeX-Workshop/actions/runs/9595844297 .

James-Yu:
I was quite confident on the new implementation until I revised #3754 and noticed so many edge but practical usages 🤣

James-Yu:
@quoc-ho The latest commit should work well with pages of different sizes.

quoc-ho:
There's some scrolling issue with pdf reload when trimming level is changed. Steps to reproduce:
- Open a tex file, compile, and open preview.
- Scroll to some page > 1
- Increase the trim level
- Make some small changes in the tex file and recompile

After each compile, the scrolling position will move up. This issue will disappear if the pdfviewer is closed and reopened.

James-Yu:
> After each compile, the scrolling position will move up. This issue will disappear if the pdfviewer is closed and reopened.

Caused by a racing condition on setting the `scrollTop` of viewer container. Fixed in ec788b8.

quoc-ho:
The newest commit makes things flicker much more than before because the first page is displayed briefly before jumping to the original location. This happens even without trimming enabled. With #4281 incorporated, the flicker is still there (but slightly less). This is a new source of flicker.

quoc-ho:
The newest commit doesn't fix the introduction of more flicker. Moreover, the previous problem with scrolling position reappears.

quoc-ho:
Unfortunately, the extra flicker is still there (where the first page is briefly displayed) with trimming on. Moreover, the only zoom level that works with trimming is `Automatic Zoom`. Changing the zoom level reintroduces the bug with the scrolling position and moreover, the zoom level gets changed after a reload.

James-Yu:
Flickering is not the top priority issue to be handled in this PR. I am first working on the appearance of trimming.

Regarding zoom level, does trimming work on fit-width and fit-page? Besides auto, these two should be the only mode trimming work, as all others are the absolute percentage zooming, which trimming should not change the scale.

Further, I've tested fit-page and 200% zooming, no offset in reloading pdf found. May I know which zoom level and browser/in-editor viewer you are using?

James-Yu:
cc8dbc3 should fix the zoom-level-reset problem. I am yet to reproduce the scrolling offset issue.

One more tip, you can use `Refresh all LaTeX PDF viewers` command to quickly reload pdfs, without needing to re-compile.

James-Yu:
I figure (again) that the PDF.js viewer is too convoluted to make serious modifications. I unfortunately give up again.

To future me: Unless you have two weeks' free time (8h per day), don't touch trimming. Or, don't touch anything related to PDF.js scaling.

James-Yu:
The latest commit should be working in any cases (hopefully).

The jump-to-page-1-then-back was indeed not a regression. Original implementation also behave as is. I will look into improving that after the new trimming is polished.

quoc-ho:
Everything seems to work, even when incorporating my newest patch for the flicker! 🎉🎉🎉🎉🎉

quoc-ho:
I celebrated too early. The issue with losing the scroll position happens when there's a landscape page in the middle. Also, sometimes, the portrait pages are aligned centrally compared to the landscape page and sometimes they are aligned on the left, seemingly depending on where the landscape page is in the viewport. And I know that it's not the main issue here, but maybe it's related: when there are landscape pages, the scrolling to top problem is more visible.

https://github.com/James-Yu/LaTeX-Workshop/assets/10014608/0b719d14-caa2-43f7-97e8-68c4e709ab50

quoc-ho:
On this branch, Page-width/page-fit depends on the current page rather than the largest page in the document whereas on the master branch, things behave differently. So this seems to be a minor issue having to do with computing the right number.

James-Yu:
> The issue with losing the scroll position happens when there's a landscape page in the middle.

On it.

> Also, sometimes, the portrait pages are aligned centrally compared to the landscape page and sometimes they are aligned on the left, seemingly depending on where the landscape page is in the viewport.

That was really a good catch. However, this is a pdf.js viewer issue. Same thing happens with main branch, and I confirmed on the viewer.mjs code and figured that the height and width of pages are calculated based on the current page. So the scale is not fixed for such pdfs. I don't think it's possible to fix it on the extension side though.

> And I know that it's not the main issue here, but maybe it's related: when there are landscape pages, the scrolling to top problem is more visible.

On it soon.

> Page-width/page-fit depends on the current page rather than the largest page in the document.

pdf.js viewer issue. L12614 and L12180-L12181 to be specific.

quoc-ho:
The behavior is still different from the one in the master branch, and I think the one in the master branch is more desirable. Namely, in the master branch, the zoom level (in the case of page width/fit) might change on refresh but it only changes at most once. On the current branch, it changes whenever the current page switches from a landscape to a portrait page (or vice versa). So, a small scroll can mix everything up.

It might be the case that the behavior in the current branch is the default one for pdfjs while the one on the master branch happens to be better :).

James-Yu:
This new implementation of trimming should be in a good shape to merge. Un-appeared bugs will be fixed.

Now will turn to 1) clean up `latexworkshop.ts` and remove/re-implement server-client communication code, 2) reduce flickering.

PR #4301, opened 2024-06-25
L10n Language Support

akenejie:
Please forgive my English ability.
Could you please consider making the extension multilingual for those who use VSCode in non-English environments?
There are some parts of them that are not yet Japanese-enabled, but I have created the basic part of the multilingualization for now.


James-Yu:
Thank you for your contribution. Regarding i18n, I may need to think of a more robust framework to support more languages. I will return to this PR with more concrete ideas in mind.

James-Yu:
@akenejie Thank you for your contribution. In f2072e8, I set up a unified localization (l10n) framework. May you kindly help on the following two things?
1. I've set up the default English language translation in `package.nls.json`. May you follow the pattern and translate to `package.nls.jp.json`?
2. I've reworked the translation of `snippetview.html`, but only two or three lines. May you create placeholders in the pattern `{%snippetview.somekey%}` in the file, and record the new keys in `package.nls.json` as well as `package.nls.jp.json`? You may take a look at `{%snippetview.symbol%}`, `{%snippetview.tikz%}`, `{%snippetview.search%}` and see how it works.

Many thanks!

PR #4308, opened 2024-07-01
Live share integration

quoc-ho:
Continue the discussion https://github.com/James-Yu/LaTeX-Workshop/discussions/4305#discussioncomment-9927467.


hoang06kx1:
Awesome! Great contribution @quoc-ho . Let's wait for review.

quoc-ho:
The following functionalities have been implemented on guests (on the host, everything is the same as before):
- View pdf by clicking on the pdf file
- View pdf by the view command*
- Auto refresh pdf view when the pdf file changes
- Inverse and forward* synctex

Here, * means doesn't work when the guest is on a Windows computer. This is because on Windows, `getPdfPath` always returns smth with `c:\` at the beginning, which does not make sense. For guests, the input will be something like `vsls:/file.tex` regardless of the system. I only have limited access to a Windows computer so at least for now, I cannot really troubleshoot this. For now, Windows users can double click on the pdf file to open it instead. Forward synctex is still a problem.

-----

**Notes on the implementation**
1. On the host, everything should function exactly the same as before.
2. There are two minor but very frequent changes involving URI schemes:
        a. Previously, `vscode.Uri.file` was used, which has `file` scheme and which doesn't work on the guest side since the appropriate scheme there should be `vsls`. I implemented `fileUriFromPath` in `file.ts` which handles schemes more appropriately.
        b. Previously, there were many checks `.scheme === 'file'`. I implemented `isAcceptedScheme` which allows `vsls` as well and replaced the checks by using this function.
3. On the guest, the HTTP server now switches to proxy mode during a live share session which pipes the requests to the host via the shared port (the share is facilitated by the live share extension). The guest also opens a WebSocket connection to the host WebSocket server and uses this to handle refresh and synctex.
4. When starting a new session, the host will automatically share its port via the live share extension and the guest will automatically obtain that port when joining. Since live share doesn't expose an API to acquire the port (on the guest), it's done via triggering the appropriate command and accessing the clipboard. If the initial sharing fails for some reason (for example, when the host doesn't click allow for the share request), the user can initiate share (on the host) and acquire port (on the guest) via two new commands.

James-Yu:
I'm back. On it in the next week(s).

package.json
```diff
@@ -285,6 +285,16 @@
...
+      {
+        "command": "latex-workshop.acquireHostPort",
+        "title": "Acquire Live Share host port",
+        "category": "LaTeX Workshop"
+      },
+      {
+        "command": "latex-workshop.shareHostPort",
+        "title": "Share Live Share host port",
+        "category": "LaTeX Workshop"
+      },
```
James-Yu, line 297: The title of commands should follow the recently introduced i10n naming.

src/core/commands.ts
```diff
@@ -5,6 +5,26 @@ import { getSurroundingMacroRange, stripText } from '../utils/utils'
...
+export async function shareHostPort() {
+    logger.log('SHAREHOSTPORT command invoked.')
+    if (lw.liveshare.isHost) {
+        await lw.liveshare.shareServer()
+    }
+    else {
+        void vscode.window.showInformationMessage('This command only works for a host in a Live Share session.')
+    }
+}
+
```
James-Yu, line 27: I'm thinking of merging these two commands to one but haven't come up with a good name for it.

src/extras/liveshare.ts
```diff
@@ -0,0 +1,86 @@
+import * as vsls from 'vsls/vscode'
+import * as vscode from 'vscode'
+import { lw } from '../lw'
+import * as url from 'url'
+import * as utils from '../utils/utils'
+
+export class LiveShare {
```
James-Yu, line 7: Current design of the extension avoid using classes to implement features. You may refer to other modules in the `extra\` folder to see the implementation. Generally we rely on `export` to expose functions to LW.

src/locate/synctex.ts
James-Yu, line 1: Can you briefly introduce the edits to this file, especially the new APIs exposed to the extension? It seems at a glance that they are not related to Live Share?

src/preview/host-connection.ts
James-Yu, line 1: For better maintainability, documentations for each function may be required.

src/preview/server.ts
```diff
@@ -184,7 +196,52 @@ function sendOkResponse(response: http.ServerResponse, content: Buffer, contentT
     response.end(content)
 }
 
+async function proxyHandler(req: http.IncomingMessage, res: http.ServerResponse) {
```
James-Yu, line 199: We follow a minimal intrusion principle for the extra features. It means that functions like this one should be put into `liveshare.ts` and add minimal number of changed lines to other files.

src/preview/viewer.ts
```diff
@@ -53,26 +53,34 @@ function reload(): void {
...
+    const pdfUri = pdfFile ? lw.file.getUri(pdfFile) : undefined
     if (pdfUri === undefined) {
         manager.getClients()?.forEach(client => {
-            client.send({type: 'refresh'})
+            client.send({type: 'refresh', pdfFileUri: client.pdfFileUri})
         })
         return
     }
-    const clientSet = manager.getClients(pdfUri)
+    let clientSet = manager.getClients(pdfUri)
```
James-Yu, line 63: Similarly, minimal intrusion.

src/preview/viewer.ts
```diff
@@ -297,9 +309,51 @@ function handler(websocket: ws, msg: string): void {
         }
         case 'reverse_synctex': {
             const uri = vscode.Uri.parse(data.pdfFileUri, true)
+            if (lw.liveshare.isGuest) {
```
James-Yu, line 312: Similarly, minimal intrusion.

James-Yu (changes requested):
I did not check carefully (next step will do so), yet it seems that currently synctex from guest to host then back to guest is adding a lot of lines to the code, which is making the code quite convoluted.

I would recommend removing all host-guest communications related to synctex. For guest, rely on its local optionally-installed `synctex` and fallback to the bundled `synctexjs` implementation. This will make the code much more maintainable.

src/completion/latex.ts
```diff
@@ -57,7 +57,7 @@ export class Provider implements vscode.CompletionItemProvider {
     async resolveCompletionItem(item: vscode.CompletionItem, ctoken: vscode.CancellationToken): Promise<vscode.CompletionItem> {
         const configuration = vscode.workspace.getConfiguration('latex-workshop')
         if (item.kind === vscode.CompletionItemKind.Reference) {
-            if (!('file' in item) || !configuration.get('hover.ref.enabled')) {
+            if (!('file' in item || 'vsls' in item) || !configuration.get('hover.ref.enabled')) {
```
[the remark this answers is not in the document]
quoc-ho: Hover doesn't work correctly inside LiveShare. The purpose of this check is to make sure that it runs correctly even if the file under consideration has `vsls`  scheme (i.e., in guest in LiveShare mode). But it turned out that LiveShare itself has some problems with hovering in general. Namely, all hovering code seems to run on host for some reason. I tried to remove all changes wrt. hover but seems like this one escaped my attention.

src/locate/synctex.ts
quoc-ho: All the changes to this file are simply refactors. Originally, the various functions in this file had both pure and side effect code. So, I refactored these functions to separate the pure and the side effect parts. The reason is that when the guest requests synctex, the host simply calls the pure part and returns the values. So, even though the number of changes is large, nothing is actually changed.

src/core/commands.ts
quoc-ho: How about "Share/Acquire port"?

quoc-ho:
Thanks for looking into this. I think the changes are quite minimal. The large changes in `synctex.ts` are just refactoring. I didn't use the local synctex or synctex.js because the `tar.gz` file is also unavailable on the guest side due to binary file limitation imposed by LiveShare.

I'm traveling for quite a while so I will be available only for comments if you have questions but probably not actually writing code.

James-Yu:
I've done some refactoring to follow minimal intrusion. May you have a check if the new code is equivalent when available?

Further, as there are some limited functionalities on Windows, it would be good to have sort of notices when related functions are invoked on win.

James-Yu (approved):

quoc-ho:
Great! I'll go over the refactor by Tuesday.

quoc-ho:
Sorry for letting this slip. I'm super swamped right now, unfortunately. I'll try to do it soon.

krokosik:
I would love to see this feature merged. Could I help with the refactor in any way? @quoc-ho @James-Yu

krokosik:
@quoc-ho Sure, I'll take a swing at it over the weekend. Do you want me to make another fork or PR or could you add me as a collaborator?

quoc-ho:
@krokosik Great! I added you as a collaborator.

krokosik:
@James-Yu I have resolved merge conflicts and added a workaround for Windows. Apparently LiveShare works halfway as a remote VSCode environment, as it uses POSIX paths for its URI scheme, but on the other hand uses the local OS APIs. This is why `path.resolve` introduces the `C:` prefix, while the parsed `vscode.Uri` later converts that to POSIX.

I also added some small comments in `liveshare` and tested that both PDF refreshing and SyncTex still work. I tried different platform configurations, between Windows, WSL and a DevContainer. Also I noticed the `.devcontainer` directory still uses Node 16, which does not even allow to build the package anymore, so I updated it to the same version as used in CI. Hope that such a small scope extension is ok with you, but I can also make a separate PR for that.

Thank you so much for creating this extension and I would like to express my willingness to help with merging this as soon as possible. 🙌

James-Yu:
@krokosik Great work! Before adding docs, please bear me with a few days to check the current implementation. As the changes are too hard to review in its current form, I would propose two things for us to complete:
1. I will make a wrapper function of vscode.Uri.file in file.ts, so that we only need to change this function to include vsls file scheme, reducing the change pollution. This will be reflected in `master`.
2. May you merge this change to the branch and edit the new file.ts function to cope with vsls file scheme?

After that, I can proceed to review the real changes.

Further, I would suppose there may be an influx of issues regarding liveshare. Would you mind go on maintaining the feature in the following few weeks/months, until it becomes stable? I am very happy to invite new repository collaborators if you are also interested in any other parts of the extension 👍

James-Yu:
Here it is: bf03458

I do think the original `getUriScheme` is unnecessary: `toUri('').scheme` should work perfectly.

Further, `isUriScheme` can be replaced by a new constant in lw.ts and use a `.include(scheme)` check.

viewer/components/utils.ts
```diff
@@ -42,7 +42,7 @@ export function parseURL(): { encodedPath: string, pdfFileUri: string, docTitle:
 
     for (let i = 0, ii = parts.length; i < ii; ++i) {
         const param = parts[i].split('=')
-        if (param[0].toLowerCase() === 'file') {
+        if (param[0].toLowerCase() === 'file' || param[0].toLowerCase() === 'vsls') {
```
krokosik, line 45: I'm not sure how to proceed here. It would be nice to have only one constant, like the one in `lw.ts`, but I guess `utils` should import close to nothing. This place does not seem like the right place for such a constant, however, the `pdfFilePrefix` is also duplicated in `lw.ts` so I decided to also do it.

James-Yu:
I've (roughly) checked the changes and they look great! May you help document the changed functions in `synctex.ts` and more importantly, maybe some small paragraphs at the top of `extra/liveshare.ts` explaining the internals of this feature? After that, the final thing might be a small section in https://github.com/James-Yu/LaTeX-Workshop/wiki/ExtraFeatures and a link to it introducing how to use, requirements and limitations.

krokosik:
@James-Yu Thank you! I have added the comments and submitted a PR for the wiki:
https://github.com/jlelong/LaTeX-Workshop-wiki/pull/125

quoc-ho:
That sounds great. But maybe except for #4305 as it would be quite circular? 😅

James-Yu:
Well done! Let users' voice come!

krokosik:
Second that! This project is truly remarkable in how well it is maintained and developed. Thank you @James-Yu and of course @quoc-ho for doing the Live Share heavy lifting!

PR #4343, opened 2024-08-23
Update LaTeX3 completion to l3kernel 2024-07-20

muzimuzhi:
Script is also updated. Now the path to l3kernel dtx can be set through `PATH_TO_DTX` envvar, which makes the script usable to a wider range of users. The existence of the resolved dtx path is also ensured.

cc @mbertucci47


dev/latex3command.py
```diff
@@ -14,7 +15,14 @@
 COMMANDS_FILE = CWD.joinpath('../data/commands.json').resolve()
 ENVS_FILE = CWD.joinpath('../data/environments.json').resolve()
 OUT_DIR = CWD.joinpath('../data/packages').resolve()
-dtx_files = Path('/usr/local/texlive/2023/texmf-dist/source/latex/l3kernel/').glob('*.dtx')
+
+PATH_TO_DTX = os.environ.get(
+    "PATH_TO_DTX",
+    default='/usr/local/texlive/2024/texmf-dist/source/latex/l3kernel/')
```
muzimuzhi, line 21: I need this setup because I don't have package sources installed but do have a local clone of the LaTeX3 repository.

It may also help non-TeX Live users or TeX Live users using a custom installation directory. For example the path for basicTeX users (including me), if exists, is
```
/usr/local/texlive/2024basic/texmf-dist/source/latex/l3kernel/
```

The default path to dtx can also be generalized to
`$(kpsewhich --var-value TEXMFDIST)/source/latex/l3kernel/`.

Issue #4349, opened 2024-08-27
Math preview for empheq environment


fk128:
> _Please fill the following fields with a star (*) and provide as much related information as possible._

## Pre-checks*
> _Please change the following `[ ]` to `[x]` for confirmation._
- [x] The feature request has not been suggested in this repository.

## The Missed*
> _Is your feature request related to a problem? Please provide a clear and concise description of what the problem is._
- I want to be able to see math previews for equations defined inside `empheq` env. E.g. the following does not generate a math preview:

```tex
\documentclass[]{article}
\usepackage{empheq}

\begin{document}
\begin{empheq}[box=\fbox]{align}
    A & = B \\
    C &= D
\end{empheq}
\end{document}
```

## The Solution*
> _Please provide a solution you would like to have._

The solution seems to be simply adding `empheq` to `ENV_NAMES` https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/parse/find.ts#L6 

One issue, however, is that `\begin{empheq}[box=\fbox]{align}` would result in `Error: Invalid option: box` whereas `\begin{empheq}[]{align}` would work to generate a preview. Perhaps it would be possible to get mathjax to ignore the options when it's an empheq env or strip the options out in the mathjaxify function https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/preview/hover/utils.ts#L8? I can make a PR for this, if this sounds sensible.

## Anything Else?
> _Add any other context about the feature request below._
-

James-Yu:
https://docs.mathjax.org/en/v3.2-latest/input/tex/extensions/empheq.html

We generally won't alter tex string passed to MathJax other than new command definitions. Making ENV_NAMES in MathJaX finding would be fine, and I am happy to review PR on that.

PR #4351, opened 2024-08-27
Enable preview for empheq env

fk128:
Related to #4349 .

From https://docs.mathjax.org/en/v3.2-latest/input/tex/extensions/empheq.html :

> Note that the current implementation of the empheq environment supports only the left and right options.

MathJax fails when the `box` option is used, i.e. when doing

```
\begin{empheq}[box=\fbox]{align}
    A & = B \\
    C &= D
\end{empheq}
```

but not 

```
\begin{empheq}[]{align}
    A & = B \\
    C &= D
\end{empheq}
```

 It's not clear from the MathJax docs whether it would be possible to just ignore the option rather than fail to generate a preview.

### DIFF
--- a/src/preview/hover/utils.ts
+++ b/src/preview/hover/utils.ts
@@ -15,6 +15,10 @@ export function mathjaxify(tex: string, envname: string, opt = { stripLabel: tru
     if (envname.match(/^(aligned|alignedat|array|Bmatrix|bmatrix|cases|CD|gathered|matrix|pmatrix|smallmatrix|split|subarray|Vmatrix|vmatrix)$/)) {
         s = '\\begin{equation}' + s + '\\end{equation}'
     }
+    if (envname == 'empheq') {
+        const empheqPat = /\\begin\{empheq\}\[.*?\]\{(\w+)\}/g;
+        s = s.replace(empheqPat, '\\begin{empheq}[]{$1}');
+    }
     return s
 }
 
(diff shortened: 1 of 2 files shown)


src/preview/hover/utils.ts
```diff
@@ -15,6 +15,10 @@ export function mathjaxify(tex: string, envname: string, opt = { stripLabel: tru
     if (envname.match(/^(aligned|alignedat|array|Bmatrix|bmatrix|cases|CD|gathered|matrix|pmatrix|smallmatrix|split|subarray|Vmatrix|vmatrix)$/)) {
         s = '\\begin{equation}' + s + '\\end{equation}'
     }
+    if (envname == 'empheq') {
+        const empheqPat = /\\begin\{empheq\}\[.*?\]\{(\w+)\}/g;
+        s = s.replace(empheqPat, '\\begin{empheq}[]{$1}');
+    }
```
James-Yu, line 21: We won't do special treatments to any particular macro. This need to be removed. A proper fix should be on MathJax side.

James-Yu: I see.

The current implementation won't work properly for highly-ill formats like `\begin{empheq}[box=[\fbox]{test}]{align}`. You may need to use `getLongestBalancedString()` utility function in `src/utils/utils.ts` to extract the real m-arg.

James-Yu:
Close for inactivity.

PR #4384, opened 2024-09-13
feat: add check citations command to see unused cites

buehler:
Hi there :-)

I really like using your extension and thought to add a new command
for BibTeX files. It would be very cool to have this inserted into the
workshop extension.

This PR adds a new command to the action-bar under the bibtex
category. It allows people to use the "checkcites" command line tool
that is shipped with recent Tex installations. The user then sees
a list of unused citations and can directly jump to them.

As such, you can cleanup your citations at the end of your paper.

I know that I still need to provide translations. Do I need to do them for all
languages? Because I'll not be able to do that.

I'll open this PR as a draft until I know how to proceed.


James-Yu:
Thank you for your contribution. If I understand correctly, this feature is somewhat of the same kind with `lw.extra.counter` right? These features should reside in `src/extra` and have minimal code injections in other source files, in order to be maintainable.

I agree that files in `extra/` are not quite minimally intrusive at their current state. However, new standalone features need to be as "sneaky" as possible.

buehler:
fair enough :) I'll rearrange stuff according to your guidelines

buehler:
Is it ok, if the command stays inside the bibtex section in the action bar? It is related to bibliography and as such it makes sense to keep it there.

I change the files to be in extra.

James-Yu:
Thanks for the changes. There is one minor thing before the next step: there are unnecessary changes in `linter.ts`. Please revert.

In the meantime, I am still considering how to better arrange the `extra/` features. It is more an architectural design, and any final thought won't require you to make additional changes. Just maybe some more time before I confirm my thoughts and merge.

buehler:
I just checked, linter.ts has no changes in it (according to the changed files tab)

buehler:
ah, I get it, you mean `src/lint/index.ts`

buehler:
Reverted the changes.

James-Yu:
Sorry for the way too long wait. Merge time!

buehler:
No worries!
Thanks! :-)

Happy to see the thing in action.

PR #4406, opened 2024-09-22
tectonic with --print to generate compile message for lw parse

Xiao-Chenguang:
The original latex tool config for tectonic missing the "--print" flag for tectonic to output compile message to console.
Without this, lw failed to parse the compile log and pass the problems like **overfullbox** to vscode.
With the `--print` flag, tectonic functions same as latexmk in lw.
This would benefit tectonic users.


James-Yu:
Please add the final line break.

Xiao-Chenguang:
> Please add the final line break.

Now it's back.

Issue #4415, opened 2024-10-01
PDF viewer turns blank after refreshing when spreading is enabled


LinqLover:
## Pre-checks*
> _Please change the following `[ ]` to `[x]` for confirmation._
- [ ] ~~The issue has not been reported in this repository.~~ **No, it has been mentioned: #4300 (closed as not reproducible) and https://github.com/James-Yu/LaTeX-Workshop/pull/4295#issuecomment-2188891983 (issue there was apparently fixed)**
- [ ] The issue remains after disabling all other extensions and restarting Visual Studio Code.
- [x] The [FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) cannot address the issue.
- [x] The issue is not related to compiling a document, or the document can be successfully compiled in the OS terminal but not in Visual Studio Code with this extension.

## Environment*
> _Please write exact version numbers instead of descriptors such as `latest`._
- Operating System: Linux x64 6.5.0-45-generic (Ubuntu 22)
- Visual Studio Code Version: 1.93.1
- LaTeX Workshop Version: v10.4.1
- TeX Distribution Version: TeX Live 2022

## The Issue*
> _Please briefly describe the issue you come across._
- PDF viewer (in browser or in VS Code tab) becomes blank after refreshing document when spreading (Odd Spreads/Even Spreads) is enabled

### Reproduction Steps
> _Please list out the steps to reproduce your bug. Include relevant environmental variables or any other configuration._
1. MWE:
   ```tex
   % !TeX root = test.tex
   \documentclass[twoside]{scrbook}
   \usepackage{blindtext}
   \begin{document}
   \Blinddocument
   \end{document}
   ```
2. Save
3. Command `latex-workshop.view` or `latex-workshop.viewInBrowser`
4. In the menu at the right top corner, select "Odd Spreads" or "Even Spreads"
5. Command `latex-workshop.refresh-viewer`
6. Do some scrolling in the PDF viewer
7. Now, the viewer "crashes": it turns blanks and the scroll bar is lost. F5 in the browser helps but then the viewer state (scroll position, view mode, etc.) is lost. See errors below.

### Expected Behavior
> _What were you expecting to see? Include any relevant examples or documentation links._
- The file should just be updated

### LaTeX Workshop Output*
> _Please paste the whole log messages below, not parts of ones. The log should start with `New log placeholder %WS1% registered`._
```
[...]

PR #4423, opened 2024-10-03
Flicker free refresh

James-Yu:
This is a re-implementation of #4295 , following the discussion from https://github.com/James-Yu/LaTeX-Workshop/issues/4415#issuecomment-2390389131, and more importantly, migrating (reads: copy-and-paste) codes from https://github.com/tamuratak/latex-toybox/

In our previous implementation, we inject into PDF.js viewer refreshing logic to prevent flickering during PDF refresh. It caused a few issues, for example, spreading mode bounding box computation as in #4415 .

@tamuratak proposed a very elegant and non-intrusive solution for the problem in his maintaining branch by first creating images of visible PDF pages and masking them on top of canvas, then removing them once all pages are rendered. We "shamelessly" copy the idea and code.

@quoc-ho may you take a look and have a test on this branch and see if there are issues to fix? Many thanks.

### DIFF
--- a/dev/editviewer.py
+++ b/dev/editviewer.py
@@ -1,5 +1,4 @@
 import os
-import re
 import argparse
 
 parser = argparse.ArgumentParser()
@@ -22,9 +21,6 @@ with open(args.web + '/viewer.mjs', 'rt', encoding='utf-8') as fin:
     with open(args.viewer + '/viewer.mjs', 'wt', encoding='utf-8') as fout:
         currentClass = ''
         for line in fin:
-            r = re.match(r'class (.*?) \{', line)
-            if r:
-                currentClass = r[1]
             line = line.replace('''const MATCH_SCROLL_OFFSET_TOP = -50;''', '''const MATCH_SCROLL_OFFSET_TOP = -100;''') \
                 .replace('''this.switchView(view, true);''', '''this.switchView(view, false);''') \
                 .replace('''console.warn(`[fluent] Missing translations in ${locale}: ${ids}`);''', '''// console.warn(`[fluent] Missing translations in ${locale}: ${ids}`);''') \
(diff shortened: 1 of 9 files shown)


quoc-ho:
Wow. So fast! Thanks!!! 

There's one problem though: when trimming is on, the mask has the wrong dimension.


https://github.com/user-attachments/assets/6127f3ef-6d9c-4264-9c2e-e81bce19e950

James-Yu:
Yeah I forgot that we are using CSS-based trimming. This latest commit should fix the problem 🤣

quoc-ho:
Thanks! That fixes the previous problem. But on some pages, the mask is off by a couple of pixels. The problem seems to appear (mostly) on every other pages. See the attached video.


https://github.com/user-attachments/assets/a3c8902d-5cdb-4e1b-8398-994a759c2eb5

James-Yu:
Done!

It turns out that the offset values can only be integer values, but the center position of dom can be sub-pixel. In the latest commit, we use `getBoundingClientRect()` to obtain the floating values of the position and size of dom.

quoc-ho:
Unfortunately, when we scroll all the way to the top, then the position of DOM is calculated correctly. But when we are in the middle of a page, then it's wrong. See video below.


https://github.com/user-attachments/assets/8d45f5e5-85ee-445a-b3bc-ad672f8eaa90

James-Yu:
This ... should work? Checked a few use cases.

I noticed one issue: when spreading is on, auto-width seems to be a bit wider than should be. It's not related to this PR, yet I will work on it later.

quoc-ho:
Very awesome!

One last thing (hopefully): if you scroll all the way down and then refresh pdf, the position of the mask is wrong, resulting in the view being shifted briefly. This doesn't happen when the scroll position is not completely bottom.


https://github.com/user-attachments/assets/f712d522-7fc3-4367-be70-efeda0526154

James-Yu:
Nice catch! This should be caused by pdfjs randomly destroying the last page and add it back with a small period when the page bottom margin of 10px is not added. The latest commit adds an invisible anchor div to cover the margin space.

Indeed, there is a better but behavior-changing solution: remove the 10px bottom margin of the last page, so that there is no margin in the top and at the bottom. If so, there's no need to use this anchor workaround. However, I have not settled the thought on the margin change.

quoc-ho:
The experience is so smooth! Best pdf viewing experience ever! 

I noticed one issue: when spread mode is on and the tab being refreshed is not visible, synctex fails after a couple of refreshes. This does not happen when spread mode isn't on. The error (from the webview developer tools) is

```
Uncaught TypeError: Cannot read properties of undefined (reading 'getPagePoint')
    at callSynctex (synctex.js:32:65)
    at pageDom.onclick (synctex.js:46:21)
```

I checked the DOM but the mask seems to be correctly removed already. So not sure what the cause is.

James-Yu:
That's strange. Anyways this commit should prevent the error log.

I think the new implementation should be ready. Good to push to users.

quoc-ho:
I want to clarify that reverse synctex (pdf -> tex) was the problem. It's still a problem, at least for me. I don't use spread mode so I'm not affected by it. But just so you know. :)

PR #4426, opened 2024-10-06
Support code folding for DocTeX

huangyxi:
This pull request enhances the LaTeX-Workshop extension by adding support for VSCode code folding for DocTeX files. The new functionality includes:

1. Sections Folding: Adds support for folding LaTeX section commands in DocTeX, such as `% \section{abc}`.
2. DocStrip Guards Folding: Adds support for folding DocStrip guard ranges with patterns like `%<*abc>` to `%</abc>`, where `abc` can be replaced with characters such as `|, -, _, (, )`.

Showcase for the `l3doc` class:
image

### DIFF
--- a/src/language/folding.ts
+++ b/src/language/folding.ts
@@ -2,10 +2,17 @@ import * as vscode from 'vscode'
 
 export class FoldingProvider implements vscode.FoldingRangeProvider {
     private readonly sectionRegex: RegExp[] = []
+    private readonly doctex: boolean
 
-    constructor() {
+    constructor(doctex: boolean = false) {
         const sections = vscode.workspace.getConfiguration('latex-workshop').get('view.outline.sections') as string[]
-        this.sectionRegex = sections.map(section => RegExp(`\\\\(?:${section})(?:\\*)?(?:\\[[^\\[\\]\\{\\}]*\\])?{(.*)}`, 'm'))
+        this.doctex = doctex
+        if (this.doctex) {
+            this.sectionRegex = sections.map(section => RegExp(`%\\s*\\\\(?:${section})(?:\\*)?(?:\\[[^\\[\\]\\{\\}]*\\])?{(.*)}`, 'm'))
+        }
+        else {
+            this.sectionRegex = sections.map(section => RegExp(`\\\\(?:${section})(?:\\*)?(?:\\[[^\\[\\]\\{\\}]*\\])?{(.*)}`, 'm'))
+        }
     }
 
     public provideFoldingRanges(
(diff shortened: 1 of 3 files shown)


James-Yu:
Thanks for your contribution. Please extend a new doctex folding class for the implementation. It would be fine if you need to make `envRegex` a class instance variable.

huangyxi:
Thank you for your response. I have updated the structure of the new DocTeX folding class, and added a new folding pattern that matches from `% \iffalse meta-comment` to `% \fi`.

huangyxi:
I think it is ready for another review.

src/language/folding.ts
```diff
@@ -105,6 +114,12 @@ export class FoldingProvider implements vscode.FoldingRangeProvider {
                 keyword = 'group'
             } else if (match[7] || match[8]) {
                 keyword = 'region'
+            } else if (match[9]) {
+                keyword = '%<' + match[9] + '>'
+            } else if (match[10]) {
+                keyword = '%<' + match[10] + '>'
+            } else if (match[11] || match[12]) {
+                keyword = '%\\iffalse meta-comment'
```
James-Yu, line 122: They are only for DocTeX

src/language/folding.ts
```diff
@@ -2,10 +2,11 @@ import * as vscode from 'vscode'
 
 export class FoldingProvider implements vscode.FoldingRangeProvider {
     private readonly sectionRegex: RegExp[] = []
+    protected readonly envRegex = /\\(begin){(.*?)}|\\(begingroup)[%\s\\]|\\(end){(.*?)}|\\(endgroup)[%\s\\]|^%\s*#?([rR]egion)|^%\s*#?([eE]ndregion)|^%\s*<\*([|_()\-a-zA-Z0-9]+)>|^%\s*<\/([|_()\-a-zA-Z0-9]+)>|^%\s*\\iffalse\s*(meta-comment)|^%\s*\\(fi)/gm
```
James-Yu, line 5: Seems you add new matches to this regexp. Are they only for DocTeX?

src/language/folding.ts
```diff
@@ -113,11 +128,12 @@ export class FoldingProvider implements vscode.FoldingRangeProvider {
             }
             const lastItem = opStack[opStack.length - 1]
 
-            if ((match[4] || match[6] || match[8]) && lastItem && lastItem.keyword === item.keyword) { // match 'end' with its 'begin'
+            if ((match[4] || match[6] || match[8] || match[10] || match[12]) && lastItem && lastItem.keyword === item.keyword) { // match 'end' with its 'begin'
```
James-Yu, line 131: Same as above

James-Yu (commented):
The original `FoldingProvider` class should not contain codes related to DocTeX.

huangyxi:
The `getEnvironmentFoldingRanges` method has been separated into both classes.

Issue #4457, opened 2024-11-11
Compiling files with extensions other than `.tex`


zhengxinhao-comp:
## Environment
- Operating System: Windows 11
- Visual Studio Code Version: 1.95.2
- LaTeX Workshop Version: 10.5.6
- TeX Distribution Version: TeX Live 2024

## The Issue
LaTeX Workshop refuses to compile source file with file extension other than `.tex`.

### Reproduction Steps
I use a music typesetter called [LilyPond](https://lilypond.org/). The way it corporates with LaTeX is that I input the LaTeX source code with LilyPond code in the environment `\begin{lilypond}` `\end{lilypond}`, and save the file with the extension `.lytex`, and use the command `lilypond-book` to compile it. `lilypond-book` would compile the `.lytex` file to a `.tex` file, then the user can use a regular LaTeX command to compile the intermediate file to PDF.

I've set the following tool
```json
        {
            "name": "lilypond-book",
            "command": "lilypond-book",
            "args": [
                "%DOC_EXT%"
            ],
            "env": {}
        }
```
and recipe
```json
        {
            "name": "Compile lytex",
            "tools": [
                "lilypond-book",
                "xelatexmk"
            ]
        }
```
to automate the task but I can't get the compilation running when I click the recipe in LaTeX panel.

### Expected Behavior
LaTeX Workshop should be able to compile source file with extensions other than ".tex". I don't know if it's possible, or it would be better to have an option in the settings defining file extensions LaTeX Workshop recognizes and is able to compile.

James-Yu:
Compiling files other than `tex` is out of scope of this extension. You may resort to vscode tasks for the job.

jlelong:
It reminds me of what I implemented to support literate programming for the R and Julia languages. 
@James-Yu If you are fine with it, I can try to make a PR to support to `.lytex`. From my point of view, it is not so much a matter of compiling the file as there are alternative approaches but rather to make other features such as intellisense work.

The first step is to closely look at the structure of a `.lytex` file to determine if it is feasible with little modification. @zhengxinhao-comp could you provide a representative example of such a file?

zhengxinhao-comp:
> It reminds me of what I implemented to support literate programming for the R and Julia languages. @James-Yu If you are fine with it, I can try to make a PR to support to `.lytex`. From my point of view, it is not so much a matter of compiling the file as there are alternative approaches but rather to make other features such as intellisense work.
> 
[...]

PR #4466, opened 2024-11-21
Handle LaTeX files with non `.tex` extension

jlelong:
Related to #4457 

This PR adds a new configuration variable to let the user specify file extensions to be treated as LaTeX files similarly to `.tex` files.

@James-Yu What do you think of this very little intrusive approach?


James-Yu:
I’m good with the change, just one thing: it seems that changing the config cannot change the respective behavior on-the-fly, until a vscode reload. May be we move the new ext var to file.ts, and register an onConfigChange handler to change it lively?

jlelong:
It is not completely clear to me because allowing to change the value on-the-fly will only work for adding more extensions. Files with extensions that were cached will not be removed from the caching mechanism. So I am a bit hesitating.

James-Yu:
That's mostly for consistency. iirc changes on all other configs are now reflected immediately without the need of reload. Further, I ignored the impact on caches when changing the config. It might be more complex than I thought.

jlelong:
Fine. I completely agree with the consistency argument. I will make the appropriate changes.

jlelong:
Thanks for the tests. 
I still need to think of it a little as currently building will not work properly because of `langId` computation.

jlelong:
@James-Yu Everything seems fine to me now. What do you think?

PR #4532, opened 2025-02-15
Correct translation for the word "recipe" in "zh-cn" and "zh-tw"

zeyu10:
In language [zh-cn] and [zh-tw], this extension translates "recipe" as "配方", which is not accurate.

I suggest correct the translation for "recipe" as "配置" or other Chinese both in [zh-cn] and [zh-tw].

Hope consider my suggestion. With regards.


James-Yu:
The translation is intentional: 配置 is configuration or setting, 配方 is recipe, a concept introduced by LW.

PR #4572, opened 2025-04-10
\left \right bracket snippets

lukketotte:
Maybe too specific, but suggestion of adding snippets for 

- \left( \right)
- \left[ \right]
- \left\{ \right\}


lukketotte:
Thanks for pointing that out, had missed it! 

Considered them to be somewhat superfluous in one sense, but figured I could check with you since they are frequently used and quite intense in terms of strain on the hands (angled brackets particularly) :)

PR #4600, opened 2025-05-31
Fix `*.bib` resolving (glossaries and bib2gls)

hellmrf:
This PR introduces a minor change to the cache behavior, so that `core.cache.updateGlossaryBibFiles` behaves similarly to `core.cache.updateBibfiles` in the sense of respect `kpsewhich`.

Currently, `core.cache.updateGlossaryBibFiles` is defined as:

https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/core/cache.ts#L536-L549

Which is similar to `core.cache.updateBibfiles`:

https://github.com/James-Yu/LaTeX-Workshop/blob/master/src/core/cache.ts#L499-L513

This PR updates `core.cache.updateGlossaryBibFiles` to use the same mechanism as `core.cache.updateBibfiles` to load `bibPaths`, calling `lw.file.getBibPath`.

This is aligned with `bib2gls`' own mechanism:

https://github.com/nlct/bib2gls/blob/master/src/lib/resources/bib2gls-en.xml#L14-L18

### DIFF
--- a/src/core/cache.ts
+++ b/src/core/cache.ts
@@ -541,15 +541,17 @@ async function updateGlossaryBibFiles(fileCache: FileCache) {
         const bibs = (result[1] ? result[1] : result[2]).split(',').map(bib => bib.trim())
 
         for (const bib of bibs) {
-            const bibPath = await utils.resolveFile([path.dirname(fileCache.filePath)], bib, '.bib')
-            if (!bibPath || isExcluded(bibPath)) {
-                continue
-            }
-            fileCache.glossarybibfiles.add(bibPath)
-            logger.log(`Glossary bib ${bibPath} from ${fileCache.filePath} .`)
-            const bibUri = lw.file.toUri(bibPath)
-            if (!lw.watcher.glossary.has(bibUri)) {
-                lw.watcher.glossary.add(bibUri)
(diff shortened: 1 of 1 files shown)


jlelong:
Thanks for the PR. I have a question, which would similarly apply to `updateBibfiles`. 

For a given bib entry, we may resolve several full paths and watch all of them. Isn't it a bit too aggressive?  In the end the latex/bibtex/bib2gls compiler will only use the first resolved file.

hellmrf:
> For a given bib entry, we may resolve several full paths and watch all of them. Isn't it a bit too aggressive? In the end the latex/bibtex/bib2gls compiler will only use the first resolved file.

I'm not very familiar with `lw.file` internals and I confess I didn't fully understand why `lw.file.getBibPath` returns an array. I would expect a single `string`/`object`.

For example, I have a `B:/example/assets/bib/glossary/symbols.bib` and I'm at `B:/example`. and  If I search for `symbols.bib`

```pwsh
PS B:\example> kpsewhich symbols.bib

PS B:\example> $Env:BIBINPUTS = "B:/example/assets/bib//"

PS B:\example> kpsewhich symbols.bib
B:/example/assets/bib/glossary/symbols.bib
```

PR #4688, opened 2025-10-17
Mouse wheel-driven page turning in viewer when Page Scrolling is enabled

nilsjor:
## Overview
This PR adds a very small change to viewer.mjs (`onWheel`) so the mouse wheel (vertical or horizontal) will turn the page when there is no content to scroll in that wheel direction.

## Motivation
Users expect the wheel to behave similar to PgDn/PgUp for single-page viewing: if there is nowhere to scroll on the page, the mouse wheel should advance/rewind the document.

## Behavior changes
- New: In Page Scrolling mode (`scrollMode=3`), when a wheel event has no scrollable content in its dominant axis:
  - Scrolling down/right → next page
  - Scrolling up/left → previous page
- Native behavior preserved:
  - If there is content to scroll in the wheel direction, the event is not intercepted and the browser handles scrolling.
  - Ctrl/Meta/pinch zoom handling is unchanged and still takes precedence.
  - "Presentation mode" (whatever that is) is unaffected.

## Implementation
- File: [viewer.mjs](viewer.mjs)
- Function: `onWheel(evt)`
  - Keep the existing zoom branch intact and return early after handling zoom.
  - Detect dominant axis (horizontal vs vertical; support Shift+wheel).
  - If the axis is scrollable, return to allow native scrolling; otherwise do nothing (conservative: no flip when zoomed in).

## Backward compatibility
- Minimal, low-risk change:
  - No new settings or flags.
  - Existing zoom and scrolling semantics preserved except for the added wheel→page action when there is no scrollable content.
  - Presentation mode and other keyboard shortcuts remain unchanged.

## How to test
1. Choose zoom mode "Page Fit" or smaller:
   - Wheel down/right → next page immediately.
   - Wheel up/left → previous page immediately.
2. Zoomed in (page larger than viewport):
   - Wheel scrolls within the page normally; no page flip while there is content to scroll.

## Follow-ups:
I have also developed a more comprehensive, "edge-aware" Page Scrolling mode, which might be preferrable, let me know what you think. If so, I can move that content onto this branch:
- When fully zoomed out, let the user change pages with the mouse wheel—just like this PR.
- When zoomed in, flipping to the previous/next page is triggered when the viewport is at the top/bottom (or left/right) edge and the user keeps “pushing” in that direction.
- Additionally, Up/Down/Space and PageUp/PageDown keys perform a similar behavior, allowing for flipping pages when reaching the end of the scroll bar.

### DIFF
--- a/viewer/viewer.mjs
+++ b/viewer/viewer.mjs
@@ -14845,6 +14845,42 @@ function onWheel(evt) {
       }
       this.updateZoom(ticks, null, origin);
     }
+    return;
+  }
+
+  // Minimal page-turn-on-wheel behavior: 
+  // If there is remaining content to scroll in the wheel direction, let native scrolling occur, otherwise turn page.
+  if (pdfViewer.scrollMode === ScrollMode.PAGE) {
+    const container = pdfViewer.container;
+    if (!container) {
+      return;
+    }
+
+    const absX = Math.abs(evt.deltaX || 0);
+    const absY = Math.abs(evt.deltaY || 0);
+    const useHorizontal = absX > absY || (evt.shiftKey && absY > 0);
+
+    let goNext = false;
+    if (useHorizontal) {
+      const goingRight = (evt.deltaX || (evt.shiftKey ? evt.deltaY : 0)) > 0;
+      if (container.scrollWidth > container.clientWidth + 1) {
+        return; // allow native horizontal scrolling
+      }
+      goNext = goingRight; // right => next, left => previous
+    } else {
+      const goingDown = (evt.deltaY || 0) > 0;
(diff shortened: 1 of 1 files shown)


James-Yu:
Thank you for your contribution. That is a nice addition. However, we don't directly edit `viewer.mjs`; we use a script https://github.com/James-Yu/LaTeX-Workshop/blob/master/dev/editviewer.py to auto edit it. Thus the changes can persist across PDF.js upgrades.

Further, I do think this contribution should be directed to https://github.com/mozilla/pdf.js as it may benefit a broader user base. I would suggest you to file a PR there.

PR #4708, opened 2025-11-04
Improve Checkcites can read the output dir

sukiElaina:
fix: edit auxFile can read from latex.outDir

### DIFF
--- a/src/extras/checkcites.ts
+++ b/src/extras/checkcites.ts
@@ -8,17 +8,12 @@ const logger = lw.log('Citations', 'Linter')
 export function checkCitations() {
   logger.log('Checking citations.')
   const rootPath = lw.root.file.path
-  if (!rootPath) {
-    logger.log('No root file found.')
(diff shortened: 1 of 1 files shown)


src/extras/checkcites.ts
```diff
@@ -1,12 +1,17 @@
...
 import { lw } from '../lw'
+import * as path from 'path'
 
 const logger = lw.log('Citations', 'Linter')
 
 export function checkCitations() {
   logger.log('Checking citations.')
-
-  const auxFile = lw.root.file.path?.replace(/\.tex$/, '.aux')
+  const configuration = vscode.workspace.getConfiguration('latex-workshop', lw.file.toUri(lw.root.file.path!))
```
James-Yu: Should not use force operator here.

sukiElaina (commented):
FIx the bug that checkcite command cannot read the outdir of config

Issue #3045, opened 2022-01-08
Build error when working in path containing symbolic link to a mounted drive


constantin-huetterer:
Thank you for writing this awesome extension! It looks great and I can't wait to use it. :slightly_smiling_face: 

__Edit:__ After some investigation, I am not sure if this is a bug from the extension or a VS Code Bug / security feature. Please close if you find that this bug doesn't belong here. The workaround of just opening the project where it really is works well.

### Disable all the other extensions except for LaTeX Workshop, restart VS Code, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes, I didn't find anything related to this problem.

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: Yes, running `latexmk ... %DOC% " manually works

## Describe the bug [Required]

When opening a workspace inside of a folder containing a symbolic link, bibtex reports an error, making the build fail and displaying an error message on each build.

### To Reproduce

Setup:
1. Create symbolic link (in my case it was a mounted drive)
2. Open Tex workspace in a sub-folder through the symbolic link
3. Try to build using latexmk recipe

My setup looks like this:
```
/home/user/docs/ --> symbolic link that points to /mnt/drive/docs
```
- When I open the workspace `/home/user/docs/project/` in VS Code, builds fail
- In contrast to opening `/mnt/drive/docs/project` in VS Code, then builds succeed
- If I run the latexmk command in the VS-Code internal console, however, it succeeds no matter which version of the workspace I had opened.

### Expected behavior

Builds succeed in sub directories of symbolic links as well.

### LaTeX Workshop Output [Required]
<details>
<summary> Workshop Log</summary>

```
[15:19:43] Initializing LaTeX Workshop.
[15:19:43] Extension root: /home/redacted/.vscode/extensions/james-yu.latex-workshop-8.23.0
[15:19:43] $PATH: redacted
[15:19:43] $SHELL: /bin/bash
[15:19:43] vscode.env.appName: Visual Studio Code
[15:19:43] vscode.env.remoteName: undefined
[15:19:43] vscode.env.uiKind: 1
[15:19:43] editor.acceptSuggestionOnEnter: "on"
[15:19:43] latex-workshop.bind.enter.key: true
[...]

Issue #3023, opened 2021-12-26
BibTeX doesn't work with custom `outDir` (like `%TMPDIR%`)


maltejur:
### Disable all the other extensions except for LaTeX Workshop, restart VS Code, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: Yes

### To Reproduce
With the two files

`main.tex`
```tex
\documentclass{report}

\begin{document}
\cite{a-reference}

\bibliographystyle{plain}
\bibliography{bibliography}
\end{document}
```
`bibliography.bib`
```bib
@misc{a-reference,
  title  = {A Reference},
  author = {Leandro Facchinetti},
  year   = {2019}
}
```

Make sure first the setting `latex-workshop.latex.outDir` is set to `%DIR%`.

Clicking on `Build LaTex project` in the LaTeX tab should work without any errors.

Then change `latex-workshop.latex.outDir` to `%TMPDIR%`.

Click on `Build LaTex project` again and get an error.

### Expected behavior

I would expect everything would compile the same with `%TMPDIR%` as with `%DIR%`.

### LaTeX Workshop Output [Required]

```
[14:31:45] Initializing LaTeX Workshop.
[14:31:45] Extension root: /home/maltejur/.vscode/extensions/james-yu.latex-workshop-8.22.0
[14:31:45] $PATH: /home/maltejur/repos/kinopub_downloader/bin:/home/maltejur/repos/winapps/bin:/home/maltejur/repos/scripts:/home/maltejur/repos/pandoc-watch:/home/maltejur/repos/kinocli/dist:/home/maltejur/scripts:/home/maltejur/Programme/kinocli/dist:/home/maltejur/Programme/audio-share-discord-linux/bin:/home/maltejur/Android/Sdk/platform-tools:/home/maltejur/Android/Sdk/cmdline-tools/latest/bin:/home/maltejur/.yarn/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl
[14:31:45] $SHELL: /bin/fish
[14:31:45] vscode.env.appName: Visual Studio Code
[...]

Issue #2303, opened 2020-09-27
latexmk fails with a bibtex error "Not writing to .blg (openout_any = p)"


emersonmello:
### Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

And I found this [issue here](https://github.com/James-Yu/LaTeX-Workshop/issues/186)

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: Yes, using the line below.

```bash
latexmk -pdf -synctex=1 -interaction=nonstopmode -file-line-error  main.tex 
```

## Describe the bug [Required]

I can't compile a simple .tex + .bib using ` latexmk recipe` and  LaTeX Workshop default settings. But, I'm able to compile using 
latexmk on terminal and I'm able to compile using pdflatex -> bibtex -> pdflatex x2 recipe.

### To Reproduce

1. Use Latex Workshop default settings and recipes
2. Create a minimal .tex file
```latex
\documentclass[a4paper,12pt]{article}
\usepackage{lipsum}

\begin{document}
\section{Introduction}
\lipsum[1-3]

\nocite{*}
\bibliographystyle{plain}
\bibliography{references.bib}

\end{document}
```
3. Create a minimal .bib file
```bibtex
@book{goossens94,
  title={{The LATEX companion}},
  author={Goossens, M. and Mittelbach, F. and Samarin, A.},
  year={1994},
  publisher={Addison-Wesley}
}

@book{lamport94,
  title={{La T E X: A Document Preparation System}},
  author={Lamport, L. and others},
  year={1994},
  publisher={Addison-Wesley, Reading, Massachusetts,}
}
```
4. Click on Recipe: latexmk
5. See error dialog pop-up 

### Expected behavior

A pdf file with the references.

## Logs [Required]

Please paste the whole log messages here, not parts of ones. The log should start with `Initializing LaTeX Workshop`. It is very important to identify problems.

### LaTeX Workshop Output [Required]

```
[19:19:19] Initializing LaTeX Workshop.
[19:19:19] Extension root: /Users/user/.vscode/extensions/james-yu.latex-workshop-8.13.2
[19:19:19] Creating PDF file watcher.
[...]

Issue #2042, opened 2020-04-17
bibtex couldn't open .aux file 


happyTonakai:
### Disable all the other extensions except for LaTeX Workshop, and check that you still see this issue. [Required]

You still see this issue?: Yes

### Make sure to visit the [wiki FAQ](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ) before filling an issue.

You visited the wiki?: Yes

### If your issue is with compiling a document (not having to do with finding the root file of a project), check first that you can compile manually.

You can compile a TeX document manually?: Yes

## Describe the bug [Required]
After I added `--output-directory`  option to xelatex, bibtex reported an error saying `I couldn't open file name 'template.aux'`.

Output directory is `./Tmp`
```
 "latex-workshop.latex.outDir": "%DIR%/Tmp",
```
And my recipe is as follows
```
        {
            "name": "xelatex ➞ bibtex ➞ xelatex`×2",
            "tools": [
                "xelatex",
                "bibtex",
                "xelatex",
                "xelatex"
            ]
        }
```
` latex-workshop.latex.tools` are
```
        {
            "name": "xelatex",
            "command": "xelatex",
            "args": [
                "-synctex=1",
                "-interaction=nonstopmode",
                "-file-line-error",
                "--output-directory=%OUTDIR%",
                "%DOC%"
            ],
            "env": {}
        },
        {
            "name": "bibtex",
            "command": "bibtex",
            "args": [
                "%DOCFILE%"
            ],
            "env": {}
        }
```

When I set  `"latex-workshop.latex.outDir": "%DIR%"`, everything works fine. 

I think the problem is that bibtex couldn't find the `.aux` file in the root directory. How do I set the arguments of `bibtex`? I don't see any placeholder that meets my need.

## Logs [Required]

Please paste the whole log messages here, not parts of ones. It is very important to identify problems.

### LaTeX Workshop Output [Required]

<details>
<summary>LaTeX Workshop extension log</summary>

```
[12:52:23] Initializing LaTeX Workshop.
[12:52:23] Creating file watcher for .pdf files.
[12:52:24] Creating LaTeX Workshop http and websocket server.
[12:52:24] LaTeX Workshop initialized.
[12:52:24] Found root file from active editor: c:\Users\Tonakai\Desktop\test\template.tex
[12:52:24] Root file changed from: undefined to c:\Users\Tonakai\Desktop\test\template.tex. Find all dependencies.
[...]

PR #4714, opened 2025-11-06
improve outdir for bibtex

wanstarge:
Although Even though a solution was [mentioned](https://github.com/gallandarakhneorg/autolatex/issues/83) many years ago , if users configure `outDir` as `%WS1%/build`, the BibTeX parameters must then be configured as `%OUTDIR%/%DOCFILE%`. At this point, the parameters expand to `%WS1%/build/%DOCFILE%`, and `%WS1%` is, in turn, expanded into an absolute path.

Calling the BibTeX recipe at this stage will result in an error. This is because BibTeX, under the default security settings of TeXLive and MacTex (`openout_any = p`), is not allowed to write files (such as .blg/.bbl) to an absolute path outside the working directory. Many users will, sooner or later, encounter this `openout_any = p` problem. https://github.com/James-Yu/LaTeX-Workshop/issues/3045 https://github.com/James-Yu/LaTeX-Workshop/issues/3023 https://github.com/James-Yu/LaTeX-Workshop/issues/2303 https://github.com/James-Yu/LaTeX-Workshop/issues/2042 https://github.com/James-Yu/LaTeX-Workshop/issues/1932 https://github.com/James-Yu/LaTeX-Workshop/issues/384

To address this, I am attempting to solve the issue with a minimal change, which is to add the variable `env['openout_any']='a'
Thanks for your review again.
```
"latex-workshop.latex.outDir": "%DIR%/build",
  {
    "name": "bibtex",
    "command": "bibtex",
    "args": ["%OUTDIR%/%DOCFILE%"]
  }
]
```

img

### DIFF
--- a/src/compile/build.ts
+++ b/src/compile/build.ts
@@ -186,6 +186,36 @@ async function buildLoop() {
     isBuilding = false
     setTimeout(() => lw.compile.compiledPDFWriting--, vscode.workspace.getConfiguration('latex-workshop').get('latex.watch.pdf.delay') as number * 2)
 }
+/** Normalizes a command-line argument that represents a file path to be
+ * relative to the current working directory (`cwd`) if it is under the root
+ * directory (`rootDir`). If the argument does not represent a path or is not
+ * under the root directory, it is returned unchanged.
+ *
+ * @param {string} arg - The command-line argument to normalize.
+ * @param {string} cwd - The current working directory.
+ * @param {string} rootDir - The root directory of the LaTeX project.
+ * @returns {string} - The normalized command-line argument.
+ */
+function normalizeArgForCwd(arg: string, cwd: string, rootDir: string): string {
+    if (!arg) { return arg }
+    let abs: string
+    try {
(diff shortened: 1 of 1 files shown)


src/compile/build.ts
```diff
@@ -236,6 +236,10 @@ function spawnProcess(step: Step): ProcessEnv {
         if (step.command === 'latexmk' && step.rootFile === lw.root.subfiles.path && lw.root.dir.path) {
             cwd = lw.root.dir.path
         }
+        if(step.command==='bibtex')
+        {
+            env['openout_any'] = 'a'
+        }
```
jlelong: Even though I understand the point of adding this, it cannot be done without warning the users about the potential security breach. I am against merging such a change, but it could documented in the wiki or even in the _Discussions_ tab.

wanstarge:
Have you see the latest change, I have change the way to solve this question, they can only access the file in it's workspace

jlelong:
The new approach looks better, although the function `normalizeArgForCwd` seems a bit complicated. I think the file argument must be the last entry of `step.args`. This should simplify the function quite a bit.

I am reopening this PR in the hope that we can come up a little intrusive solution.

PR #4738, opened 2025-11-20
feat: Add LaTeX formatting provider that converts math delimiters

JuliusBairaktaris:
This PR implements a new formatting feature that automatically converts TeX-style math delimiters to their LaTeX equivalents, addressing ChkTeX Warning 46 ("Use \( ... \) instead of $ ... $").

The feature performs the following conversions:
- Inline math: `$ ... $` -> `\( ... \)`
- Display math: `$$ ... $$` -> `\[ ... \]`

## Key Features
- **Configurable:** Controlled by the new setting `latex-workshop.format.fixMath.enabled` (default: `false`).
- **Safe:**
    - Preserves escaped dollar signs (`\$`).
    - Respects verbatim environments (`\begin{verbatim}`, `\verb`, `\lstlisting`, etc.) and does not modify content within them.
    - Handles multi-line math environments correctly.

## Changes
- **New:** `src/extras/math-fixer.ts` - Implements the `MathFixer` class (logic similar to the existing `QuoteFixer`).
- **Modified:** `src/lint/latex-formatter.ts` - Integrates `fixMath` into the formatting pipeline.
- **Modified:** `package.json` - Adds the configuration option.

**Doc to verify:**
```
\documentclass{article}
\usepackage{listings}
\usepackage{fancyvrb}

\begin{document}

% ==========================================
% SECTION 1: Basic Replacement (Target Cases)
% ==========================================

% 1. Simple Inline Math
% Expected: Becomes \( a^2 + b^2 = c^2 \)
Let $ a^2 + b^2 = c^2 $ be the Pythagorean theorem.

% 2. Multiple Inline Math on one line
% Expected: Becomes \( x > 0 \) and \( y < 0 \)
Assume $ x > 0 $ and $ y < 0 $.

% 3. Basic Display Math
% Expected: Becomes \[ E = mc^2 \]
The formula is:
$$ E = mc^2 $$

% 4. Multi-line Display Math
% Expected: Becomes \[ ... \] enclosing the block
$$
    \int_{a}^{b} f(x) dx = F(b) - F(a)
$$

% ==========================================
% SECTION 2: Escaped Characters (Should NOT change)
% ==========================================

% 1. Escaped Dollar
% Expected: Should remain \$100.00
The cost is \$100.00 exactly.

% 2. Escaped Dollar inside Math (Tricky edge case)
% The code regex `(\\.)` captures escaped chars first, so this should remain \$ inside the math delimiters.
% Expected: \( \{ x \mid x \textit{ costs } \$5 \} \)
The set is $ \{ x \mid x \textit{ costs } \$5 \} $.

% ==========================================
% SECTION 3: Verbatim Environments (Should NOT change)
% ==========================================

% 1. Standard Verbatim
% Expected: Delimiters inside remain $ ... $ and $$ ... $$
\begin{verbatim}
    This is code:
    var x = $value$;
    var y = $$value$$;
\end{verbatim}

% 2. Starred Verbatim
% Expected: Remain $ ... $
\begin{verbatim*}
    Spaces are visible here $ but math is ignored $
\end{verbatim*}

% 3. lstlisting (from listings package)
% Expected: Remain $ ... $
\begin{lstlisting}
    // This is a comment with $ math syntax $
    int main() { return 0; }
\end{lstlisting}

% 4. Capitalized Verbatim (fancyvrb)
% Expected: Remain $ ... $
\begin{Verbatim}
    Text inside Verbatim $ should stay $
\end{Verbatim}

% ==========================================
% SECTION 4: Inline Verbs (Should NOT change)
% ==========================================

% 1. Standard \verb with different delimiters
% Expected: Remain $ ... $
Use the command \verb|$HOME| to see the var.
Use delimiters \verb+$+ to enter math mode.

% 2. \Verb (Capitalized)
% Expected: Remain $ ... $
\Verb+$math$+

% ==========================================
% SECTION 5: Comments (Should NOT change)
% ==========================================

% The code specifically checks: if (line.trimStart().startsWith('%'))
% Expected: The following line should NOT change
% $ This is a comment with dollars $

% Note: The current code implementation likely DOES convert comments 
% if they appear in the middle of a line (e.g. "text % $comment$"), 
% but ignores full line comments.

\end{document}
```

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -40,7 +40,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v5
+      uses: actions/checkout@v6
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
     - uses: actions/setup-node@v6
       with:
         node-version: '18'
--- a/.github/workflows/texlive_on_linux.yml
+++ b/.github/workflows/texlive_on_linux.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
     - name: Set up PATH
       run: |
         echo "/tmp/texlive/bin/x86_64-linux" >> $GITHUB_PATH
--- a/.github/workflows/texlive_on_mac.yml
+++ b/.github/workflows/texlive_on_mac.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: macos-15
     timeout-minutes: 30
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
     - name: Set up PATH
       run: |
         echo "/Users/runner/texlive/bin/x86_64-darwin" >> $GITHUB_PATH
--- a/.github/workflows/texlive_on_win.yml
+++ b/.github/workflows/texlive_on_win.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
     - name: Prepare git
       run: git config --global core.autocrlf false
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
     - name: Set up PATH
       run: |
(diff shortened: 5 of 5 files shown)


src/utils/utils.ts
```diff
@@ -79,9 +79,26 @@ export function stripComments(text: string): string {
...
-    const envsAlt = envs.join('|')
-    const pattern = `\\\\begin{(${envsAlt})}.*?\\\\end{\\1}`
-    const reg = RegExp(pattern, 'gms')
+    if (envs.length === 0) {
+        return text
+    }
+
+    // Build alternation of environment names, each with optional star
+    // Use non-capturing group for the star to keep backreference simple
+    const envPatterns = envs.map(env => `${env}(?:\\*?)`).join('|')
```
jlelong: No need of a group, just use `\\*?`.

jlelong (changes requested):

PR #4750, opened 2025-12-04
Add a setting to decide from which directory to run the recipes

jlelong:
Related to #4747 

This PR adds the new setting `latex-workshop.latex.build.fromWorkspaceFolder` to decide if the recipe is run from the toplevel workspace folder or from the root file directory.


jlelong:
@James-Yu I am wondering why the build test suite is disabled

https://github.com/James-Yu/LaTeX-Workshop/blob/1975b8d39f03ce274e1b4ffde62cbcfcd9821f37/test/suites/01_build.test.ts#L8

James-Yu:
> @James-Yu I am wondering why the build test suite is disabled
> 
> https://github.com/James-Yu/LaTeX-Workshop/blob/1975b8d39f03ce274e1b4ffde62cbcfcd9821f37/test/suites/01_build.test.ts#L8

That was an intended skip. The new unit tests are (or are supposed to) cover all test cases in our old test suites, so the latter is disabled. I have not taken a look into the tests for a while. Todo list +1.

Issue #4762, opened 2025-12-23
Incorrect color of scrollbar in dark mode PDF preview window (SOLUTION provided)


JoeYe-233:
> Unfortunately, this cannot be achieved without side effect. The viewer page and scrollbar has to be synced in terms of css filter. So a dark theme inverted PDF can only have a light scrollbar. I cannot come up with workarounds. 

 _Originally posted by @James-Yu in [#4649](https://github.com/James-Yu/LaTeX-Workshop/issues/4649#issuecomment-3484146742)_

It is actually possible to achieve this, since the *invert filter* only applies to the portion of PDF preview **below toolbar** (i.e. , the **viewerContainer** part), we can force this *portion* to use light theme, while keeping the rest as dark theme.

The dark mode process at present, which produces this unwanted "light color scrollbar", can be described as: VSCode dark mode -> Built-in webview follows this dark mode (result: toolbar dark; pages LIGHT; scrollbar dark) -> invert filter applies to *viewerContainer* (result: toolbar dark; pages dark; scrollbar LIGHT).

---
The key solution is to add a line `color-scheme: light !important;` to the `#viewerContainer` part of the `viewer.css` file, located at Row 6074 of `%UserProfile%\.vscode\extensions\james-yu.latex-workshop-10.12.0\viewer\viewer.css` After editing, the file should look like this: 

```css
#viewerContainer{
  overflow:auto;
  position:absolute;
  inset:var(--toolbar-height) 0 0;
  outline:none;
  z-index:0;
  color-scheme: light !important;
}
```

Now the whole process becomes: VSCode dark mode -> Built-in webview follows this dark mode, while keeping the *viewerContainer* part in light mode (result: toolbar dark; pages LIGHT; scrollbar LIGHT) -> invert filter applies to *viewerContainer* (result: toolbar dark; pages dark; scrollbar dark), which elegantly fixes this problem. :)

 Screenshots for reference:
<img width="327" height="345" alt="Image" src="https://github.com/user-attachments/assets/b7d285e5-ef85-462d-b085-10dc14b48fc6" />

<img width="350" height="159" alt="Image" src="https://github.com/user-attachments/assets/dade8f76-3471-48f5-98b1-13018dcf5ae5" />

PR #4763, opened 2025-12-23
Fix scrollbar color inconsistency in dark mode

JoeYe-233:
#4762


Harmonese:
Yes, we definitely need this

Issue #1825, opened 2019-11-21
Dark background in the pdf viewer


id0158:
Would it be possible to provide an option or (even better) follow the editor theme for the background color in the pdf viewer ?

By _pdf viewer background_ I mean the area around the document and not the background of the document itself.
It's not so impactful for a regular document, but very useful for Beamer (see screenshots below).
Actual settings provide a (very) light background independently of the editor theme and I did not find an option to change this. 

![VSCode](https://user-images.githubusercontent.com/14804964/69366968-a66d5700-0c97-11ea-9fd7-ebf66b31674f.png)
![Firefox](https://user-images.githubusercontent.com/14804964/69366977-aa00de00-0c97-11ea-8c3b-c732de21893e.png)

Firefox pdf viewer background is more consistent with a dark mode, even if for them it's always dark :smile: 

Thank you otherwise for the great work with this extension !

tylerliu1998:
You can invert both the area around the document and the background of the document. Why do you want to invert only one of them?
![image](https://user-images.githubusercontent.com/48684632/69600817-0293fa80-104d-11ea-9b8b-38adb2be3d25.png)

phdoerfler:
I think they want the area of the viewer that is **not** covered by actual pages and content to be coloured dark

tecosaur:
@jlelong do you know if it would be possible to pass the css variables from the outer iframe to the inner iframe (with the PDF viewing method we use), or just use the outer iframe?

That way we could have the pdf viewer match the user's theme, which would be very nice.

jlelong:
Honestly, I do not know if we could achieve this and I am even wondering whether it would be wise to try doing so considering how delicate the viewer may be. Remember the broken links issue #1793 that was cause by badly positioned css.  Detecting the background tone is already quite tricky in mathjax so I do not know how we could do really better here. Suggestions are welcome.

id0158:
Thank you @tylerliu1998 for pointing out this option, I didn't know it exist.
But as @phdoerfler already said, I usually prefer to preserve the original appearance of a document while reducing as much as possible the light around it in the dark mode (at least).
However, `latex-workshop.view.pdf.invert` can be very handy for night editing !

Thank you very much @jlelong for implementing this !
As @tecosaur, I would have preferred that the pdf viewer automatically match the editor theme.

Meanwhile, could it be possible to provide some different predefined background color options for `latex-workshop.view.pdf.tab.backgroundColor`?
[...]

PR #4830, opened 2026-03-24
Change dark mode pdf view background color to dark in settings

mu94-csl:
Change dark mode pdf view background color to darker shade #1b1b1b in the settings, taken from Adobe Acrobat pdf background color in dark mode.

Made the following suggestion permanent: https://github.com/James-Yu/LaTeX-Workshop/issues/1825


James-Yu:
You can set your local config for customization.

PR #4846, opened 2026-04-09
Fix Windows output directory resolution

### LINKED ISSUE #4845: Relative outDir is resolved against the wrong cwd on Windows, and aux/output path mapping is fragile with drive-letter case differences
On Windows, LaTeX Workshop can place build artifacts outside the expected output directory when relative output paths are involved.

After tracing this locally, I found two related path-handling problems:

1. Relative `-outdir` / `-auxdir` arguments can be interpreted against the process cwd instead of the root file directory.
2. Some aux/output path mapping logic relies on string replacement of absolute Windows paths, which breaks when the same path is reported with different drive-letter casing (for example `d:\...` vs `D:\...`).

#### Case 1: relative outDir resolved against the wrong base
1. On Windows, set:
   - `latex-workshop.latex.outDir` to `./output`
   - `latex-workshop.latex.auxDir` to `./output` or `%OUTDIR%`
   - `latex-workshop.latex.build.fromWorkspaceFolder` to `true`
2. Put the root TeX file in a subdirectory of the workspace, for example `samples/sample/t.tex`.
3. Use a recipe that passes `-outdir=%OUTDIR%` / `-auxdir=%AUXDIR%`.
4. Trigger build from VS Code / LaTeX Workshop.

### Expected behavior

A relative outDir should be resolved relative to the root file directory, as documented by the extension settings.

### Actual behavior

The compiler can interpret the relative outDir relative to the workspace cwd instead, so artifacts are written outside the expected output directory.

#### Case 2: Windows drive-letter casing differences
When aux/output paths are mapped back to source directories, absolute path prefix replacement can fail if Windows reports the same path with different drive-letter casing.

This can affect:
- creation of aux subdirectories for included TeX files
- mapping `.aux` files from `.fls` output back to the correct source directory

### Notes

This is separate from user recipes/tools that do not pass any `-outdir` at all. If a selected recipe omits `-outdir`, artifacts next to the cwd are expected. The issue here is about incorrect handling of relative output paths and Windows path normalization in LaTeX Workshop itself.

### Suggested fix direction

- Normalize relative `-outdir` / `-auxdir` arguments against the root file directory before spawning the compiler.
- Replace string-based absolute path prefix substitution with `path.relative()` / `path.resolve()`-based logic for Windows paths.

### Additional context

I verified the problem locally on Windows with TeX Live and added regression tests around:
- relative outDir normalization when building from the workspace folder
- aux/output path handling when Windows path casing differs

aphage:
FIx #4845 
Assisted-by: GitHub Copilot

### DIFF
--- a/src/compile/recipe.ts
+++ b/src/compile/recipe.ts
@@ -67,6 +67,7 @@ export async function build(rootFile: string, langId: string, buildLoop: () => P
 
     // Create build tools based on the recipe system
     const tools = await createBuildTools(rootFile, langId, recipeName)
+    normalizeBuildDirectoryArgs(rootFile, cwd, tools)
 
     // Create output subdirectories for included files
     if (tools?.map(tool => tool.command).includes('latexmk') && rootFile === lw.root.subfiles.path && lw.root.file.path) {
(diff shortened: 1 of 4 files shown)


James-Yu:
https://github.com/James-Yu/LaTeX-Workshop/issues/4845#issuecomment-4211535015

PR #4863, opened 2026-04-21
Fix `querySelector` and SyncTeX keybindings

edemaine:
The root issue I was seeing is that my `latex-workshop.view.pdf.internal.synctex.keybinding` setting was no longer respected. (Ctrl-click worked but double-click did not, despite `"double-click"` setting.) This was caused by `setParams` crashing when `invert` is set.

1. Update viewer DOM selectors to match the current PDF.js markup (`#thumbnailView` → `#thumbnailsView`, `#sidebarContent` → `#viewsManagerContent`)
   * Apparently caused by 9b92b704 Update to PDF.js v5.6.205
2. Add internal-error logs when expected viewer DOM elements are missing, to hopefully prevent this in the future (or at least make it easier to debug).
   * If `throw`ing or something would be better for automated testing purposes, let me know.
   * But note that I wasn't seeing the existing thrown error in logs, so it was difficult to debug. Perhaps both log and throw?
3. Protect access to the queried elements in case they don't exist.
   * This fixes SyncTeX not registering the correct keybinding from config.
   * Could be avoided if we change 2 into a `throw` or something.
4. Deregister click handlers so that reverse SyncTeX handler re-registration always has up-to-date click handlers on page elements
   * Not sure this is important when config changes. Shouldn't hurt, but I can remove it if desired.

### DIFF
--- a/test/units/03_viewer/pdf-viewer.test.ts
+++ b/test/units/03_viewer/pdf-viewer.test.ts
@@ -290,6 +290,22 @@ describe(path.basename(__filename).split('.')[0] + ':', () => {
             assert.strictEqual(execSpy.callCount, 1)
             assert.strictEqual(execSpy.firstCall.args[0], 'workbench.action.moveEditorToLeftGroup')
         })
+
+        it('should not log internal viewer DOM errors while applying params', async () => {
+            set.config('view.pdf.invertMode.enabled', 'always')
+
+            const promise = waitMessage('loaded')
+            await viewInWebviewPanel(pdfUri, 'current', true)
+            await promise
+
+            const internalErrorMessages = handlerSpy.getCalls()
+                .map(call => JSON.parse((call.args?.[1] as Uint8Array)?.toString()) as ClientRequest)
+                .filter((message): message is Extract<ClientRequest, { type: 'add_log' }> => message.type === 'add_log')
+                .map(message => message.message)
+                .filter(message => message.startsWith('Internal error:'))
+
+            assert.deepStrictEqual(internalErrorMessages, [])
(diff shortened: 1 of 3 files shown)


James-Yu:
Nice catch! Can you add a description comment to the change in `viewer/components/synctex.ts`?

edemaine:
> Can you add a description comment to the change in `viewer/components/synctex.ts`?

I revised the in-source comment. Is that what you mean? It's also item 4 above. The idea is that the control-click handler may register, then config loads, then double-click handler registers, so it's maybe better to deregister before reregistering. I don't feel strongly about this though; I don't mind if both control-click and double-click in fact work.

Also fixed the lint bugs - oops.

edemaine:
I found a simple way to test that none of the new `Internal error` messages get logged. Hopefully this helps prevent future regression when pdf.js changes `id`s. I tested that the added test fails without the other changes of this PR.

```
  1) pdf-viewer:
       lw.viewer->viewer.viewInWebviewPanel
         should not log internal viewer DOM errors while applying params:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

+ [
+   'Internal error: #thumbnailView is missing.',
+   'Internal error: #sidebarContent is missing.'
+ ]
- []

      + expected - actual

      -[
      -  "Internal error: #thumbnailView is missing."
      -  "Internal error: #sidebarContent is missing."
      -]
      +[]
```

edemaine:
Any chance this change could be released? I'm unable to use ctrl-clicking because I'm frequently working one-handed.

PR #4869, opened 2026-04-26
Correctly display Latex 'info' as information instead of warning

### LINKED ISSUE #4639: A PackageNote is classified as a warning
I think a PackageNote is rather additional information and should not classified as a warning. This also holds for PackageNoteNoLine, ClassNote, and ClassNoteNoLine. (See https://www.latex-project.org/news/latex2e-news/ltnews34.pdf)

agoeckner:
This PR modifies the Latex log parsing to display 'info' messages as information rather than as a warning in the problem panel.

Here is an example, making use of the `\PackageWarning` and `\PackageNote` macros to push messages into the log at compile time. The notes are then displayed as information in VS Code:
image

This resolves #4639.


agoeckner:
Thank you for fixing the tests @James-Yu! I missed those.

PR #4878, opened 2026-05-01
Fix blurry PDF viewer rendering at fractional DPI

yimchn:
This patch avoids blurry PDF canvas rendering at fractional device pixel ratios, such as Windows 125% display scaling.

PDF.js renders pages into canvases. With fractional DPR values, Chromium may resample the canvas during compositing, making text visibly soft. For fractional DPR values greater than 1, this patch makes PDF.js use an integral backing canvas scale while leaving integer DPR values unchanged.

Tested with:
- npm run compile
- npm run lint

### DIFF
--- a/viewer/latexworkshop.ts
+++ b/viewer/latexworkshop.ts
@@ -10,6 +10,43 @@ import { registerSyncTeX } from './components/synctex.js'
 
 declare const PDFViewerApplication: PDFViewerApplicationType
 declare const PDFViewerApplicationOptions: PDFViewerApplicationOptionsType
+declare const pdfjsLib: {
+    OutputScale?: {
+        readonly pixelRatio: number
+    }
+}
+
+const FRACTIONAL_DPI_EPSILON = 1e-3
+const MIN_FRACTIONAL_DPI_CANVAS_SCALE = 2
+
+// Chromium-based viewers are visibly soft when PDF.js renders page canvases at
+// fractional DPR values such as Windows 125%. Render at an integral backing
+// scale so Chromium does not have to stretch the PDF canvas.
+function getCanvasPixelRatio() {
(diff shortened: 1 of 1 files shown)


James-Yu:
Thanks for your contribution. However, the patch should be applied to upstream pdf.js instead of this repo.

yimchn:
Thanks for the review. I agree that the previous patch was not the right shape: it changed PDF.js' `OutputScale.pixelRatio` from the LaTeX Workshop wrapper, which was only a workaround.

After further investigation, I found that this does not appear to be a PDF.js upstream issue. I compared against the upstream PDF.js viewer/demo and other VS Code PDF viewers that use PDF.js, and they do not reproduce the same blur in this scenario. PDF.js itself already has logic to keep the canvas backing size and CSS display size aligned at fractional device pixel ratios.

The actual issue is in LaTeX Workshop's trimming CSS. The trimming rules generated in `viewer/components/trimming.ts` override PDF.js' page and layer dimensions with `!important`, but they used `--scale-factor` and a fixed `1px` rounding granularity. This bypasses PDF.js' `--total-scale-factor` and `--scale-round-x/y` variables, which are used to keep the CSS display size aligned with the canvas backing size at fractional DPI.

At 125% DPI, PDF.js rounds the page width to a multiple of `4px`, for example `792px`, while the old trimming rule rounded the same value to `793px`. That makes the browser display a `990px` canvas backing store in a `793px` CSS box, producing a non-exact scale ratio and visible interpolation blur.

I replaced this PR with a smaller LaTeX Workshop-side fix in #4879. It only changes the trimming CSS generation so it follows PDF.js' own sizing variables:

- `--total-scale-factor` instead of `--scale-factor`
- `--scale-round-x` / `--scale-round-y` instead of hard-coded `1px`

I will close this PR in favor of #4879.

PR #4908, opened 2026-05-27
Fix build failure when file path contains spaces in shell mode

### LINKED ISSUE #4907: Build fails when file path contains spaces (shell mode with magic comment)
## Bug
When .tex file path contains spaces, building with `% !TEX program = xelatex` fails because the path gets split at spaces in shell mode.

## Root cause
In `src/compile/recipe.ts`, `populateTools()` joins args into a single shell string via `.join(' ')`. When `%DOC%` expands to a path with spaces, the shell splits it.

## Fix
Quote args containing spaces before joining. See PR for details.

kasc0206:
## Summary

Fix #4907

When a `.tex` file is located in a path containing spaces, building with a `% !TEX program` magic comment fails because the expanded path gets split at spaces in shell mode.

## Root Cause

In `src/compile/recipe.ts`, `populateTools()` joins tool arguments into a single shell string when magic comment options are present:

```ts
tool.args = [ `--max-print-line=... ${tool.args.join(" ")}` ]
```

`%DOC%` expands to a path like `/Users/xxx/Library/Mobile Documents/file.tex`. After joining with spaces, the shell interprets the path as two arguments.

## Fix

Arguments containing spaces are now quoted before joining:

```ts
const quoted = tool.args.map((a) => a.includes(" ") ? `"${a}"` : a).join(" ")
tool.args = [ `--max-print-line=... ${quoted}` ]
```

## Testing

- [x] Path with spaces + magic comment: builds successfully
- [x] Path without spaces: builds successfully (regression)
- [x] No magic comment: unaffected (uses array args, not shell)
- [x] Docker mode: unaffected (uses different path format)
- [x] Windows: path with spaces in `%DOC_W32%` also handled

### DIFF
--- a/src/compile/recipe.ts
+++ b/src/compile/recipe.ts
@@ -389,7 +389,9 @@ function populateTools(rootFile: string, buildTools: Tool[]): Tool[] {
             if ((isLaTeXmk || tool.command === 'pdflatex') && isMikTeX()) {
                 if (tool.name === lw.constant.TEX_MAGIC_PROGRAM_NAME) {
                     // %!TeX options is present. All args are provided in a string and { shell: true }
-                    tool.args = [ `--max-print-line=${lw.constant.MAX_PRINT_LINE} ${tool.args.join(' ')}` ]
+                    // Quote arguments containing spaces to prevent path splitting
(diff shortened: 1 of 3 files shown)


kasc0206:
thank you

PR #4912, opened 2026-05-29
Improve Log Parsing for Repeated-Run LaTeX Build Tools

yarakos95:
## Summary

LaTeX has build tools that run multiple times to resolve cross-references and other issues.

LaTeX Workshop supports latexmk, but I want to support other build tools such as [llmk][] and custom-built Make tools as well.

[llmk]: https://ctan.org/pkg/light-latex-make "CTAN: Package Light LaTeX Make"

### Output logs and log analysis by build tools

The build tool we are currently working with outputs a continuous stream of LaTeX logs, so there is no marker to indicate that each LaTeX execution has started, like as latexmk.

Therefore I will add a marker to the log output when LaTeX starts, and use this marker to perform log analysis.

```regexp
/^This\sis\s(pdf|LuaHB|Xe|e-up)TeX,\sVersion/
```

### Log analysis conditions

Add a new property, `parseRepeatedRuns`, to your recipe. This property is not required.

Usage:

```json
"latex-workshop.latex.recipes": [
  {
    "name": "LaTeX",
    "tools": [ "latex build tool" ],
    "parseRepeatedRuns": true
  }
]
```

If this property is set to `true`, analysis will be performed using the LaTeX startup log as a marker. Since it is set to `false` by default, it will not affect existing build recipes.

In addition, since this property can be configured per recipe, you can ensure that log analysis based on this rule is performed only for the recipes that require it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1012,9 +1012,6 @@
                 "items": {
                   "type": "string"
                 }
-              },
-              "parseRepeatedRuns": {
-                "type": "boolean"
               }
             },
             "required": [
--- a/package.json
+++ b/package.json
@@ -1012,6 +1012,9 @@
                 "items": {
                   "type": "string"
                 }
+              },
+              "parseRepeatedRuns": {
+                "type": "boolean"
               }
             },
             "required": [
--- a/src/compile/recipe.ts
+++ b/src/compile/recipe.ts
@@ -6,10 +6,6 @@ import { lw } from '../lw'
 import type { Recipe, Tool } from '../types'
 import { queue } from './queue'
 
-export {
-    parseRepeatedRunsBool
-}
-
 const logger = lw.log('Build', 'Recipe')
 
 let state: {
@@ -281,8 +277,6 @@ function createBuildMagic(rootFile: string, magicTex: Tool, magicBib?: Tool): To
     }
 }
 
-let parseRepeatedRunsBool: boolean
-
 /**
  * Find a recipe based on the provided recipe name, language ID, and root file.
  *
@@ -313,8 +307,6 @@ function findRecipe(rootFile: string, langId: string, recipeName?: string): Reci
     }
     if (recipeName) {
         recipe = recipes.find(candidate => candidate.name === recipeName)
-        const parseRepeatedRuns = recipe?.parseRepeatedRuns ?? true
-        parseRepeatedRunsBool = parseRepeatedRuns
 
         if (recipe === undefined) {
             logger.log(`Failed to resolve build recipe: ${recipeName}.`)
--- a/src/compile/recipe.ts
+++ b/src/compile/recipe.ts
@@ -6,6 +6,10 @@ import { lw } from '../lw'
 import type { Recipe, Tool } from '../types'
 import { queue } from './queue'
 
+export {
+    parseRepeatedRunsBool
+}
+
 const logger = lw.log('Build', 'Recipe')
 
 let state: {
@@ -277,6 +281,7 @@ function createBuildMagic(rootFile: string, magicTex: Tool, magicBib?: Tool): To
     }
 }
 
+let parseRepeatedRunsBool: boolean
 
 /**
  * Find a recipe based on the provided recipe name, language ID, and root file.
@@ -308,6 +313,9 @@ function findRecipe(rootFile: string, langId: string, recipeName?: string): Reci
     }
     if (recipeName) {
         recipe = recipes.find(candidate => candidate.name === recipeName)
+        const parseRepeatedRuns = recipe?.parseRepeatedRuns ?? false
+        parseRepeatedRunsBool = parseRepeatedRuns
+
         if (recipe === undefined) {
             logger.log(`Failed to resolve build recipe: ${recipeName}.`)
             void logger.showErrorMessage(`[Builder] Failed to resolve build recipe: ${recipeName}.`)
--- a/src/compile/recipe.ts
+++ b/src/compile/recipe.ts
@@ -313,7 +313,7 @@ function findRecipe(rootFile: string, langId: string, recipeName?: string): Reci
     }
     if (recipeName) {
         recipe = recipes.find(candidate => candidate.name === recipeName)
-        const parseRepeatedRuns = recipe?.parseRepeatedRuns ?? false
+        const parseRepeatedRuns = recipe?.parseRepeatedRuns ?? true
         parseRepeatedRunsBool = parseRepeatedRuns
 
         if (recipe === undefined) {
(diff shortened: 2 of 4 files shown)


James-Yu:
Thanks for the contribution. However, the current form cannot be merged, as `latexmk` is out of control of the new
```
else if (msg.match(latexRepeatPattern) && parseRepeatedRunsBool) {
        msg = trimLatexRepeat(msg)
    }
```
logic. It is still using its own
```
if (msg.match(latexmkPattern)) {
        msg = trimLaTeXmk(msg)
    }
```
one. Further, the `latexmkSkipped(msg)` is not generic-ized.

If a new generic solution is proposed, I would unify all possible and availabl tools under one same code logic. We can give first-class citizenship to `latexmk`, e.g. default `true` to `parseRepeatedRuns` and only override when it is explicitly set.

yarakos95:
@James-Yu, thank you for your review.

> However, the current form cannot be merged, as `latexmk` is out of control of the new logic.

Would it be okay to branch using `parseRepeatedRunBool` as shown below?

```typescript
if (parseRepeatedRunsBool) {
    if (msg.match(latexRepeatPattern)) {
        msg = trimLatexRepeat(msg)
    }
} else {
    if (msg.match(latexmkPattern)) {
        msg = trimLaTeXmk(msg)
    } else if (msg.match(texifyPattern)) {
        msg = trimTexify(msg)
    }
}
```

----------

> It is still using its own
>
> ```typescript
> if (msg.match(latexmkPattern)) {
>         msg = trimLaTeXmk(msg)
>     }
> ```
>
> one.

This pull request aimed not to replace existing processes, but to extend the functionality to support build tools that had not been considered before.
Therefore, I believe that existing, proven, proprietary logic should be left as is.

In that sense, I believe the default value for `parseRepeatedRuns` should be `false`, and `trimLatexRepeat(msg)` should only be executed when the user explicitly sets it to `true`.

----------

> Further, the `latexmkSkipped(msg)` is not generic-ized.

Other build tools do not have a message equivalent to the log message processed by `latexmkSkipped(msg)`. Even with `llmk`, LaTeX is executed at least once even if there are changes.
Therefore, I don't think it's necessary to touch on this part.

----------

I apologize if my reply was off-topic.

James-Yu:
I may not iterate myself clearly. What I mean `latexRepeatPattern` and `latexmkPattern` should be merged, and `latexmk` is the same as any other tools with `parseRepeatedRunsBool=true`.

I'd love to see a universal solution to all multi-run tools.

yarakos95:
@James-Yu, thank you for the clarification.

Is the expected design to have only the following logic, with `parseRepeatedRuns` defaulting to `true`?

```typescript
if (parseRepeatedRunsBool) {
    if (msg.match(latexRepeatPattern)) {
        msg = trimLatexRepeat(msg)
    }
}
```

I have confirmed that this approach works correctly for build tools including `latexmk`.

However, I would prefer not to change the current default behavior for `latexmk`. My concern is mainly backward compatibility. Since the existing `latexmk` and `texify` logic is already used by current users, I would prefer `parseRepeatedRuns` to remain opt-in and default to `false`, so that existing recipes continue to behave exactly as before unless users explicitly enable the new logic.

```typescript
if (parseRepeatedRunsBool) {
    if (msg.match(latexRepeatPattern)) {
        msg = trimLatexRepeat(msg)
    }
} else {
    if (msg.match(latexmkPattern)) {
        msg = trimLaTeXmk(msg)
    } else if (msg.match(texifyPattern)) {
        msg = trimTexify(msg)
    }
}
```

This would extend support to additional multi-run tools while preserving the current behavior for existing users.

I agree that a fully unified solution would be valuable, but I see that as a larger architectural change. For this pull request, I would appreciate it if this backward-compatible approach could be accepted...

James-Yu:
> @James-Yu, thank you for the clarification.
> 
> Is the expected design to have only the following logic, with `parseRepeatedRuns` defaulting to `true`?
> 
> ```ts
> if (parseRepeatedRunsBool) {
>     if (msg.match(latexRepeatPattern)) {
>         msg = trimLatexRepeat(msg)
>     }
> }
> ```

Yes. That's what in my mid.

> However, I would prefer not to change the current default behavior for `latexmk`. My concern is mainly backward compatibility. Since the existing `latexmk` and `texify` logic is already used by current users, I would prefer `parseRepeatedRuns` to remain opt-in and default to `false`, so that existing recipes continue to behave exactly as before unless users explicitly enable the new logic.
>
> This would extend support to additional multi-run tools while preserving the current behavior for existing users.
> 
> I agree that a fully unified solution would be valuable, but I see that as a larger architectural change. For this pull request, I would appreciate it if this backward-compatible approach could be accepted...

I understand your concern. However, I don't think that objects the unified design, for the most compatibility we can still use the follow (pseudo-code)
> ```ts
> if (parseRepeatedRunsBool || ['latexmk', 'texify'].contains(tool.command)) {
>     if (msg.match(latexRepeatPattern)) {
>         msg = trimLatexRepeat(msg)
>     }
> }
> ```
that just effectively disable `parseRepeatedRunsBool` check for the said tools. We may have a better detection here, even with explicit regexp check, but under the hood, we are using the same logic.

yarakos95:
@James-Yu, thank you for understanding my concerns. That said, I have decided to agree with the proposal to unify the logic specific to `latexmk` and `texify`.

I apologize for dragging out this discussion.

However, there is one point:
If we proceed this way, `parseRepeatedRunsBool` will no longer be necessary. Even if it is set to `true` or `false` when the build tools are not used, log analysis will still run without any issues.

Should I remove this?

James-Yu:
That's right, we can (and should) always extract the last part of log for parsing.

yarakos95:
@James-Yu, I apologize for the late commit. 

As a this commit, the log parsing pattern now always starts with the followiing string:

```regex
/^This\sis\s(pdf|LuaHB|Xe|e-up)TeX,\sVersion/
```

James-Yu:
I think this is a very interesting modification to the log parsing logic. @jlelong May I have your opinion on this?

jlelong:
It looks great to me.

Just a tiny concern regarding maintenance: the variables `latexRepeatLog` and `latexRepeatLogLatex` hold exactly the same pattern. I am not sure to understand the need for that all the more so as the difference in their names does not sound so meaningful. I am not saying that it should be changed, but a little comment would certainly help.

yarakos95:
@jlelong, this follows the variable naming conventions used for processing logs in previous versions of `latexmk.`

https://github.com/James-Yu/LaTeX-Workshop/blob/19468dbb183c5db0b6dcdc45e69a909527458429/src/parse/parser.ts#L82-L84

I added a comment to that effect.

Issue #4538, opened 2025-02-24
Improving citation filter despite 128 character limit


ChristopherBrix:
## Pre-checks*
> _Please change the following `[ ]` to `[x]` for confirmation._
- [x] The feature request has not been suggested in this repository.

## The Missed*
Filtering suggested citations can only access the first 128 characters. This makes using non-standard fields such as `keywords` unstable.

## The Solution*
Could the order of the fields in `latex-workshop.intellisense.citation.format` be used to define in what order the text is generated that will be used for the filtering? Then I could rank them by importance and e.g. add a custom `keywords` entry at the top, so it is guaranteed to be included.

## Anything Else?
The limit of 128 characters (previously 54) is enforced by VS Code (https://github.com/microsoft/vscode/issues/74133). Maybe it's also worth a shot to reopen that ticket to ask for  another increase (or ideally, just to drop this limit)?

ChristopherBrix:
After writing this, I realized the relevant order is the order in the `.bib` file. If I move the `keywords` field to the top of the entry, I can use if for my search.

I don't think this is documented somewhere. Maybe doing so would be useful? I'll keep the issue open so this can be decided, but the feature request can obviously be dropped.

James-Yu:
Added. Also, the message can be found in the description of config `latex-workshop.intellisense.citation.filterText`.

ChristopherBrix:
In fact, none of the two versions seem to reliably control the order.

I've set `"latex-workshop.intellisense.citation.format": ["author", "title"]` and created the `.bib` file
```
@inproceedings{article1,
  author    = {First Author and Second Author},
  title     = {aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeeeffffffffffgggggggggghhhhhhhhhhiiiiiiiiiijjjjjjjjjj},
}
```

If I now create a new citation `\parencite{aaaaa` I get `article1` as a recommendation. Even if I enter the entire title, it does not disappear. Even though the title has 100 characters, and the `author` field has 30. In addition to the "Author: " and "Title: " text, the end of the title should be cut off for search.
But if I write `\parencite{se`, the article disappears from the list of recommendations. So apparently, the string that's used for searching always uses the `title` field first, even though both the `.bib` order and the config setting place `author` first.

James-Yu:
You mis-understand the two configs. Please read their descriptions again.

ChristopherBrix:
[...]

Issue #4522, opened 2025-02-06
Unable to restrict the search target for auto-completion of bib entries


mayukonomoto:
When I type one or a few letters to narrows down the candidates using automatic completion for bib entries, what is being searched for?
I want to narrow down the candidates to only those that contain the input characters in the citationkey, and sort them in order of similarity to the input characters. Is there a way to set this?

In the example of the attached image, it shows the citationkey on the candidate list. However, even though there are other citationkeys that include all or part of the characters “mil” in them, they appear very late in the list or not at all.

![Image](https://github.com/user-attachments/assets/db4232e0-4d45-41ca-ae0b-893f2747a4bf)

James-Yu:
https://github.com/James-Yu/LaTeX-Workshop/blob/7c8a632aca6a3859fe4896c0fc3ea0a49623f402/src/completion/completer/citation.ts#L137C27-L137C43

The filtering is based on key, title, and other bib entry fields. The sorting is controlled by vscode: we cannot change that.

@jlelong This issue makes me think, will it be a good idea to customize the `filterText` of bib entries? Current implementation can be a bit ambigious?

jlelong:
@James-Yu it would great to make the `filterText` customizable through a configuration variable.

mayukonomoto:
Thank you for your quick response.
I will make use of the new setting.

I forgot to mention it at the beginning, but your plugin is always helping me with my work.
Thank you very much!

PR #4969, opened 2026-08-28
Add opt-in fuzzy ranking for \cite{} completion

isaac-tes:
Adds an opt-in setting, `latex-workshop.intellisense.citation.fuzzy` (default `false`), that ranks `\cite{}` suggestions by how well they match what you've typed instead of leaving the order to VS Code's built-in filter. This targets the ranking request in #4522.

## Why

Citation suggestions are filtered through a `filterText` string that VS Code caps at 128 characters (#4538). Anything past that cap, usually a later co-author, can't be matched, so you can't find an entry by an author who isn't near the front of the string. Ordering is left to VS Code as well. On a large `.bib` this makes it hard to jump to the entry you want by typing part of a key, author, or year.

## What it does

With the setting on, entries are scored by a small fzf-style matcher and the best matches come first. The bibtex key carries the most weight, then the fields from `intellisense.citation.format` in the order they're listed. Terms are space-separated with AND semantics, so `smith 2020` requires both.

There's no new runtime dependency. The matcher is a self-contained port of fzf's scoring model (word-boundary, camelCase and consecutive-run bonuses, gap penalties, first-character multiplier, smart-case) with fzf's V1 span selection, about 100 lines with no global state. A term only counts if it matches a field reasonably tightly, so a surname smeared as scattered letters across a ten-author list doesn't produce a false hit.

It works in both citation modes. In inline mode the provider ranks the items, sets `sortText`/`filterText` so VS Code keeps that order, and returns an incomplete `CompletionList` so it re-queries as you type. In browser mode the QuickPick is driven directly so the ranking decides what appears, with matched characters still highlighted across key, title and author.

## Default behaviour is unchanged

The setting defaults to `false`. With it off, the current completion behaves exactly as before.

## Tests

Unit tests under `test/units/05_completion/` cover the ranker (matching, weighting, AND semantics, smart-case, long-field handling, weak-match rejection) and both the inline and browser integrations.


James-Yu:
Besides Copilot's suggestions, I noticed that a new `insertCitationKey` is included but seems to be unrelated to fuzzy ranking. Please move these changes to a new PR and make this one atomic.

isaac-tes:
Thanks for the review. Addressed all of it.

- Removed the `insertCitationKey` refactor so this PR is fuzzy-only. The `browser()` legacy path is back to its original code, and the fuzzy path does its own insertion inline. I'll open the `insertCitationKey` extraction (deduping both call sites) as a separate PR once this lands.
- Copilot: the `intellisense.citation.fuzzy` read now passes the document URI so resource-scoped overrides apply; the inline path ranks only when the current word is non-whitespace (`query.trim().length > 0`), matching the browser path; and the duplicated comment line is gone.

James-Yu:
Great work!

PR #4975, opened 2026-09-04
Improve citation browser labels and multi-selection

sparkwenc:
## Summary
- make the citation browser honor `latex-workshop.intellisense.citation.label` and sort by the selected primary label
- allow selecting multiple entries and insert their citation keys as a comma-separated list
- keep citation keys available independently of their display labels

## Tests
- `npm run compile`
- `npm run lint`
- VS Code unit suite: 1224 passing

### DIFF
--- a/package.json
+++ b/package.json
@@ -2202,11 +2202,11 @@
             "authors"
           ],
           "default": "bibtex key",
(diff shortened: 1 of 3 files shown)


sparkwenc:
Withdrawing this proposal after further usability testing. The multi-select QuickPick requires explicitly checking at least one item before Enter, which adds more friction than the existing single-select workflow.

PR #4978, opened 2026-09-09
Allow compiler log to preserve editor focus

doraTeX:
`latex-workshop.compilerlog` currently always focuses the compiler output because it calls `OutputChannel.show()` without exposing VS Code's `preserveFocus` argument.

This makes the command harder to compose with other commands. For example, when using VS Code's `runCommands` to:

1. show the compiler log,
2. start a LaTeX Workshop recipe,
3. open the generated PDF,

the focus change caused by `latex-workshop.compilerlog` can interfere with subsequent commands.

This PR adds an optional `preserveFocus` argument to `latex-workshop.compilerlog` and passes it through to `OutputChannel.show(preserveFocus)`.

Existing behavior is unchanged because `preserveFocus` defaults to `false`.

For example:

```json
{
    "command": "latex-workshop.compilerlog",
    "args": [true]
}
```

This shows the LaTeX Compiler output while preserving editor focus, which allows it to be used before commands such as `latex-workshop.recipes` in a `runCommands` sequence.

I verified the change in an Extension Development Host and confirmed that the compiler output is shown before the build starts, the build proceeds normally, and the compiler log is updated in real time.

`npm run lint`, `npm run compile`, and `npm run test` all pass.

### DIFF
--- a/src/core/commands.ts
+++ b/src/core/commands.ts
@@ -176,10 +176,10 @@ export function wordcount() {
     }
 }
 
-export function showLog(compiler?: string) {
+export function showLog(compiler?: string, preserveFocus = false) {
     logger.log(`SHOWLOG command invoked: ${compiler || 'default'}`)
     if (compiler) {
-        logger.showCompilerLog()
+        logger.showCompilerLog(preserveFocus)
     } else {
         logger.showLog()
     }
--- a/src/main.ts
+++ b/src/main.ts
@@ -187,7 +187,7 @@ function registerLatexWorkshopCommands(extensionContext: vscode.ExtensionContext
         vscode.commands.registerCommand('latex-workshop.addtexroot', () => lw.commands.addTexRoot()),
         vscode.commands.registerCommand('latex-workshop.wordcount', () => lw.commands.wordcount()),
         vscode.commands.registerCommand('latex-workshop.log', () => lw.commands.showLog()),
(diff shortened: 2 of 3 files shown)


doraTeX:
For consistency, it may also make sense to give `latex-workshop.log` the same optional `preserveFocus` argument, since it calls `LOG_PANEL.show()` in exactly the same way. 

The corresponding change would be symmetric with the one in this PR:

```ts
// src/utils/logger.ts
function showLog(preserveFocus = false) {
    LOG_PANEL.show(preserveFocus)
}
```

```ts
// src/core/commands.ts
export function showLog(compiler?: string, preserveFocus = false) {
    logger.log(`SHOWLOG command invoked: ${compiler || 'default'}`)
    if (compiler) {
        logger.showCompilerLog(preserveFocus)
    } else {
        logger.showLog(preserveFocus)
    }
}
```

```ts
// src/main.ts
vscode.commands.registerCommand('latex-workshop.log', (preserveFocus?: boolean) => lw.commands.showLog(undefined, preserveFocus)),
```

As with `latex-workshop.compilerlog`, the default remains `false`, so existing behavior is unchanged.

James-Yu:
I agree with you. May you make a new commit?

doraTeX:
Done! I've added and pushed a new commit implementing this change.

PR #4984, opened 2026-09-15
feat: allow disabling language providers

CyberStefNef:
Add `latex-workshop.languageFeatures.enabled` so users can use another language server while keeping LaTeX Workshop for builds, PDF viewing, and SyncTeX.

The setting defaults to `true`. Setting it to `false` disables completion, hover, definitions, symbols, formatting, code actions, folding, selection ranges, and document colors after reloading the window. Linters and build diagnostics retain their existing settings.


James-Yu:
Thanks for the PR. However, the extension won't support disabling part of the feature set.

